PR #10, opened 2026-03-13
chore(ci): Add integration test for workflows

bnchrch:
## Problem
Currently there is a possibility we push a change that could break the CLI or test_workflows at runtime

## Solution
Add an integration test that
1. Runs the cli
2. Runs a test workflow

closes https://linear.app/growthx/issue/OUT-289/add-manual-integration-test-for-test-workflows-scenarios


sdk/cli/src/assets/docker/docker-compose-dev.yml
```diff
@@ -109,7 +109,7 @@ services:
       interval: 3s
       timeout: 10s
       retries: 20
-      start_period: 30s
+      start_period: 300s
```
catsby: This affects everyone, not just CI. If local development causes the worker to crash, users will wait ~5 minutes for the worker to come up even if it crashed immediately. 

What do you think about an file just to override this `start_period` value, and invoking it in the `dev_integration_test.sh`?

.github/workflows/dev_integration_test.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: actions/checkout@v4
+
+      - name: Create test_workflows/.env
+        run: |
+          printf 'ANTHROPIC_API_KEY=%s\nOPENAI_API_KEY=%s\n' \
+            "${ANTHROPIC_API_KEY:-dummy}" \
+            "${OPENAI_API_KEY:-dummy}" > test_workflows/.env
+        env:
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
```
catsby: Do we need this `.env` file? The workflow we run doesn't use any credentials

.github/workflows/dev_integration_test.yml
```diff
@@ -0,0 +1,34 @@
+name: Workflow System Integration Test
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'test_workflows/**'
+      - 'api/**'
+      - 'sdk/**'
```
catsby, line 9: to confirm, this will run on every PR to `main` that touches any of these code paths, from any contributor?

.github/workflows/dev_integration_test.yml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ github.actor != 'dependabot[bot]' }}
+    permissions:
+      contents: read
+    timeout-minutes: 30
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
```
catsby, line 23: We don't setup node or enable corepack specifically here, so I guess `ubuntu-latest` just has them enabled? Should we be concerned about version matching and/or drift?

ops/dev_integration_test.sh
```diff
@@ -0,0 +1,99 @@
...
+    echo "--- worker logs ---"
+    docker compose -f "$COMPOSE_FILE" --project-directory "$(pwd)" \
+      logs worker --tail 200 2>/dev/null || true
+    exit 1
+  fi
+  sleep 3
+done
+print "Worker connected and catalog available (after $((ATTEMPT * 3))s)"
+
+# Phase 3: run the simple workflow using the CLI — pure math, no LLM required.
```
catsby:
```suggestion
# Step 4: run the simple workflow using the CLI — pure math, no LLM required.
```

catsby (commented):
I left some questions for discussion

sdk/cli/src/assets/docker/docker-compose-dev.yml
bnchrch: Yup good call

.github/workflows/dev_integration_test.yml
bnchrch: Yup that is correct. We can make this a manual or a nightly run

.github/workflows/dev_integration_test.yml
bnchrch: were running docker in docker so those version are more or less set by our cli docker compose. But yeah good idea to pin them here too

ops/dev_integration_test.sh
```diff
@@ -0,0 +1,100 @@
...
+  if [ "$ATTEMPT" -ge "$MAX_ATTEMPTS" ]; then
+    print "TIMEOUT: Worker did not connect after $((MAX_ATTEMPTS * 3))s" "✗"
+    echo "--- worker logs ---"
+    docker compose -f "$COMPOSE_FILE" --project-directory "$(pwd)" \
+      logs worker --tail 200 2>/dev/null || true
+    exit 1
+  fi
+  sleep 3
+done
+print "Worker connected and catalog available (after $((ATTEMPT * 3))s)"
```
szanata: Wouldnt be better to wait for the API to init, since the API depends on the worker being healthy? This way is guaranteed that everything is on

ops/docker-compose.ci.yml
```diff
@@ -0,0 +1,4 @@
+services:
+  worker:
+    healthcheck:
+      start_period: 300s
```
szanata: The same question I posted earlier, why do we need this, can't we depend solely on the API starting?

szanata (approved):

PR #13, opened 2026-03-16
chore(cli): Fix creds editor env

bnchrch:
## Problem

- `EDITOR="cursor --wait"` caused `spawnSync` to look for an executable literally named `"cursor --wait"`, failing with `status: null`
- The command returned "Saved successfully" immediately without waiting for the editor to close

## Solution

- Split the `EDITOR` env var on whitespace so flags like `--wait` are passed as separate args to `spawnSync`
- Added an explicit `result.error` check to surface spawn failures with a clear message

### DIFF
--- a/sdk/cli/src/commands/credentials/edit.ts
+++ b/sdk/cli/src/commands/credentials/edit.ts
@@ -46,14 +46,19 @@ export default class CredentialsEdit extends Command {
       );
     }
 
-    const editor = process.env.EDITOR || process.env.VISUAL || 'vi';
+    const editorEnv = process.env.EDITOR || process.env.VISUAL || 'vi';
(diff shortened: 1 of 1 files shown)


catsby (approved):

sdk/cli/src/commands/credentials/edit.ts
```diff
@@ -46,14 +46,19 @@ export default class CredentialsEdit extends Command {
       );
     }
 
-    const editor = process.env.EDITOR || process.env.VISUAL || 'vi';
+    const editorEnv = process.env.EDITOR || process.env.VISUAL || 'vi';
```
szanata, line 49: we could test vim and nano as well?

szanata (approved):

PR #19, opened 2026-03-16
chore(ops): New npm publish scripts

szanata:
Refactoring publish scripts:
- Prod publish: Massively simplifying the script;
- Dev publish: Also simplifying, but updating it to use `npm version prerelease` strategy instead of what was set before. With this method, packages will be published with version `x.x.x-dev.Y`, where Y is incremental. It will also set a tag `dev` to each package.

Output sample:image


test_workflows/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "description": "A simple output core abstract",
   "type": "module",
+  "private": true,
```
szanata, line 6: Prevent from being published

.gitignore
```diff
@@ -24,3 +24,5 @@ agent-os/
 .claude/settings.local.json
 .claude/worktrees/
 
+# Pnpm publish report
+pnpm-publish-summary.json
```
szanata, line 28: This file will be generated by the publish_npm_dev script and will have the summary. Eg:

```json
{
	"publishedPackages": [
		{
			"name": "@outputai/core",
			"version": "0.1.1-dev.0"
		},
		{
			"name": "@outputai/credentials",
			"version": "0.1.1-dev.0"
		},
		{
			"name": "@outputai/http",
			"version": "0.1.1-dev.0"
		},
		{
			"name": "@outputai/llm",
			"version": "0.1.1-dev.0"
		},
		{
			"name": "@outputai/evals",
			"version": "0.1.1-dev.0"
		},
		{
			"name": "@outputai/cli",
			"version": "0.1.1-dev.0"
		},
		{
			"name": "@outputai/output",
			"version": "0.1.1-dev.0"
		}
	]
}
```

This will be send as a comment to the PR after the publish

szanata:
Publish worked but since I changed the return of the script, it broke the current GH Action. Logs here: https://github.com/growthxai/output/actions/runs/23168822335

bnchrch (dismissed):

ops/publish_npm_prod.sh
```diff
@@ -1,59 +1,19 @@
 #!/bin/bash
 
-# Publish prod versions of the SDK packages to NPM.
-# For each package in the sdk it will check if the package version is different from NPM.
-# If it is, the script will publish the new version.
+# Publish each package to NPM (if version is not published)
 
 set -e
 
 cd "${0%/*}"
```
catsby: Does it matter that we're in the `ops/` folder when we run this? Does `pnpm publish` walk up the directory to find `pnpm-workspace.yml` in the root?

catsby (dismissed):
Question on where we execute `pnpm` other than that looks good

ops/publish_npm_prod.sh
szanata: apparently yes. But I will update the path to make it consistent.

catsby (approved):

PR #22, opened 2026-03-17
chore: Adding changeset workflow

szanata:
This PR adds [changesets](https://github.com/changesets/changesets), a library to help manage the release cycle of the project.

New release flow:

1. Write your code in a branch;
2. If you change `sdk/*` or `api/` (non–chore PRs):
  2.1. Run `pnpm changeset` and follow the on-screen instructions;
  2.2. Commit the new file(s) under `.changeset/*`;
3. Open your PR:
  3.1. If the PR needs rework: remove the current changeset file from the branch and go back to step 2.1;
4. Merge and delete the branch;
5. The release GH workflow will open or update a PR titled “Version Packages” with the version bumps and changelog updates; merge that PR;
6. Merging that PR re-runs the workflow; that run publishes the packages to npm and the API image to Docker Hub. It will also push a tag named after the version being bumped.


test_workflows/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "test_workflows",
-  "version": "0.1.0",
+  "version": "0.0.1",
```
szanata: This package does not need bumps.

bnchrch (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,65 @@
...
+          commit: 'chore(release): version packages'
+          title: 'Version Packages'
+          publish: |
+            ./ops/validate.sh &&
+            ./ops/publish_npm_prod.sh &&
+            ./ops/publish_api_image.sh
+
+      - name: Get version
+        id: version
+        if: steps.changesets.outputs.published == 'true'
```
catsby, line 58: I'm not sure this will work, have you been able to verify it? `changesets/action@v1` runs whatever script(s) you give it in `publish` and ends up parsing that output for values to populate it's `.output`, and I'm not sure we give it the right output to set `published` to `true`:

- https://github.com/changesets/action/blob/a7c500273b5be4fb897d21ceededa4767223a843/src/run.ts#L81

I'm not 100% sure though, but worth verifying

ops/release.sh
```diff
@@ -1,32 +0,0 @@
...
-  ./ops/print_warning.sh "error" "Invalid version." "Content of the version file is not semver (found \"$version\")"
-  exit 1
-fi
-
-print "Bumping workspace"
-update_cmd="npm version \"$version\" --no-git-tag-version --allow-same-version --silent >/dev/null && echo \"- \$PNPM_PACKAGE_NAME: v$version\""
-pnpm -r --include-workspace-root exec sh -c "$update_cmd"
-
-print "Updating CLI version"
-pnpm run --silent --filter @outputai/cli set-sdk-version $version
```
catsby, line 30: Is running `set-sdk-version` no longer needed?

.github/workflows/release.yml
```diff
@@ -0,0 +1,65 @@
...
+
+env:
+  NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+  DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+  DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+
+jobs:
+  release:
+    name: Version PR or Publish
+    runs-on: ubuntu-latest
```
catsby, line 15: Do we need to guard against dependabot like we did in previous release workflows?

.github/workflows/release.yml
```diff
@@ -0,0 +1,65 @@
...
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+env:
+  NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+  DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+  DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
```
catsby: we should move these to only the steps that need them, which i believe is just the `Create Release Pull Request or Publish` step, but `Install dependencies` may need the `NPM_TOKEN` as well

ops/release.sh
szanata: Very good catch!

.github/workflows/release.yml
szanata: No, if no changesets were added this will not trigger the version/publish

.github/workflows/release.yml
szanata: After double checking, I think it will still work, but since it is hard to test, I will have to verify after merging.

.github/workflows/release.yml
szanata: Moved

PR #23, opened 2026-03-18
feat(credentials): credential: env var convention for automatic secret resolution

bnchrch:
## Summary

- Introduces a \`credential:<dot.path>\` convention for env var values — any env var set to e.g. \`ANTHROPIC_API_KEY=credential:anthropic.api_key\` is automatically resolved from encrypted credentials at worker startup
- Adds a lightweight startup hook registry to \`@outputai/core\` (\`registerStartupHook\` / \`runStartupHooks\`) so packages can register startup behaviour without creating a circular dependency — core never imports credentials
- \`@outputai/credentials\` self-registers \`resolveCredentialRefs\` as a startup hook on import; worker calls \`runStartupHooks()\` after loading activities, before creating the Temporal worker
- Fixes build scripts to copy \`*.key\` files to \`dist/\` alongside \`*.yml.enc\` — without this, per-workflow credentials fell back to the global key and failed to decrypt at runtime
- Adds 3 new Claude Code skills: \`output-credentials-init\`, \`output-credentials-edit\`, \`output-credentials-env-vars\`; updates skill inventory (29 → 32), agent skill tables, and session start hook
- Adds env var convention section to both credentials docs pages (\`packages/credentials.mdx\` and \`operations/credentials.mdx\`)
- Integration tests verify: resolution, non-destructiveness (real env vars take precedence), missing path handling, and idempotency
- Verified end-to-end: \`credentials_demo\` workflow runs successfully, reading \`test.secret\` → \`"workflow_specific_secret"\` from per-workflow encrypted credentials

## Test plan

- [x] \`npm run lint\` — no errors
- [x] \`npm run build:packages\` — all packages build
- [x] \`npm run test:integration\` — integration tests pass (including 4 new \`resolveCredentialRefs\` tests)
- [x] \`npm run output -- workflow run credentials_demo --input '{"path": "test.secret"}'\` — returns \`{"value": "workflow_specific_secret"}\`
- [ ] Set \`ANTHROPIC_API_KEY=credential:anthropic.api_key\` in \`.env\`, put the real key in encrypted credentials, start the worker — verify log line \`Startup hooks resolved env vars\` and LLM calls succeed
- [ ] Set a real \`ANTHROPIC_API_KEY\` directly in the environment — verify the credential ref is NOT used (value doesn't start with \`credential:\`)


sdk/core/src/worker/index.js
```diff
@@ -40,6 +41,12 @@ const callerDir = process.argv[2];
   log.info( 'Loading activities...', { callerDir } );
   const activities = await loadActivities( callerDir, workflows );
 
+  log.info( 'Running startup hooks...' );
+  const resolved = runStartupHooks();
+  if ( resolved.length > 0 ) {
+    log.info( 'Startup hooks resolved env vars', { vars: resolved } );
+  }
+
```
szanata: I don't think this is a good approach.

1. Probably it isn't concern of core to know it is suppose to resolve env vars, so the log info assumes only the usage from credentials, inverting responsibilities;

2. Also, adding these each time we need callbacks will end up polluting core too much.

What I am thinking is either:
a. We find a way to avoid needing this altogether;
b. We use the event bus:

```js
import {  on } from '@outputai/core/hooks';

...
on( 'before_start', () => .... )
```

And in the `core/src/worker`:
```js
messageBus.emit( BusEventType.BEFORE_START );
```
But I would vote to avoid adding this if possible.

szanata (changes requested):

szanata (approved):

PR #68, opened 2026-03-23
feat(cc-plugin): add workflow cost skill and fix null pricing config crash

bnchrch:
Closes OUT-327

## Problem

- `npx output workflow cost` crashes with `Cannot read properties of null (reading 'models')` when a project-level `config/costs.yml` exists but is empty, because `yaml.load()` returns `null` and the spread `...project.models` throws
- The bundled pricing config had no null guard, so any empty/missing YAML would also crash rather than degrade gracefully
- No CC plugin skill existed to teach the agent how to use `npx output workflow cost`, making it hard to discover during demos

## Solution

- Fixed `loadPricingConfig` in `cost_calculator.ts` to guard against `null` from `yaml.load()`: falls back to `{ models: {}, services: {} }` with a warning if bundled config is null, and uses optional chaining on project-level config
- Added new skill `output-dev-workflow-cost` documenting the command, its flags, the workflowId vs workflowName distinction, and fallback steps for getting a run ID
- Bumped plugin version to `0.1.2` and updated the skills inventory in `output-meta-project-context` (32 → 33 skills)

### DIFF
--- a/.changeset/clean-ads-matter.md
+++ b/.changeset/clean-ads-matter.md
@@ -1,0 +1,5 @@
+---
+"@outputai/cli": patch
+---
+
+Fix null crash in `workflow cost` when pricing config is empty or missing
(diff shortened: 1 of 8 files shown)


bnchrch:
Confirmed working
image

szanata (approved):

PR #71, opened 2026-03-23
chore: Improving pub/release GH workflows and ops scripts

szanata:
Streamlining ops scripts for release/publish operations:
- `bump_prerelease.sh <preid>`: Bumps all packages to a new prerelease version with given preid (eg: dev)
- `publish_api.sh`: Pushes the API do Dockerhub. Accepts optional `--no-aliases` to skip publishin `:latest`,`:x` and `:x.y` aliases
- `publish_npm.sh <tag>`: Builds and Publishes pkgs to npm, optionally setting a tag.
- `bump.sh` -> `bump_release.sh`: Rename only

All GH pub actions are in line as well:
- `publish.yml`: publishes npm, publishes api (with aliases) and tag. Skips validation and bump, since it happened when merging to main;
- `publish_dev.yml`: validate, bumps, publish npm, publishes api (no aliases);
- `publish_next.yml`: bumps, publish npm, publishes api (no aliases). Skips validation, since it happened when merging to main.

Finally, with these changes both "dev" and "next" versions now publish the api as well.


.github/workflows/publish.yml
```diff
@@ -34,18 +34,19 @@ jobs:
...
-      - name: Install dependencies
-        run: pnpm install --frozen-lockfile
-
-      - name: Publish
+      - name: Publish NPM
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: ./ops/publish_npm.sh
+
+      - name: Publish API
```
bnchrch, line 40: ❗ We should publish the api before the npm packages.

Why? the docker compose in the cli points to the api. So if the npm publishes without the api image ready then `output dev` could break.

The reverse isnt true

bnchrch (changes requested):

bnchrch (approved):

PR #80, opened 2026-03-25
refactor(api): Moving swagger docs from comments to swagger.yml

szanata:
## Changes
- Removing all jsdoc style comments with `@swagger` tag from `api/api.js` and copying the content to a new file `openapi.yml`: this drastically reduces the file size and makes it much easier to read by keeping only actual code there;
- Renaming `api/api.js` to `api/index.js`: same pattern as other projects;
- Refactoring script `generate_openapi.js` to use the new source `openapi.yml`: now it simple converts a `.yml` spec to the `.json`.
- Removing `generate_openapi.js` secondary output (`docs/guides/openapi.json`): The `api/` should not know about things in the sibling folders of the project, so this was replaced by a simple `cp` when building the project at the root command `build:packages`.

## QA
- Q: Why keeping the swagger docs in yml, why not cutting the middleman an just keeping all the docs in the JSON file as the source of truth?
  A: A long JSON file with the api spec will be very hard to read, even considering the LLM will do most of the changes there, it will still be hard to review them.

- Q: Reason of the refactor?
  A: It was becoming to hard to maintain the `api/api.js` (now `api/index.js`). For each line of code there were 5 lines of comments.

- Q: Did this change the output spec file?
  A: No, the `openapi.yml` file generate by this new flow is exactly the same as it was before, so much so that it didn't even appear here.


szanata:
This will be revisited later.

PR #91, opened 2026-03-29
chore: fix docker compose and service to allow new package managers and surface error states sooner

bnchrch:
## Summary
- Reduces the time we allow for a worker to be unhealthy before reporting it back to docker
- Updates our docker service logic in the CLI to be more clear about what is starting and what is healthy
- Enables corepack in the worker container so the project's declared package manager (yarn, pnpm, etc.) is available
- Detects the package manager from `packageManager` in `package.json` at runtime and uses it for all install/build/start commands instead of hardcoding `npm`
- Adds `COREPACK_ENABLE_DOWNLOAD_PROMPT=0` to suppress interactive prompts during package manager downloads
- Fixes signal handling by using `exec` before nodemon

## Context

Node 24's Corepack blocks `npm` when a project sets `"packageManager": "yarn"` (or `pnpm`) in `package.json`. The dev container was hardcoded to use `npm`, making it impossible to use any other package manager.

## Test plan

- [ ] Start dev environment with a project that has `"packageManager": "yarn@4.x.x"` — worker should install deps with yarn and start successfully
- [ ] Start dev environment with a project using `"packageManager": "pnpm@x.x.x"` — same
- [ ] Start dev environment with a project using npm (no `packageManager` field) — should work as before
- [ ] Verify e2e tests still pass: `./run.sh validate`

Closes OUT-330

### DIFF
--- a/sdk/cli/src/assets/docker/docker-compose-dev.yml
+++ b/sdk/cli/src/assets/docker/docker-compose-dev.yml
@@ -131,7 +131,7 @@ services:
         corepack enable &&
         npm run output:worker:install &&
         echo 'Installed dependencies' &&
-        exec npx nodemon --watch src --watch package.json --ext ts,js,json,prompt --ignore 'dist/**' --ignore '**/*.test.ts' --ignore '**/*.spec.ts' --exec 'npm run output:worker:install && npm run output:worker:build && npm run output:worker:start'
+        npx nodemon --watch src --watch package.json --ext ts,js,json,prompt --ignore 'dist/**' --ignore '**/*.test.ts' --ignore '**/*.spec.ts' --exec 'npm run output:worker:install && npm run output:worker:build && npm run output:worker:start'
       "
     working_dir: /app/${OUTPUT_WORKFLOWS_DIR:-.}
     volumes:
--- a/sdk/cli/src/assets/docker/docker-compose-dev.yml
+++ b/sdk/cli/src/assets/docker/docker-compose-dev.yml
@@ -129,10 +129,9 @@ services:
     command: >
       sh -c "
         corepack enable &&
-        PM=$$(node -p 'try{(JSON.parse(require(\"fs\").readFileSync(\"package.json\",\"utf8\")).packageManager||\"npm\").split(\"@\")[0]}catch(e){\"npm\"}') &&
-        $$PM run output:worker:install &&
+        npm run output:worker:install &&
         echo 'Installed dependencies' &&
-        exec npx nodemon --watch src --watch package.json --ext ts,js,json,prompt --ignore 'dist/**' --ignore '**/*.test.ts' --ignore '**/*.spec.ts' --exec \"$$PM run output:worker:install && $$PM run output:worker:build && $$PM run output:worker:start\"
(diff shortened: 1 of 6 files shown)


sdk/cli/src/assets/docker/docker-compose-dev.yml
```diff
@@ -107,9 +107,9 @@ services:
     healthcheck:
       test: [ 'CMD', 'npx', '--yes', 'output-healthcheck' ]
       interval: 3s
-      timeout: 10s
-      retries: 20
-      start_period: 30s
+      timeout: 3s
+      retries: 2
+      start_period: 60s
```
szanata, line 112: Could you explain the reasoning behind these values?

sdk/cli/src/assets/docker/docker-compose-dev.yml
```diff
@@ -127,9 +128,10 @@ services:
       - NODE_OPTIONS=${NODE_OPTIONS:---max-old-space-size=4096}
     command: >
       sh -c "
+        corepack enable &&
         npm run output:worker:install &&
         echo 'Installed dependencies' &&
-        npx nodemon --watch src --watch package.json --ext ts,js,json,prompt --ignore 'dist/**' --ignore '**/*.test.ts' --ignore '**/*.spec.ts' --exec 'npm run output:worker:install && npm run output:worker:build && npm run output:worker:start'
+        exec npx nodemon --watch src --watch package.json --ext ts,js,json,prompt --ignore 'dist/**' --ignore '**/*.test.ts' --ignore '**/*.spec.ts' --exec 'npm run output:worker:install && npm run output:worker:build && npm run output:worker:start'
```
szanata: Wouldnt the `exec` change the execution chain of this command? How this allows for others package managers?

szanata:
> Detects the package manager from packageManager in package.json at runtime and uses it for all install/build/start 
commands instead of hardcoding npm

Where it does this?

sdk/cli/src/assets/docker/docker-compose-dev.yml
bnchrch: These are the grace periods before a healthcheck failing is actually reported as a failure to docker consumers.

We were being overly generous leading to the worker failing not showing in the docker ps status for many minutes.

So I reduced the time we accept a failing health check

and increased the start_period to properly compensate for the workers startup time

bnchrch:
> > Detects the package manager from packageManager in package.json at runtime and uses it for all install/build/start
commands instead of hardcoding npm

> Where it does this?

This is an old description from the very first version claude put up before I reviewed.

I figured it was unnecessary to have the packageManager change `npm run outputai:worker:build` to `yarn run outputai:worker:build`

sdk/cli/src/assets/docker/docker-compose-dev.yml
bnchrch: Ill revert this. Claude made this change to fix up some exit code handling when nodemon failed.

But thinking about this I think it was only important for when it was updating npx to use yarn/pnpm equivalents 

which we no longer do

will remove

szanata (approved):

PR #93, opened 2026-03-30
fix(cli): fix plan and generate command save instructions and logic

bnchrch:
## Problem

- **OUT-331** — When `workflow plan` is invoked via the CLI, Claude attempts to write the plan file itself and fails because write tools are not in its allowed set, resulting in the message "I wasn't able to write the plan file due to permissions" and no file on disk
- **OUT-332** — Claude's post-flight instructions tell the user to run `/outputai:build_workflow` (the Claude slash command) instead of `npx output workflow generate`, and references the plan file before it has been written
- **OUT-336** — `workflow generate --plan-file` creates the boilerplate skeleton before validating the plan file exists; if the plan is missing (or build fails), the orphaned skeleton is left on disk with no rollback; additionally, re-running after a failed attempt shows `Workflow "" already exists` with an empty name

## Solution

- Extend `ADDITIONAL_INSTRUCTIONS` in `claude_client.ts` to explicitly tell Claude not to write the plan file and not to suggest follow-up commands — the CLI owns both responsibilities
- In `generate.ts`, validate that `--plan-file` points to an existing file **before** calling `generateWorkflow()` so no skeleton is created on a bad path
- Wrap the `buildWorkflow` call in a try/catch that removes the created `targetDir` on failure, ensuring a clean state for retry
- Fix `WorkflowExistsError` in `workflow_generator.ts` to pass `config.name` instead of `''`, producing a meaningful error message on re-run

Closes OUT-331, OUT-332, OUT-336

### DIFF
--- a/sdk/cli/src/commands/workflow/generate.ts
+++ b/sdk/cli/src/commands/workflow/generate.ts
@@ -6,7 +6,7 @@ import { getWorkflowGenerateSuccessMessage } from '#services/messages.js';
 import { DEFAULT_OUTPUT_DIRS } from '#utils/paths.js';
 import type { WorkflowGenerationResult } from '#types/generator.js';
 import path from 'node:path';
-import fs from 'node:fs/promises';
+import * as fsSync from 'node:fs';
 
 export default class Generate extends Command {
   static override description = 'Generate a new Output workflow from a skeleton or plan file';
@@ -62,14 +62,10 @@ export default class Generate extends Command {
     }
 
     const projectRoot = process.cwd();
+    const absolutePlanPath = planFile ? path.resolve( projectRoot, planFile ) : '';
 
-    if ( planFile ) {
-      const absolutePlanPath = path.resolve( projectRoot, planFile );
-      try {
-        await fs.access( absolutePlanPath );
-      } catch {
-        this.error( `Plan file not found: ${absolutePlanPath}` );
-      }
+    if ( planFile && !fsSync.existsSync( absolutePlanPath ) ) {
+      this.error( `Plan file not found: ${absolutePlanPath}` );
     }
 
     const result = await generateWorkflow( {
@@ -87,16 +83,14 @@ export default class Generate extends Command {
 
       await ensureOutputAISystem( projectRoot );
 
-      const absolutePlanPath = path.resolve( projectRoot, planFile );
+      const buildOutput = await buildWorkflow( absolutePlanPath, result.targetDir, args.name )
+        .catch( ( error: unknown ): never => {
+          fsSync.rmSync( result.targetDir, { recursive: true, force: true } );
+          const message = error instanceof Error ? error.message : String( error );
+          this.error( `Workflow implementation failed, created files have been rolled back: ${message}` );
+        } );
 
-      try {
-        const buildOutput = await buildWorkflow( absolutePlanPath, result.targetDir, args.name );
-        await buildWorkflowInteractiveLoop( buildOutput );
-      } catch ( error ) {
-        await fs.rm( result.targetDir, { recursive: true, force: true } );
-        const message = error instanceof Error ? error.message : String( error );
-        this.error( `Workflow implementation failed, created files have been rolled back: ${message}` );
-      }
+      await buildWorkflowInteractiveLoop( buildOutput );
 
       this.log( ux.colorize( 'green', '\nWorkflow implementation complete!\n' ) );
     }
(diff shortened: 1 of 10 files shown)


sdk/cli/src/services/workflow_builder.ts
```diff
@@ -76,7 +76,7 @@ async function processModification( modification: string, currentOutput: string
   }
 
   try {
-    const updatedOutput = await replyToClaude( modification, BUILD_COMMAND_OPTIONS );
+    const updatedOutput = await replyToClaude( modification, BUILD_COMMAND_OPTIONS, 'build' );
```
szanata: NIT: A this point named arguments would be easier to understand

sdk/cli/src/commands/workflow/generate.ts
```diff
@@ -59,44 +61,50 @@ export default class Generate extends Command {
...
 
-        await buildWorkflowInteractiveLoop( buildOutput );
+      const absolutePlanPath = path.resolve( projectRoot, planFile );
 
-        this.log( ux.colorize( 'green', '\nWorkflow implementation complete!\n' ) );
+      try {
+        const buildOutput = await buildWorkflow( absolutePlanPath, result.targetDir, args.name );
+        await buildWorkflowInteractiveLoop( buildOutput );
+      } catch ( error ) {
+        await fs.rm( result.targetDir, { recursive: true, force: true } );
```
szanata: `.rmSync`

sdk/cli/src/commands/workflow/generate.ts
```diff
@@ -59,44 +61,50 @@ export default class Generate extends Command {
...
-        force: flags.force
-      } );
+    const projectRoot = process.cwd();
 
-      if ( planFile ) {
-        this.log( '\nStarting AI-assisted workflow implementation...\n' );
+    if ( planFile ) {
+      const absolutePlanPath = path.resolve( projectRoot, planFile );
+      try {
+        await fs.access( absolutePlanPath );
```
szanata: Some cool think I didnt know:
> Using fs.access() to check for the accessibility of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file is not accessible.

Saying that. I think this should be `.accessSync`.

szanata (commented):
I think there is no reason to use async verion of fs for this. I think `accessSync()`, `readSync()`, `rmSync()` etc make much more sense for this use case. But this is non blocker.

sdk/cli/src/commands/workflow/generate.ts
```diff
@@ -59,44 +61,50 @@ export default class Generate extends Command {
...
-        );
+      await ensureOutputAISystem( projectRoot );
 
-        await buildWorkflowInteractiveLoop( buildOutput );
+      const absolutePlanPath = path.resolve( projectRoot, planFile );
 
-        this.log( ux.colorize( 'green', '\nWorkflow implementation complete!\n' ) );
+      try {
+        const buildOutput = await buildWorkflow( absolutePlanPath, result.targetDir, args.name );
+        await buildWorkflowInteractiveLoop( buildOutput );
```
catsby: nit: should we `try/catch` these 2 actions separately? If the workflow first gets created successfully but the user cancels out of the interactive loop, do we lose everything?

sdk/cli/src/commands/workflow/generate.ts
```diff
@@ -59,44 +61,50 @@ export default class Generate extends Command {
...
 
-        const buildOutput = await buildWorkflow(
-          absolutePlanPath,
-          result.targetDir,
-          args.name
-        );
+      await ensureOutputAISystem( projectRoot );
 
-        await buildWorkflowInteractiveLoop( buildOutput );
+      const absolutePlanPath = path.resolve( projectRoot, planFile );
```
catsby: nit: we have `const absolutePlanPath` here an on line 67, maybe extract it out of their respective `if ( planFile )` blocks so both can reference it?

catsby (commented):
some nits for now. verifying other things....

sdk/cli/src/services/workflow_builder.ts
catsby: nit: I know this code wasn't introduced or changed in this PR, however I think that `ux.error()` throws an error, so we'll just sort of crash here and never hit the `ux.stdout()` and `return currentOutput` lines. 

It looks like there's options that can control that though?

https://github.com/oclif/core/blob/30c128daf47ff9b6c45e7bde71f3efcfc57ff5f5/src/errors/error.ts#L9

catsby (approved):

sdk/cli/src/services/workflow_builder.ts
bnchrch: Good catch. Will change this right now.

PR #95, opened 2026-03-31
chore: Improving local dev startup

szanata:
- Adding "nodemon" on the `./test_workflows` project and updating `docker-compose.yaml` startup command to use it for hot reload;
- Refactoring root `package.json` startup commands designed for CLI so it uses `output` from `sdk/cli` instead of invoking the bin script directly;

## checklists
- [x] Starts with CLI
- [x] Starts with Docker
- [x] E2E tests working
- [x] Live reload working

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ agent-os/
 
 # Claude Things
 .claude/settings.local.json
+.claude/settings.json
 .claude/worktrees/
 
 # Pnpm publish report
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -46,7 +46,20 @@ services:
       - OUTPUT_TRACE_LOCAL_ON=true
       - OUTPUT_TRACE_HOST_PATH=${PWD}/test_workflows/logs
       - NODE_OPTIONS=${NODE_OPTIONS:---max-old-space-size=4096}
-    command: bash -c "corepack enable && npm run build:packages && cd test_workflows && npm run output:worker:build && exec npm run output:worker:start"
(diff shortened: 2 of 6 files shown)


szanata:
I cancelled e2e validation because it uses file from `main`, which breaks with these changes.

docker-compose.dev.yml
```diff
@@ -46,7 +46,20 @@ services:
       - OUTPUT_TRACE_LOCAL_ON=true
       - OUTPUT_TRACE_HOST_PATH=${PWD}/test_workflows/logs
       - NODE_OPTIONS=${NODE_OPTIONS:---max-old-space-size=4096}
-    command: bash -c "corepack enable && npm run build:packages && cd test_workflows && npm run output:worker:build && exec npm run output:worker:start"
+    command:
```
bnchrch, line 49: 📌  Worth noting this file isn't used in the test_e2e.sh script.

We invoke `output dev` to run the dev service which uses the cli's compose file

.gitignore
```diff
@@ -22,6 +22,7 @@ agent-os/
 
 # Claude Things
 .claude/settings.local.json
+.claude/settings.json
```
bnchrch, line 25: 💎 ty

bnchrch (approved):
So long is this will run in CI properly seems fine to me! Approved.

docker-compose.dev.yml
szanata: Yes, it is true. I tested both anyway. Thanks for pointing out

PR #97, opened 2026-03-31
fix(cli): use ink to ensure a correct re-render of dev panel  (OUT-341)

bnchrch:
## Problem

- The `output dev` panel used `log-update` + ANSI strings, whose diff algorithm left ghost lines when the panel shrank after a service recovered

## Solution

- Replaced `log-update` with [Ink](https://github.com/vadimdemedes/ink) — React components for the terminal — so re-renders are always correct

## Screenshot

image

image

Fixes [OUT-341](https://linear.app/growthx/issue/OUT-341/cli-text-overlaps-after-service-health-recovers)


sdk/cli/src/commands/dev/index.ts
```diff
@@ -191,41 +103,22 @@ export default class Dev extends Command {
...
+      const instance = render(
+        React.createElement( DevApp, { dockerComposePath, onCleanup: cleanup } ),
+        { exitOnCtrlC: false }
+      );
 
-      const services = await getServiceStatus( dockerComposePath );
-      this.log( getDevSuccessMessage( services ) );
+      const handleSignal = async () => {
+        await cleanup();
+        instance.unmount();
```
catsby, line 112: `instance.unmount();` will never run, because `cleanup()` calls `process.exit(0)`

https://github.com/growthxai/output/blob/7b8340cb143dec92fbf1dba04bbb0e868016ea7c/sdk/cli/src/commands/dev/index.ts#L170-L177

sdk/cli/src/commands/dev/index.ts
```diff
@@ -191,41 +103,22 @@ export default class Dev extends Command {
...
+      };
 
-      await this.pollServiceStatus( dockerComposePath );
+      process.on( 'SIGINT', handleSignal );
```
catsby, line 115: do we still need this handler?

sdk/cli/src/views/dev.tsx
```diff
@@ -0,0 +1,299 @@
...
+const resolveServiceDisplay = ( service: ServiceStatus ) => {
+  const key = service.health === SERVICE_HEALTH.NONE ? service.state : service.health;
+  return { icon: STATUS_ICONS[key] ?? '?', color: STATUS_COLORS[key] ?? 'white', status: key };
+};
+
+const fetchServices = async ( dockerComposePath: string ): Promise<ServiceStatus[] | null> => {
+  try {
+    return await getServiceStatus( dockerComposePath );
+  } catch {
+    return null;
```
catsby, line 44: should we log something of some kind here?

sdk/cli/src/views/dev.tsx
```diff
@@ -0,0 +1,299 @@
...
+};
+
+const useCtrlC = ( onCleanup: () => Promise<void> ): void => {
+  const { exit } = useApp();
+  const isExitingRef = useRef( false );
+
+  useInput( ( input, key ) => {
+    if ( key.ctrl && input === 'c' && !isExitingRef.current ) {
+      isExitingRef.current = true;
+      void onCleanup().then( () => exit() ).catch( () => exit() );
```
catsby: should we handle any errors here? I think the cleanups call `stopDockerCompose()` which maybe could fail and leave things running, in which case we'd just exit here without saying anything I think?

catsby (commented):
a few comments / questions

sdk/cli/package.json
```diff
@@ -39,16 +39,19 @@
...
-    "log-update": "7.2.0",
+    "react": "^19.2.0",
     "semver": "catalog:",
     "yaml": "^2.8.3"
   },
   "devDependencies": {
     "@types/cli-progress": "3.11.6",
     "@types/debug": "4.1.13",
     "@types/js-yaml": "catalog:",
+    "@types/react": "^19.1.0",
```
szanata: We should use fixed versions, not `^`.

bnchrch: @szanata what are we trying to prevent?

The lockfile would pin our install to the patch.

and we would only increase to subsequent patches if we update the lockfile.

or am I mistaken?

sdk/cli/src/commands/dev/index.ts
bnchrch: Yeah, so `useCtrlC` only catches raw terminal keypresses, not OS-level signals (e.g. `kill -INT`, CI runners, `docker stop`).

sdk/cli/src/views/dev.tsx
bnchrch: This is intentionally silent. Transient docker connectivity errors during startup are expected and logging every retry would spam the terminal.

sdk/cli/package.json
bnchrch: Closing the loop. Stefano proved to me that both dependabot and pnpm still can easily update to the latest patch relatively automatically. Fine to pin

szanata (approved):

sdk/cli/src/commands/dev/index.ts
```diff
@@ -191,41 +102,22 @@ export default class Dev extends Command {
         this.error( `Docker process error: ${getErrorMessage( error )}`, { exit: 1 } );
       } );
 
-      this.log( '⏳ Waiting for services to become healthy...\n' );
-      await waitForHealthy();
+      const instance = render(
+        React.createElement( DevApp, { dockerComposePath, onCleanup: cleanup } ),
+        { exitOnCtrlC: false }
+      );
```
catsby: I might be mistaken, but I believe `this.error()` in the `dockerProc.on()` listener (?) is oclif's and will exit with status 1 completely, and ignore the cleanup stuff Ink sets up on line 105.

I think we should rearrange and change it to something like this:

```typescript
const instance = render(
  React.createElement( DevApp, { dockerComposePath, onCleanup: cleanup } ),
  { exitOnCtrlC: false }
);

dockerProc.on( 'error', error => {
  // Let Ink handle the teardown, then oclif sees the error after waitUntilExit()
  instance.exit( new Error( `Docker process error: ${getErrorMessage( error )}` ) );
} );
```

catsby (commented):
I think we can remove the `log-update` mock here: 

https://github.com/growthxai/output/blob/7b8340cb143dec92fbf1dba04bbb0e868016ea7c/sdk/cli/src/services/docker.spec.ts#L11-L15

sdk/cli/src/commands/dev/index.ts
bnchrch: Good catch — moved the error handler after `instance` creation and swapped `this.error()` for `this.warn()` + cleanup + unmount, so Docker Compose is stopped and Ink tears down cleanly before the process exits.

catsby (approved):

PR #98, opened 2026-03-31
refactor(cli/dev): Renaming/refactoring worker commands 

szanata:
## Commands
Updated npm scripts in autogenerated `package.json` from @outputai/cli scaffold. Also aligned internal dev scripts.

### Added commands:
- `output:worker`: Executes install/build/start;
- `output:worker:watch`: Executes `output:worker` using nodemon to watch for changes (hot reload).

### Replaced commands:
- `dev` -> `output:dev`: Same pattern as other commands.

## Pinned dependency versions
Using pinned versions of the dependencies installed via the CLI-generated `package.json`.

## CLI Feature
Added new CLI feature "fix" (invoked via `output fix`), which realigns npm scripts at the host project's `package.json` with the canonical scripts from `@outputai/cli`:
- Removes legacy scripts (from previous versions of Output);
- Adds missing scripts (based on the CLI internal template);
- Replaces reserved scripts the user has customized (`output:` prefix).

image

_Execution example_

## Failures warning
Since these are breaking changes, projects created using older versions of the CLI will not start properly, so a warning was added to `output dev` when the worker fails to start:
Screenshot 2026-04-02 at 23 51 24

## Test checklist:
- [x] Test e2e
- [x] Run workflow via `npm run dev` + `npx --prefix="sdk/cli" output workflow run simple --input='{"values":[1,2,3]}'
- [x] Run workflow via `./run.sh dev` + `curl -X POST http://localhost:3001/workflow/run -H "Content-Type: application/json" -d '{"workflowName": "simple","taskQueue": "main","input": {"values": [1,2,3]}}'`


docs/guides/operations/deployment/render.mdx
```diff
@@ -70,7 +70,7 @@ ENV PATH="/app/node_modules/.bin:$PATH"
 # V8 heap = 80% of container memory (leaves room for Temporal Rust runtime + OS)
 ENV NODE_OPTIONS="--max-old-space-size-percentage=80 --heapsnapshot-signal=SIGUSR2"
 
-CMD ["output-worker"]
+CMD ["npm run outputai:start"]
```
catsby: Is this valid syntax? I thought the `CMD` will treat that as 1 binary command. Should it be something like `CMD ["npm", "run", "outputai:start"]`?

docs/guides/operations/deployment/railway.mdx
```diff
@@ -123,7 +123,7 @@ ENV PATH="/app/node_modules/.bin:$PATH"
 # V8 heap = 80% of container memory (leaves room for Temporal Rust runtime + OS)
 ENV NODE_OPTIONS="--max-old-space-size-percentage=80 --heapsnapshot-signal=SIGUSR2"
 
-CMD ["output-worker"]
+CMD ["npm run outputai:start"]
```
catsby: same `CMD` format question

docs/guides/operations/deployment/railway.mdx
catsby, line 222: we need to change this now, right?

```suggestion
    "startCommand": "npm run outputai:start",
```

catsby (commented):
I think we need to update the docker `CMD` statements?

sdk/cli/src/commands/reconfigure.ts
```diff
@@ -0,0 +1,71 @@
...
+import { getErrorMessage } from '#utils/error_utils.js';
+
+const Ansi: Record<string, string> = {
+  GREEN: '\x1b[32m',
+  RED: '\x1b[31m',
+  YELLOW: '\x1b[33m',
+  RESET: '\x1b[0m'
+};
+
+export default class Reconfigure extends Command {
```
bnchrch: 💅  `reconfigure` feels a bit alien to me. Havent seen it in alot of projects. WDYT about `config fix` or `config --fix`?

💡 or we alias this to `fix`?

Curious if it feels foreign to @catsby

sdk/cli/src/commands/fix.ts
```diff
@@ -0,0 +1,71 @@
...
+    try {
+      const plan = planReconfiguration( process.cwd() );
+      if ( !plan.hasChanges ) {
+        this.log( 'Nothing to change, package.json is already properly configured.' );
+        return;
+      }
+
+      this.printPlan( plan );
+
+      const shouldApply = await confirm( { message: 'Apply these changes to package.json?', default: true } );
```
bnchrch, line 55: 💎 Nice touch

sdk/cli/src/services/reconfigure_package.ts
```diff
@@ -0,0 +1,153 @@
...
+const packageJson = 'package.json';
+const templateRelativePath = path.join( 'templates', 'project', 'package.json.template' );
+
+/** Legacy script names from older versions of the project */
+export const legacyScripts = [
+  'worker:install',
+  'worker:build',
+  'worker:start',
+  'dev'
+] as const;
```
bnchrch: 💅 This should be a mapping of what scripts should map to the latest
💅  it feels like we should have the `output:` prefix here, to be explicit and in case in the future we add different namespaces

❓ Are these legacy? We still see them in the package.json
💡 Perhaps we have this structure inform whats in our package.json.template 

```suggestion
export const packageJsonScripts = {
  'output:worker:install': {
    defaultCommand: "npm install",
    legacyAliases: [],
  },
  'output:worker:build': {
    defaultCommand: "rm -rf dist/* && tsc && output-copy-assets",
    legacyAliases: [],
  },
  'output:worker:start': {
    defaultCommand: "output-worker",
    legacyAliases: [],
  },
  'output:dev': {
    defaultCommand: "output dev",
    legacyAliases: ['dev'],
  }
} as const;
```

sdk/cli/src/templates/project/package.json.template
```diff
@@ -8,16 +8,18 @@
     "output:worker:install": "npm install",
     "output:worker:build": "rm -rf dist/* && tsc -p ./ && copyfiles './src/**/*.prompt' './src/**/*.yml.enc' './src/**/*.key' dist -u 1",
     "output:worker:start": "output-worker",
-    "dev": "output dev"
```
bnchrch, line 11: 💅 We could leave the `dev` command in and have it point to output:dev

The benefit would be that
1. `npm run dev` is a very common expectation of js devs
2. We dont have to tell anyone to update how they start the project

❗ We need to update all of our docs and cli messages to reference `npm run output:dev` if we dont 
```
"dev": "npm run output:dev"
```

bnchrch (commented):
A few nits and one need to change

-----
**Legend**
- ❗Change requested
- ❓Question asked
- 📚 Documentation requested 
- 🧪 Test requested

sdk/cli/src/commands/reconfigure.ts
catsby: `reconfigure` does not feel right to me. `--fix` or `fix` are consistent with things like `eslint`, that sounds better to me

sdk/cli/src/services/reconfigure_package.ts
szanata: I dont like this structure, as it duplicated the code from package.json.

The idea of those scripts is to cleanup old versions of command we don't use anymore.

sdk/cli/src/templates/project/package.json.template
szanata: There isnt a single mention of `npm run dev` in the docs. The docs say `npx output dev`, and that is unchanged.

sdk/cli/src/services/reconfigure_package.ts
szanata: The list I provided there is wrong however. there was never `worker:*` commands, so those can be removed.

szanata: I narrowed down the list to just `dev`.

docs/guides/operations/deployment/railway.mdx
```diff
@@ -123,7 +123,7 @@ ENV PATH="/app/node_modules/.bin:$PATH"
 # V8 heap = 80% of container memory (leaves room for Temporal Rust runtime + OS)
 ENV NODE_OPTIONS="--max-old-space-size-percentage=80 --heapsnapshot-signal=SIGUSR2"
 
-CMD ["output-worker"]
+CMD ["npm", "run", "output:worker:start"]
```
catsby: This will set `npm` as `PID 1` and can interfere with our signal handling. `output:worker:start` just maps to `output-worker`, which is added to `PATH` around line 121, so it seems like this is unnecessary, what do you think?

test_workflows/package.json
```diff
@@ -7,7 +7,9 @@
   "scripts": {
     "output:worker:install": "corepack enable && CI=1 pnpm install --frozen-lockfile",
     "output:worker:build": "rm -rf dist/* && tsc && copyfiles './src/**/*.prompt' './src/**/*.yml.enc' './src/**/*.key' dist -u 1",
-    "output:worker:start": "output-worker"
+    "output:worker:start": "output-worker",
+    "output:worker": "npm run output:worker:install && npm run output:worker:build && npm run output:worker:start",
+    "output:worker:watch": "npx nodemon --watch src --watch package.json --ext ts,js,json,prompt --ignore 'dist/**' --ignore '**/*.test.*' --exec 'npm run output:worker'"
```
catsby, line 12:
```suggestion
    "output:worker:watch": "npx nodemon --watch src --watch package.json --ext ts,js,json,prompt --ignore 'dist/**' --ignore '**/*.test.*' --ignore "**/*.spec.ts" --exec 'npm run output:worker'"
```

sdk/cli/scripts/postinstall.js
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env node
+
+console.log( '\n────────────────────────────────────────────\n' );
+console.log( '\x1b[33;1mBREAKING CHANGES in @outputai/cli\n\x1b[0m' );
+console.log( 'NPM scripts generated by @outputai/cli changed in version \x1b[4m0.1.11\x1b[0m.' );
+console.log( 'Projects initialized on older versions may not start correctly in local dev mode' );
+console.log( 'Run \`\x1b[35;3mnpx output reconfigure\x1b[0m\` to reconcile the package.json.\x1b[0m\n' );
+console.log( '────────────────────────────────────────────\n' );
```
catsby: Is this ran everytime for every install? If a user does a fresh install having never installed `output` before, will they see this?

sdk/cli/src/commands/reconfigure.ts
```diff
@@ -0,0 +1,73 @@
...
+      if ( !shouldApply ) {
+        this.log( 'Cancelled.' );
+        return;
+      }
+
+      applyReconfiguration( plan );
+      this.log( 'Done, package.json is properly configured.' );
+    } catch ( error: unknown ) {
+      if ( error instanceof Error && error.name === 'ExitPromptError' ) {
+        this.log( 'Cancelled.' );
```
catsby: Should this be `this.error` instead? It's an error scenario, but we only call `log()`. What would the exit code be here, `0`?

sdk/cli/src/services/reconfigure_package.ts
```diff
@@ -0,0 +1,150 @@
...
+    throw error;
+  }
+}
+
+function parsePackageJsonObject( raw: string, packagePath: string ): Record<string, unknown> {
+  try {
+    return JSON.parse( raw ) as Record<string, unknown>;
+  } catch {
+    throw new Error( `${packagePath} is not valid JSON.` );
+  }
```
catsby: should we catch the error from `JSON.parse` and use it in the error message? Maybe it contains specifics on line number where the parse error occured?

catsby (commented):
I think we need to rename `reconfigure` to `fix` (can use `reconfigure` as an alias), and change the guides or templates back to using `CMD ["output-worker"]`

sdk/cli/src/services/reconfigure_package.ts
bnchrch: Ah so the idea I was getting at was having one source of truth for both the commands and their past aliases

1. we have a package.json.template that contains the output commands
1. we have code that reads a package.json and updates the output commands to their latest defaults

They should have the same source of truth

e.g. package.json.template should look something like this
```
{
  "name": "{{projectName}}",
  "version": "0.0.1",
  "description": "{{description}}",
  "type": "module",
  "main": "dist/worker.js",
  "scripts": {
    {{#each packageJsonScripts}}
    "{{@key}}": "{{this.defaultCommand}}"{{#unless @last}},{{/unless}}
    {{/each}}
  },
  "dependencies": {
    "@outputai/output": "^{{frameworkVersion}}"
  },
  "devDependencies": {
    "@types/node": "^24.0.0",
    "nodemon": "3.1.0",
    "typescript": "^5.7.0"
  },
  "engines": {
    "node": ">=24.3.0"
  },
  "output": {
    "hookFiles": ["node_modules/@outputai/credentials/dist/hooks.js"]
  }
}
```

szanata: I don't necessarily disagree, but I think this is being the scope of this PR.
Not all legacy commands will be alias of new commands, and the solution as implement works for now.
Modifying the template would require it to be computed with the env vars when scaffolding as well, and this involves changing much more stuff.

docs/guides/operations/deployment/railway.mdx
szanata: I think it is better to use the scripts, so we have more flexibility on how the package starts.

test_workflows/package.json
szanata: Thanks

sdk/cli/scripts/postinstall.js
szanata: It is impossible to know fresh start vs updates. As I mentioned in our thread, this can be dropped in a couple updates.

szanata: It is common normal to have this info being displayed before/after big releases

sdk/cli/src/commands/reconfigure.ts
szanata: `ExitPromptError` means user used `Ctrl/Cmd + C` to cancel the execution. 
import { ExitPromptError } from '@inquirer/core';

sdk/cli/src/services/reconfigure_package.ts
szanata: Agreed. I assigned the original error to the new Error's `cause` field. This is the proper way to propagate.

sdk/cli/src/commands/reconfigure.ts
szanata: As discussed, this was renamed to `fix`

sdk/cli/src/services/reconfigure_package.ts
szanata: As discussed, this will not be changes in this PR.

docs/guides/operations/deployment/railway.mdx
szanata: Reverted for the sake of not introducing this issue back.

sdk/cli/scripts/postinstall.js
bnchrch: ❗ I think I misunderstood the conversation earlier.

Lets remove this. I would rather our internal users wonder what happened with a breaking change

then everyone at launch next week seeing this message.

Alternatively lets just update the service failure message in `sdk/cli/src/views/dev.tsx`


```diff
- <Text dimColor>{'Check the logs with: docker compose logs worker'}</Text>
+ <Text dimColor>{'Check the logs with: docker compose logs worker'}</Text>
+ <Text dimColor>Or run <Text color="cyan">output fix</Text></Text>
```

bnchrch (approved):
Once this is addressed. I think its fine to go in.
https://github.com/growthxai/output/pull/98/changes#r3030425123
Pre-approving

PR #102, opened 2026-04-03
feat(core): add workflow aliases for backward-compatible renaming

bradherman:
# Problem

I ran into an inconvenience earlier where I wanted to rename a workflow.  If there are callers to the workflow in the wild, you need a multi-step process coordinated across your project and the callers and must duplicate the workflow for a period to avoid errors.

# Solution

Allow workflows to declare alternative names via an `aliases` property, eliminating the 3-deploy rename process. Aliases are registered as additional Temporal workflow types and resolved to the canonical name by the API, keeping execution history consistent.

### DIFF
--- a/api/src/clients/temporal_client.js
+++ b/api/src/clients/temporal_client.js
@@ -176,14 +176,15 @@ export default {
 
         // the catalog worker has the same name of the task queue
         const catalog = await getCatalog( { client, taskQueue } );
-        const workflowExists = catalog.workflows.some( w => w.name === workflowName );
-        if ( !workflowExists ) {
+        const resolved = catalog.workflows.find( w => w.name === workflowName || w.aliases?.includes( workflowName ) );
+        if ( !resolved ) {
           throw new WorkflowNotFoundError( `Workflow "${workflowName}" is not available at worker "${taskQueue}"` );
         }
 
+        const resolvedName = resolved.name;
         const workflowId = userWorkflowId ?? buildWorkflowId();
         const executionTimeout = timeout ?? workflowExecutionMaxWaiting;
-        const handle = await client.workflow.start( workflowName, { args: [ input ], taskQueue, workflowId, workflowExecutionTimeout } );
+        const handle = await client.workflow.start( resolvedName, { args: [ input ], taskQueue, workflowId, workflowExecutionTimeout } );
 
         try {
           const result = await Promise.race( [
--- a/api/src/clients/temporal_client.js
+++ b/api/src/clients/temporal_client.js
@@ -151,6 +151,26 @@ export default {
 
     logger.info( 'Temporal client connected', { address, namespace } );
 
+    /**
+     * Resolve a workflow name (or alias) to the canonical workflow name via the catalog.
+     *
+     * @param {object} catalog - The catalog object
+     * @param {string} workflowName - The workflow name or alias
+     * @param {string} taskQueue - The task queue (for error messages)
+     * @returns {string} The canonical workflow name
+     * @throws {WorkflowNotFoundError}
+     */
+    const resolveWorkflowName = ( catalog, workflowName, taskQueue ) => {
+      const resolved = catalog.workflows.find( w => w.name === workflowName || w.aliases?.includes( workflowName ) );
+      if ( !resolved ) {
+        throw new WorkflowNotFoundError( `Workflow "${workflowName}" is not available at worker "${taskQueue}"` );
+      }
+      if ( resolved.name !== workflowName ) {
+        logger.info( 'Workflow alias resolved', { alias: workflowName, resolvedName: resolved.name, taskQueue } );
+      }
+      return resolved.name;
+    };
+
     return {
       /**
        * Workflow execution result
@@ -176,12 +196,8 @@ export default {
 
         // the catalog worker has the same name of the task queue
         const catalog = await getCatalog( { client, taskQueue } );
-        const resolved = catalog.workflows.find( w => w.name === workflowName || w.aliases?.includes( workflowName ) );
-        if ( !resolved ) {
-          throw new WorkflowNotFoundError( `Workflow "${workflowName}" is not available at worker "${taskQueue}"` );
-        }
+        const resolvedName = resolveWorkflowName( catalog, workflowName, taskQueue );
 
-        const resolvedName = resolved.name;
         const workflowId = userWorkflowId ?? buildWorkflowId();
         const executionTimeout = timeout ?? workflowExecutionMaxWaiting;
         const handle = await client.workflow.start( resolvedName, { args: [ input ], taskQueue, workflowId, workflowExecutionTimeout } );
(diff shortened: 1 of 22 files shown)


api/src/clients/temporal_client.js
```diff
@@ -176,14 +176,15 @@ export default {
...
+        const resolved = catalog.workflows.find( w => w.name === workflowName || w.aliases?.includes( workflowName ) );
+        if ( !resolved ) {
           throw new WorkflowNotFoundError( `Workflow "${workflowName}" is not available at worker "${taskQueue}"` );
         }
 
+        const resolvedName = resolved.name;
         const workflowId = userWorkflowId ?? buildWorkflowId();
         const executionTimeout = timeout ?? workflowExecutionMaxWaiting;
-        const handle = await client.workflow.start( workflowName, { args: [ input ], taskQueue, workflowId, workflowExecutionTimeout } );
+        const handle = await client.workflow.start( resolvedName, { args: [ input ], taskQueue, workflowId, workflowExecutionTimeout } );
```
catsby, line 203: I think we need changes like this also added to `startWorkflow`, around line 233 in this file

api/src/clients/temporal_client.js
```diff
@@ -176,14 +176,15 @@ export default {
...
         // the catalog worker has the same name of the task queue
         const catalog = await getCatalog( { client, taskQueue } );
-        const workflowExists = catalog.workflows.some( w => w.name === workflowName );
-        if ( !workflowExists ) {
+        const resolved = catalog.workflows.find( w => w.name === workflowName || w.aliases?.includes( workflowName ) );
+        if ( !resolved ) {
           throw new WorkflowNotFoundError( `Workflow "${workflowName}" is not available at worker "${taskQueue}"` );
         }
 
+        const resolvedName = resolved.name;
```
catsby: Alias support is a great idea, but I think we might want to log when `workflowName !== resolvedName`, otherwise we don't have much insight into what workflows are continuously resolved via the alias. Since backwards compat. is a big part of this feature, I think we should log so operators have some confidence that workflows are no longer getting executed by their old alias. What do you think?

sdk/core/src/interface/validations/static.js
```diff
@@ -64,6 +64,7 @@ const baseSchema = z.strictObject( {
 const stepSchema = baseSchema;
 
 const workflowSchema = baseSchema.extend( {
+  aliases: z.array( z.string().regex( /^[a-z_][a-z0-9_]*$/i ) ).optional().default( [] ),
```
catsby, line 67: If I read this right, which I often don't read regex correctly, this should allow mixed case to both be valid, e.g. `SimpleV1` and `simplev1`. I think the `validateAliasUniqueness()` function only validates actual equality though. Will this cause a conflict?

catsby (commented):
Thanks @bradherman ! I think this is a great feature to add. I have a few thoughts to round this out, please see my comments and let me know what you think.

catsby:
Another 2 things I just thought about:

- we might want to add the aliases to the workflows map we build up:

https://github.com/growthxai/output/blob/76bcede9751462a31594c94bf014533b1aa5d09d/sdk/core/src/worker/interceptors/activity.js#L36

- the aliases might mess with how we record workflow names in our tracing:

https://github.com/growthxai/output/blob/76bcede9751462a31594c94bf014533b1aa5d09d/sdk/core/src/worker/sinks.js#L14

Don't feel pressured to solve those yourself though, I'm just calling them out

api/src/clients/temporal_client.js
bradherman: yeah i think that makes a lot of sense -- right now you'd just kind of hope you were right on sunsetting the old workflow

sdk/core/src/interface/validations/static.js
bradherman: I think you're right as well -- I relied on claude to write this regex, which I assume is better at regex than me, haha

But yes, they should probably both match

bradherman:
@catsby i addressed your comments

sdk/core/src/worker/interceptors/activity.js
```diff
@@ -25,15 +25,26 @@ import { messageBus } from '#bus';
...
     const startDate = Date.now();
     const { workflowExecution: { workflowId }, activityId: id, activityType: name, workflowType: workflowName } = Context.current().info;
     const { executionContext } = headersToObject( input.headers );
     const { type: kind } = this.activities?.[name]?.[METADATA_ACCESS_SYMBOL];
-    const workflowFilename = this.workflowsMap.get( workflowName ).path;
+    const workflowEntry = this.workflowsMap.get( workflowName );
+    if ( !workflowEntry ) {
+      const availableWorkflows = [ ...this.workflowsMap.keys() ].join( ', ' );
+      throw new Error( `Activity interceptor: workflow "${workflowName}" not found in workflowsMap. Available: [${availableWorkflows}]` );
+    }
```
szanata, line 50: I think this validation isn't necessary since all possible aliases are at the workflowMap and this is validated via the API as well.

sdk/core/src/worker/loader.js
```diff
@@ -132,18 +132,55 @@ export async function loadHooks( rootDir ) {
...
+    allNames.set( name.toLowerCase(), `workflow "${name}"` );
+  }
+
+  // Check the reserved catalog name
+  allNames.set( WORKFLOW_CATALOG.toLowerCase(), 'system workflow "$catalog"' );
+
+  // Check aliases against all names
+  for ( const { name, aliases = [] } of workflows ) {
+    for ( const alias of aliases ) {
+      if ( alias.toLowerCase() === name.toLowerCase() ) {
```
szanata: Minor improvements:
```js
  const lowerCaseName = name.toLowerCase();
  for ( const alias of aliases ) {
      const lowerCaseAlias = alias.toLowerCase();
      if ( lowerCaseAlias ===  lowerCaseName ) {
...
```

`toLowerCase()` is cheap, but there is no reason to not reuse here.

sdk/core/src/worker/loader.js
```diff
@@ -132,18 +132,55 @@ export async function loadHooks( rootDir ) {
   }
 };
 
+/**
+ * Validates that all workflow names and aliases are unique across the project.
+ *
+ * @param {object[]} workflows
+ * @throws {Error} If any alias conflicts with a workflow name or another alias
+ */
+function validateAliasUniqueness( workflows ) {
```
szanata: I would rename this to `validateWorkflowNames` or `validateWorkflowNameUniqueness`. Alias is kind of abstract without referring to the word workflow.

sdk/core/src/worker/interceptors/activity.js
szanata: If this error is thrown the trace file gets incomplete, because there will be an `error` event without an `start` event.

This has to be move after the `messageBus.emit( BusEventType.ACTIVITY_START...` line.

sdk/core/src/worker/loader.js
catsby: renamed in 647942c

sdk/core/src/worker/loader.js
catsby: updated in 647942c

szanata (changes requested):

sdk/core/src/worker/interceptors/activity.js
catsby: I moved the possible `throw` to after we emit the activity start in 0d70dd4

szanata (approved):

PR #107, opened 2026-04-06
refactor(http): Migrating to ky v2

szanata:
Migrating to ky v2 in the http library. This is a breaking change.

Ky v2 offserts significant advantages and fixes some issues we had in the past. Release notes [here](https://github.com/sindresorhus/ky/releases).

### DIFF
--- a/.changeset/deep-parts-drive.md
+++ b/.changeset/deep-parts-drive.md
@@ -1,0 +1,7 @@
+---
+"@outputai/core": patch
+"@outputai/http": patch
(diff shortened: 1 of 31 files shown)


szanata:
Instead of upgrading `ky`, we are going to provide a instrumented `fetch` (https://github.com/growthxai/output/pull/118) and change scaffolding to use `ky` with the custom `fetch` instead of `@outputai/http`.

`@outputai/http` will be retired eventually.

PR #111, opened 2026-04-07
feat: Removing /output wrapper project from CLI package.json template

szanata:
## Summary
- Dropped `@outputai/output` project.
- Updated `output init` to installs all packages directly.
- Updated `output update` to check each package version and update them individually.
- Added a warning about using the deprecated package in `output init`, `output migrate` and `output dev`.

## Reasoning
When installing the SDK modules via a third party dependency (`@outputai/output`), they all become indirect dependencies (only that is actually present in the `package.json`).

So, when the code imports "core" or "http", it is actually importing dependencies of a dependency, which is an anti-pattern in NPM and package managing in general.


sdk/cli/src/services/npm_update_service.ts
```diff
@@ -32,18 +57,42 @@ function findVersionInTree( deps: Record<string, any> | undefined ): string | nu
...
+      packageJsonDeps.set( name, { version, dependencyType: 'dependencies' } );
+    }
+
+    for ( const [ name, version ] of Object.entries( pkg.devDependencies ?? {} ) ) {
+      packageJsonDeps.set( name, { version, dependencyType: 'devDependencies' } );
+    }
+
+    return packageJsonDeps;
+  } catch ( error ) {
+    debug( 'Failed to read local package.json: %O', error );
```
catsby, line 88: I think this conflates a corrupt or unreadable file (invalid/broken structure or file perms issue) with "file doesn't exist". The "doesn't exist" path seems fine, there's no warnings, but a corrupt file also shows no issues unless you have the debug flag on. 

Build the packages locally then create a corrupt `package.json` and try it:

```
❯ printf '%s\n' '{ "name": "demo", "dependencies": { "@outputai/output": "^0.9.0" } }' > package.json

❯ cat package.json| jq
jq: parse error: Expected another key-value pair at line 1, column 67

❯ node sdk/cli/bin/run.js update --cli

Latest Output SDK version: v0.9.0

Global install: v0.9.0 (up to date)
```

with debug on:

```
❯ DEBUG=output-cli:npm-update node sdk/cli/bin/run.js update --cli

Latest Output SDK version: v0.9.0

Global install: v0.9.0 (up to date)
  output-cli:npm-update Failed to read local package.json: SyntaxError: Expected double-quoted property name in JSON at position 66 (line 1 column 67)
  output-cli:npm-update     at JSON.parse (<anonymous>)
  output-cli:npm-update     at readPackageJsonDependencies (file:///Users/clint/go/github.com/growthxai/output/.claude/worktrees/chore/remove_output_wrapper/sdk/cli/dist/services/npm_update_service.js:51:26)
```

sdk/cli/src/services/npm_update_service.ts
```diff
@@ -94,13 +158,41 @@ function spawnInherit( command: string, args: string[], cwd?: string ): Promise<
...
+    await spawnInherit( 'npm', [ 'install', '--ignore-scripts', '--save-dev', '--save-exact', ...devDependencies ], cwd );
+  }
 }
 
 export function isOutdated( current: string, latest: string ): boolean {
   return semver.lt( current, latest );
 }
+
+export function isPackageJsonVersionOutdated( version: string, latest: string ): boolean {
+  const range = semver.validRange( version, { includePrerelease: true } );
```
catsby, line 191: does `validRange()` handle `latest` etc?

sdk/cli/src/commands/update.ts
```diff
@@ -150,4 +168,55 @@ export default class Update extends Command {
...
+          this.warn( `Some Output SDK packages are still behind v${latest}: ${staleNames}` );
+        }
+      } else {
+        this.log( '\nLocal update completed (could not verify new versions)' );
+      }
+
+      return true;
+    } catch ( error: unknown ) {
+      this.warn( `Failed to update local install: ${getErrorMessage( error )}` );
+      return false;
```
catsby, line 219: should we throw here? looking at the chain, it looks like we warn that we failed to install something but eventually exit with `0` status code in the CLI

sdk/cli/src/commands/update.ts
```diff
@@ -150,4 +168,55 @@ export default class Update extends Command {
...
+      if ( newLocalPackages.length > 0 ) {
+        this.log( `\nLocal Output SDK packages updated to v${latest}` );
+
+        const stalePackages = newLocalPackages.filter( pkg => isPackageJsonVersionOutdated( pkg.version, latest ) );
+        if ( stalePackages.length > 0 ) {
+          const staleNames = stalePackages.map( pkg => `${pkg.name}@${pkg.version}` ).join( ', ' );
+          this.warn( `Some Output SDK packages are still behind v${latest}: ${staleNames}` );
+        }
+      } else {
+        this.log( '\nLocal update completed (could not verify new versions)' );
```
catsby, line 213: should we throw here? we log that we couldn't verify anything but then exit `0`?

sdk/cli/src/services/npm_update_service.ts
```diff
@@ -94,13 +158,41 @@ function spawnInherit( command: string, args: string[], cwd?: string ): Promise<
...
+    .map( pkg => `${pkg.name}@${version}` );
+
+  if ( dependencies.length > 0 ) {
+    await spawnInherit( 'npm', [ 'install', '--ignore-scripts', '--save-exact', ...dependencies ], cwd );
+  }
+
+  if ( devDependencies.length > 0 ) {
+    await spawnInherit( 'npm', [ 'install', '--ignore-scripts', '--save-dev', '--save-exact', ...devDependencies ], cwd );
+  }
 }
```
catsby, line 184: what is the difference between `updateLocal()` and `updateLocalPackages()` such that we need/want both?

sdk/cli/src/services/npm_update_service.ts
```diff
@@ -7,23 +9,46 @@ import packageJson from '../../package.json' with { type: 'json' };
...
+  '@outputai/llm',
+  '@outputai/credentials',
+  '@outputai/evals'
+] as const;
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-function findVersionInTree( deps: Record<string, any> | undefined ): string | null {
+export interface LocalSdkPackage {
+  name: string;
```
catsby, line 31: nit: This can only ever be something from `LOCAL_SDK_PACKAGE_NAMES`, right? Should or could we tighten it from `string`?

```ts
name: typeof LOCAL_SDK_PACKAGE_NAMES[number];   // '@outputai/cli' | '@outputai/core' | ...
```

szanata:
I am cancelling this, as in a future release all modules will be unified under Core, except CLI.

PR #116, opened 2026-04-08
feat(claude): enhance eval skills

bnchrch:
## Overview

- Add 5 new eval methodology skills (`output-eval-error-analysis`, `output-eval-judge-prompt`, `output-eval-dataset-design`, `output-eval-validate-judge`, `output-eval-audit`) inspired by [hamelsmu/evals-skills](https://github.com/hamelsmu/evals-skills)
- Teach the *process* of building trustworthy evals (when/why/how to design evaluators) complementing the existing API reference skills
- Cover the full eval lifecycle: trace review, judge prompt design, dataset generation, judge validation, and suite auditing
- Add cross-references from 3 existing skills (`output-dev-eval-testing`, `output-dev-evaluator-function`, `output-dev-prompt-file`) to the new methodology skills
- All skills are directive, under 500 lines, and use Output.ai's actual APIs and CLI commands in examples

### DIFF
--- a/coding_assistants/claude/plugins/outputai/skills/output-eval-judge-prompt/SKILL.md
+++ b/coding_assistants/claude/plugins/outputai/skills/output-eval-judge-prompt/SKILL.md
@@ -1,0 +1,335 @@
+---
+name: output-eval-judge-prompt
+description: Design effective LLM judge .prompt files for evaluators. Use when creating judgeVerdict/judgeScore/judgeLabel prompts, or when existing judges produce unreliable results.
+allowed-tools: [Read, Write, Edit]
+---
+
+# Designing LLM Judge Prompts
+
+## Overview
+
+An LLM judge evaluates workflow output for a **single, specific failure mode** identified during error analysis. This skill covers how to design the `.prompt` file that powers `judgeVerdict()`, `judgeScore()`, or `judgeLabel()` calls. For the file format basics, see `output-dev-prompt-file`. For error analysis, see `output-eval-error-analysis`.
+
+## Prerequisites
+
+Before writing a judge prompt:
+
+1. **Error analysis is complete** — You have identified the specific failure mode this judge targets (from `output-eval-error-analysis`)
+2. **20+ labeled examples** — At least 20 pass and 20 fail traces for this failure mode, with `ground_truth` labels in dataset YAML files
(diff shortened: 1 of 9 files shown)


szanata (approved):

coding_assistants/claude/plugins/outputai/skills/output-eval-judge-prompt/SKILL.md
```diff
@@ -0,0 +1,335 @@
...
+### 4. Structured Output
+
+Request JSON output with **critique before verdict**. This forces the judge to reason before deciding, which improves accuracy.
+
+```json
+{
+  "critique": "Detailed analysis of the output against the criterion...",
+  "verdict": "pass"
+}
+```
```
catsby, line 74: assuming these are ran though AI-SDK, then we shouldn't dictate the output structure in the prompt, instead that should be done in the `generateText()` call with output schema, etc. For example I don't think this output schema here matches the typical eval judgement schema we define here:

https://github.com/growthxai/output/blob/9118b841793aa177af23cf8ef7d84a555382ac33/sdk/evals/src/judge.ts#L17

coding_assistants/claude/plugins/outputai/skills/output-eval-judge-prompt/SKILL.md
```diff
@@ -0,0 +1,335 @@
...
+- 0.7-0.8: Good — well-structured, clear, minor issues only
+- 0.9-1.0: Excellent — polished, engaging, publication-ready
+
+## Output Format
+
+Return a JSON object with:
+- "critique": Detailed analysis of quality strengths and weaknesses (3-5 sentences)
+- "score": A number between 0.0 and 1.0
+</system>
+
```
catsby, line 238: same output schema comment as above

coding_assistants/claude/plugins/outputai/skills/output-eval-judge-prompt/SKILL.md
bnchrch: So this is where Im not sure this rule applies! This best practise is saying its good to instruct the agent in which order to output fields.

That would likely not translate directly to the order you put your jsonschema in

So this instruction is likely still needed to get the improvement in output quality it advertises

PR #117, opened 2026-04-08
feat(cli): add workflow run info to dev command

bnchrch:
## Overview

- Add live workflow summary to the dev running view showing counts of running, failed, and completed workflows
- Press `o` to open Temporal UI in the browser, `w` to open an interactive workflow list
- Browse workflow runs with arrow keys; selected workflow shows its input and output
- Press `o` on any workflow in the list to open it directly in the Temporal UI
- New `WorkflowListView` component with status-colored rows, scrolling window, and cached detail fetching

## Screenshots
image

image

image

image

### DIFF
--- a/sdk/cli/src/components/status_icon.tsx
+++ b/sdk/cli/src/components/status_icon.tsx
@@ -15,7 +15,7 @@ const STATUS_MAP: Record<string, StatusDisplay> = {
   exited: { icon: '✗', color: 'red' },
 
   // Workflow run status
-  running: { icon: '◐', color: 'yellow' },
+  running: { icon: '●', color: 'blue' },
   completed: { icon: '●', color: 'green' },
   failed: { icon: '✗', color: 'red' },
   canceled: { icon: '○', color: 'gray' },
--- a/sdk/cli/src/components/status_icon.tsx
+++ b/sdk/cli/src/components/status_icon.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import { Text } from 'ink';
+
+interface StatusDisplay {
+  icon: string;
+  color: string;
+}
+
+const STATUS_MAP: Record<string, StatusDisplay> = {
+  // Docker service health
+  healthy: { icon: '●', color: 'green' },
+  unhealthy: { icon: '○', color: 'red' },
+  starting: { icon: '◐', color: 'yellow' },
+  none: { icon: '●', color: 'blue' },
+  exited: { icon: '✗', color: 'red' },
+
+  // Workflow run status
+  running: { icon: '◐', color: 'yellow' },
+  completed: { icon: '●', color: 'green' },
+  failed: { icon: '✗', color: 'red' },
+  canceled: { icon: '○', color: 'gray' },
+  terminated: { icon: '✗', color: 'red' },
+  timed_out: { icon: '✗', color: 'red' },
+  continued: { icon: '↻', color: 'blue' }
+};
+
+const DEFAULT_DISPLAY: StatusDisplay = { icon: '?', color: 'white' };
+
+export const resolveStatus = ( status: string ): StatusDisplay =>
+  STATUS_MAP[status] ?? DEFAULT_DISPLAY;
+
+export const statusColor = ( status: string ): string =>
+  resolveStatus( status ).color;
+
+export const StatusIcon: React.FC<{ status: string }> = ( { status } ) => {
+  const { icon, color } = resolveStatus( status );
+  return <Text color={color}>{icon}</Text>;
+};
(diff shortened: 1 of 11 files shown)


sdk/cli/src/views/workflow/list.tsx
```diff
@@ -0,0 +1,252 @@
...
+  timed_out: 2,
+  terminated: 3,
+  canceled: 4,
+  continued: 5,
+  completed: 6
+};
+
+const sortRuns = ( runs: WorkflowRun[] ): WorkflowRun[] =>
+  [ ...runs ].sort( ( a, b ) => {
+    const statusDiff = ( STATUS_ORDER[a.status ?? ''] ?? 99 ) - ( STATUS_ORDER[b.status ?? ''] ?? 99 );
```
szanata: If 99 means the absolute last one, it should be `Infinity`.

szanata (approved):
Cool feature and works fine.

My only suggestion would be to maybe have a separate view for the workflows runs, to not make the `dev.tsx` too complex, but this is up to debate as well.

sdk/cli/src/components/status_icon.tsx
```diff
@@ -0,0 +1,38 @@
...
+const STATUS_MAP: Record<string, StatusDisplay> = {
+  // Docker service health
+  healthy: { icon: '●', color: 'green' },
+  unhealthy: { icon: '○', color: 'red' },
+  starting: { icon: '◐', color: 'yellow' },
+  none: { icon: '●', color: 'blue' },
+  exited: { icon: '✗', color: 'red' },
+
+  // Workflow run status
+  running: { icon: '◐', color: 'yellow' },
```
catsby: nit: it seems weird that 'running' is yellow with a half-circle icon, when that's the ideal/good state.

sdk/cli/src/utils/open_url.ts
```diff
@@ -0,0 +1,5 @@
+import { execFile } from 'node:child_process';
+
+export const openUrl = ( url: string ): void => {
+  execFile( 'open', [ url ] );
```
catsby: nit: I think `open` is macOS specific

sdk/cli/src/utils/date_formatter.ts
```diff
@@ -45,6 +45,33 @@ export function formatDate( isoString: string | null | undefined ): string {
...
  * @returns Human-readable duration string (e.g., "5 seconds", "running")
  */
+/**
+ * Format a duration in milliseconds to a compact string that fits in narrow columns.
+ * Always returns a short single-token string (e.g., "150ms", "7.56s", "24.2m", "1.3h").
+ */
+/**
+ * Calculate elapsed milliseconds between two ISO timestamps.
+ * If completedAt is null/undefined, uses current time (for in-progress durations).
+ */
```
catsby: these JSDoc blocks look like they need to be split up

sdk/cli/src/views/workflow/list.tsx
```diff
@@ -0,0 +1,252 @@
...
+        <Text bold>Status: </Text>
+        <Text color={statusColor( detail.status ?? '' )}>{detail.status}</Text>
+      </Box>
+      {detail.error && (
+        <Box marginTop={1} flexDirection="column">
+          <Text bold color="red">Error:</Text>
+          <Text color="red">{truncate( detail.error, 300 )}</Text>
+        </Box>
+      )}
+      {detail.output !== undefined && detail.output !== null && (
```
catsby, line 107: super minor nit
```suggestion
      {detail.output != null && (
```

catsby (commented):
Overall love this! I left some minor comments, I'm not sure how big of a deal it is if the `open` is in fact macOS specific, but we should find out.

sdk/cli/src/views/workflow/list.tsx
bnchrch: Addressed

sdk/cli/src/views/workflow/list.tsx
bnchrch: Unfortunately this is part of our linting rules to protect against JS's silly inclusion of multiple null types

sdk/cli/src/views/dev.tsx
```diff
@@ -296,16 +340,41 @@ export const DevApp: React.FC<{
...
       </Static>
-      {phase === 'waiting' && <WaitingView services={services} />}
-      {phase === 'running' && <RunningView services={services} />}
-      {phase === 'failed' && <RunningView services={services} />}
+      <Box flexDirection="column">
+        {activeView === 'main' && (
+          <MainDevView phase={phase} services={services} workflowSummary={workflowSummary} />
+        )}
+        {activeView === 'workflows' && (
+          <WorkflowListView runs={workflowRuns} onBack={() => setActiveView( 'main' )} />
```
bnchrch, line 375: @szanata Replying to your comment on the workflow run being a separate view.

It was already a separate view, I just cleaned up the code to make it much more obvious and extensible.

Let me know if this addresses what you meant 🙂

sdk/cli/src/components/status_icon.tsx
bnchrch: Updated it to blue!

sdk/cli/src/utils/open_url.ts
bnchrch: Added support for linux and windows

catsby (approved):

PR #118, opened 2026-04-08
feat(http): Adding instrumented fetch

szanata:
## Features 

1. Adding a spec-compliant, instrumented fetch function to `@outputai/http`, also called `fetch`:

    Usage will be the same as using `fetch` from _Node_ or _undici_ (but with actual types from _undici_):
    ```js
    import { fetch } from '@outputai/http';
    
    const response = await fetch( 'https://output.ai', { headers: { 'X-Foo': 'bar' } } );
    
    if ( response.ok ) {
      console.log( 'Alright!' );
    }
    ```
    
    Function is compliant with [Fetch Spec from WHATWG](https://fetch.spec.whatwg.org/).
    
    No learning curve required, as this is the same usage found on Node, browsers, etc.

2. Also adding the following exports to `@outputai/http`:
    - `undici`: The whole undici library, as a convenience, this includes types;
    - `RequestInfo` type: The type of the first argument of the `fetch` function. It is just a re-export from `undici` and can also be used from`undici.RequestInfo`.
    - `RequestInit` type: The type of the second argument of the `fetch` function. It is just a re-export from `undici` and can also be used from`undici.RequestInit`.

3. Refactoring the `httpClient` created with `ky` to use the instrumented `fetch` and dropping all hooks and custom code around `ky`.

## Reasoning

The current implementation of this library uses __ky__ as the HTTP client, and that just got a major change for version 2, so the signature of the library methods need to be updated, causing a breaking change for dependents. Nothing prevents this breaking change of occurring on v3, v4, etc.

The new function is the first spec into providing an fetch-based instrumentation to HTTP calls, allowing flexible choice of HTTP clients and the removal of __ky__ from this module.


sdk/core/package.json
```diff
@@ -47,7 +47,7 @@
     "@temporalio/workflow": "catalog:",
     "redis": "5.11.0",
     "stacktrace-parser": "0.1.11",
-    "undici": "8.0.2",
+    "undici": "catalog:",
```
bnchrch, line 50: 💎

sdk/http/src/index.ts
```diff
@@ -2,6 +2,7 @@ import ky from 'ky';
 import type { Options } from 'ky';
 import { assignRequestId, traceRequest, traceResponse, traceError } from './hooks/index.js';
 import { applyFetchErrorTracing } from '#hooks/trace_error.js';
+export * from './fetch/index.js';
```
bnchrch: ❓ Should we take the time to replace the fetch in our http client now or later?

sdk/http/src/fetch/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+  /cookie/i
+];
+
+/**
+ * Redacts sensitive headers for safe logging
+ *
+ * @param headers
+ * @returns Plain object with sensitive headers redacted
+ */
+export const redactHeaders = ( headers: Headers ) : Record<string, unknown> => {
```
bnchrch, line 67: ❗ This function already exists at `sdk/http/src/utils/redact_headers.ts`

sdk/http/src/fetch/index.ts
```diff
@@ -0,0 +1,76 @@
...
+ * - Adds `x-request-id` header to requests with random value (over-writable)
+ * - Sends the request, response, error and/or failure to trace (from @outputai/core).
+ *
+ * @see {@link https://fetch.spec.whatwg.org/}
+ * @param input - URL string, URL object or Request object
+ * @param init - Request options
+ * @returns The HTTP response
+ */
+export const fetch = async ( input: RequestInfo, init?: RequestInit ) : Promise<Response> => {
+  const requestId = randomUUID();
```
bnchrch: ❗ We have logic for this in our http client already. See the fns `requestIdMiddleware`,  and `createTraceId`. We should reuse it so that we dont diverge tracing+request id behaviour

sdk/http/src/fetch/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Sends the trace start event for an http request
+ *
+ * @param options
+ * @param options.requestId - id of the request
+ * @param options.request - The HTTP Request object
+ * @returns
+ */
+export const logRequest = async ( { requestId, request } : { requestId: string, request: Request } ) : Promise<void> =>
```
bnchrch: ❗ We have logic for this in our http client already. See `traceRequest`, `parseRequestBody`

bnchrch (commented):
Nothing structural but some major (❗ ) concerns around code duplication.

Should be good to go on the next round!

-----
**Legend**
- ❗Change requested
- ❓Question asked
- 📚 Documentation requested 
- 🧪 Test requested

sdk/http/src/index.ts
szanata: Later

sdk/http/src/fetch/utils.ts
szanata: The signature is different, the other one accepts objects, this one accepts on headers.

sdk/http/src/fetch/index.ts
szanata: I think it is better to have them separate, so it is easy to deprecated down the road, the duplication is minimal.

sdk/http/src/fetch/utils.ts
szanata: It is not exactly the same logic, albeit similar, so I think it is best to have it duplicate so it is easier to deprecate ky later.

szanata: It is not duplicated anymore.

sdk/http/src/fetch/index.ts
szanata: It is not duplicated anymore.

sdk/http/src/fetch/utils.ts
szanata: It is not duplicated anymore.

sdk/http/src/fetch/utils.ts
bnchrch, line 1: 💅 I would recommend we break this up into `utils/trace` `utils/response_parse` and `utils/redact` or something like that

sdk/http/src/fetch/index.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @returns The HTTP response
+ */
+export const fetch = async ( input: RequestInfo | Request, init?: RequestInit ) : Promise<Response> => {
+  // Creates a Request object with the many shapes RequestInfo can have
+  const base = new undici.Request( input as RequestInfo, init );
+  // Creates a headers object with the many shapes Request.Headers can have (object, array, Headers)
+  const headers = new undici.Headers( base.headers );
+
+  const requestId = randomUUID();
+  headers.set( 'x-request-trace-id', requestId );
```
bnchrch, line 42: ❓ do we want to give people the option of setting their own?

bnchrch (approved):
A few minor things.

nothing blocking

-----
**Legend**
- ❗Change requested
- ❓Question asked
- 📚 Documentation requested 
- 🧪 Test requested

sdk/http/src/fetch/index.ts
szanata: Since we moved to `x-request-trace-id` instead of `x-request-id` (which is more standard) I would say no. Not for now.

sdk/http/src/fetch/utils.ts
szanata: Ok, makes sense. The file is not that long, but I is a good practice.

sdk/http/src/fetch/index.ts
```diff
@@ -0,0 +1,59 @@
...
+  const request = new undici.Request( base, { headers } );
+
+  await logRequest( { requestId, request } );
```
catsby, line 45: If `OUTPUT_TRACE_HTTP_VERBOSE` is `true`, then `logRequest` will call `parseBody()` and I think can crash if there is no actual JSON in the request, which I believe is normal-ish if you're sending a `DELETE` call? 

I'm pretty sure that's the execution chain and what can happen, and if so, we have a situation where logging interferes with what is a valid request, but only if we're doing verbose logging

sdk/http/src/fetch/index.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  try {
+    const response = await undici.fetch( request );
+    if ( response.status > 399 ) {
+      logError( { requestId, response } );
+      return response;
+    }
+    await logResponse( { requestId, response } );
```
catsby, line 53: I think there could be a similar problem here if we have `OUTPUT_TRACE_HTTP_VERBOSE` set to `true`, in this case if the server returns malformed json or even something like `204 no content` with an `application/json` return header, then we'll fail to parse it even if we got a successful response.

sdk/http/src/fetch/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+  name: error.constructor.name,
+  message: error.message,
+  stack: error.stack,
+  code: ( error as { code?: string } ).code ?? undefined,
+  cause: ( () : object | string | undefined => {
+    if ( depth > 5 ) {
+      return '<Max recursion depth reached>';
+    }
+    if ( error.cause instanceof Error ) {
+      return serializeError( error.cause );
```
catsby: isn't this missing a `depth`?

pnpm-lock.yaml
```diff
@@ -15902,6 +15912,8 @@ snapshots:
 
   undici-types@7.18.2: {}
 
+  undici@8.0.2: {}
+
   undici@8.1.0: {}
```
catsby: We have 2 different `undici` versions in our lockfile

catsby (changes requested):
This is what I have so far

sdk/http/src/fetch/index.ts
```diff
@@ -0,0 +1,59 @@
...
+import type { RequestInfo, RequestInit } from 'undici';
+
+/*
+ * Unifies undici and nodes realms
+ * https://github.com/nodejs/undici#keep-fetch-and-formdata-together
+ */
+undici.install();
+
+/** Re-export undici library for convenience. */
+export * as undici from 'undici';
```
catsby, line 13: Do we really want to export everything? There's like 60 some things I think

sdk/http/src/fetch/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+ * Sensitive header patterns for redaction (case-insensitive)
+ */
+const SENSITIVE_HEADER_PATTERNS : RegExp[] = [
+  /authorization/i,
+  /token/i,
+  /api-?key/i,
+  /secret/i,
+  /password/i,
+  /pwd/i,
+  /key/i,
```
catsby: Does `/key/i,` match anything `/api-?key/i,` would match? It also looks like it would catch `Cache-Key`, is that correct?

sdk/http/src/fetch/index.ts
szanata: We are exporting under `undici.*` so there wiull be no name crashes. Yes we want to export, since people could use undici without needing to install a version themselves, avoiding the having to install two versions of the library.

sdk/http/src/fetch/utils.ts
```diff
@@ -0,0 +1,72 @@
+import type { Request, Response, Headers } from 'undici';
+
+/**
+ * Sensitive header patterns for redaction (case-insensitive)
+ */
+const SENSITIVE_HEADER_PATTERNS : RegExp[] = [
+  /authorization/i,
+  /token/i,
```
catsby: will this filter out things like `x-tokens-remaining`? I'm not sure if that's a valid header but OpenAI has a few [defined here](https://developers.openai.com/api/docs/guides/rate-limits#rate-limits-in-headers) that look like they might get redacted, e.g. `x-ratelimit-limit-tokens`

pnpm-lock.yaml
szanata: That is how it resolved it. But I can upgrade to use 8.1.0

sdk/http/src/fetch/logger.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+  return import( './logger.js' );
+}
+
+beforeEach( () => {
+  tracing.addEventStart.mockClear();
+  tracing.addEventEnd.mockClear();
+  tracing.addEventError.mockClear();
+} );
+
+describe( 'fetch/utils', () => {
```
catsby: should this be `fetch/logger`? `utils.spec.ts` has a `describe( 'fetch/utils' )` as well

sdk/http/src/fetch/logger.ts
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Sends the trace error event for an http response with error status
+ *
+ * @param options
+ * @param options.requestId - id of the request
+ * @param options.response - The HTTP Response object
+ * @returns
+ */
+export const logError = ( { requestId: id, response: { status, statusText, headers } } : { requestId: string, response: Response } ) : void =>
+  Tracing.addEventError( { id, details: { status, statusText, headers: redactHeaders( headers ) } } );
```
catsby: If I read this correctly, we only log the status and status text with errors, but both `logRequest()` and `logResponse()` log the body if verbose mode is on. Should we log the actual error message here, even if just in verbose mode?

sdk/http/src/fetch/logger.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+  } );
+
+/**
+ * Sends the trace error event for an http response with error status
+ *
+ * @param options
+ * @param options.requestId - id of the request
+ * @param options.response - The HTTP Response object
+ * @returns
```
catsby: `logRequest`, `logError`, and `logResponse` all have `@returns ` in the code docs, but they all return some kind of `void`, is that typical?

sdk/http/src/fetch/utils.ts
szanata: It was missing. I added it

sdk/http/src/fetch/index.ts
szanata: You are right. I refactored it

sdk/http/src/fetch/utils.ts
szanata: Yes. But that was copied from the previous code. It will match `Cache-Key`. I can add exceptions, but this will most likely never be "perfect"

Also, `Cache-Key` is not a standard header.

sdk/http/src/fetch/logger.spec.ts
szanata: Yes, it is a typo

sdk/http/src/fetch/logger.ts
szanata: It is reasonable

sdk/http/src/fetch/logger.ts
szanata: Fixed. It was an issue with the way I was formatting TSDocs prior. Thanks

catsby (approved):

PR #120, opened 2026-04-08
chore: feedback/suggestions to improve our coding agents in how they generate workflows

catsby:
Updates coding assistant plugin files to improve AI-generated workflow code quality and command routing. These changes are intended to improve the code generated when users ask Claude to create or iterate on workflows. These changes add things like structured output guidance, schema placement rules, a new code style skill, and try to add consistent formatting across all plugin files. Also adds an entirely new skill around code style called `output-dev-code-style`.

- Add `output-dev-code-style` skill covering project ESLint rules (trailing commas, const-only, arrow parens, operator linebreak, spacing)
- Add guidance against duplicating `Output.object()` schemas in prompts (grounded in Anthropic and Google Vertex AI best practices)
- Add rule requiring LLM schemas to be defined in `types.ts`, not inline in steps
- Update prompt writer agent with correct structured output patterns and pitfall documentation
- Add schema placement and code style checks to the post-flight checklist
- Apply consistent code style (space-in-parens, no trailing commas, `fn: async input =>`) across all plugin files
- Improve `plan_workflow` command description with trigger phrases for better semantic matching
- Add workflow creation routing directive to `SESSION_START_CONTEXT.md` so Claude always uses `/outputai:plan_workflow` instead of manual planning
- Apply style sweep to 7 remaining plugin files not covered in initial pass (error skills, workflow-result, skill-file, project-context, build command)
- Document `generateText` variable type constraint (`string | number | boolean` only) -- arrays/objects must be pre-formatted in steps
- Remove Liquid loop/nested object patterns from prompt writer agent that conflict with the variable type constraint
- Add provider preference question to `plan_workflow` planning flow
- Add LLM provider consistency rule -- all prompts in a workflow should use the same provider, default to Anthropic, unless specified in the `plan_workflow` step
- Fixes templated `.claude/settings.json` for running Output skills (was `Skills()` but it should be `Skill()`)

### DIFF
--- a/coding_assistants/claude/plugins/outputai/skills/output-dev-code-style/SKILL.md
+++ b/coding_assistants/claude/plugins/outputai/skills/output-dev-code-style/SKILL.md
@@ -27,16 +27,16 @@ Never use trailing commas in objects, arrays, function parameters, or type defin
 const config = {
   name: 'workflow',
   timeout: 30000
-}
+};
 
-const items = [ 'a', 'b', 'c' ]
+const items = [ 'a', 'b', 'c' ];
 
 export const myStep = step( {
   name: 'myStep',
   inputSchema: MyInputSchema,
   outputSchema: MyOutputSchema,
   fn: async input => {
-    return { result: input.value }
+    return { result: input.value };
   }
 } );
 ```
@@ -107,7 +107,7 @@ fn: async input => { ... }
 
 // Parens required for these cases:
 items.reduce( ( acc, item ) => acc + item, 0 )
-const run = ( { name, id } ) => `${ name }-${ id }`;
+const run = ( { name, id } ) => `${name}-${id}`;
 const noop = () => {};
 ```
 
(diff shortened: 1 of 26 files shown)


coding_assistants/claude/plugins/outputai/skills/output-dev-code-style/SKILL.md
```diff
@@ -0,0 +1,176 @@
+---
+name: output-dev-code-style
+description: Code style conventions for Output SDK projects. Use when writing or reviewing any TypeScript/JavaScript code to ensure it follows project formatting rules.
+allowed-tools: [Read]
+---
+
+# Code Style Conventions
```
bnchrch, line 7: ❗ we don't bootstrap a linter. And likely consumer projects will be using a code style that differs from ours (We use a custom rule set when many would use eslint-config-airbnb, eslint-config-standard, prettier etcc)

So in this file we may be best off having the instructions to be to check the lint rules, run the lint command and if it doesn't exist fallback to the code conventions of the repo

bnchrch (changes requested):

coding_assistants/claude/plugins/outputai/skills/output-dev-code-style/SKILL.md
catsby: Makes sense, I updated it to refer or discover the projects linting rules first, then try to follow conventions in 496896c

bnchrch (approved):

PR #121, opened 2026-04-09
chore(cli): update post generate instructions to use real scenario

bnchrch:
## Problem
- `workflow generate` success message used the raw CLI name instead of the camelCased workflow ID registered in the generated code, causing the example `workflow run` command to fail
- The scenario name was hardcoded as `test_input` rather than derived from actually generated files
- The LLM-based generation path (`--plan-file`) could produce entirely different workflow names and scenarios, but the success message always reflected the skeleton defaults

## Solution
- Added `parseWorkflowDir()` utility that scans the generated directory after all generation completes (both skeleton and AI-based) to extract the real workflow ID from `workflow.ts` and list scenario files
- Updated the success message to use the parsed workflow ID and first scenario name in example commands
- Works for both `--skeleton` and `--plan-file` generation paths since it reads the final state of the directory

### DIFF
--- a/sdk/cli/src/utils/workflow_dir_parser.spec.ts
+++ b/sdk/cli/src/utils/workflow_dir_parser.spec.ts
@@ -55,6 +55,16 @@ describe( 'parseWorkflowDir', () => {
     expect( result.workflowId ).toBeUndefined();
   } );
 
+  it( 'should return undefined workflowId when workflow file has no name property', () => {
+    writeFileSync(
+      join( tempDir, 'workflow.ts' ),
+      'export default workflow( { description: \'no name here\' } );'
+    );
+
+    const result = parseWorkflowDir( tempDir );
+    expect( result.workflowId ).toBeUndefined();
+  } );
+
   it( 'should list scenario names from scenarios directory', () => {
     const scenariosDir = join( tempDir, 'scenarios' );
     mkdirSync( scenariosDir );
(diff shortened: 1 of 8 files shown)


sdk/cli/src/utils/workflow_dir_parser.ts
```diff
@@ -0,0 +1,43 @@
...
+const SCENARIOS_DIR = 'scenarios';
+const WORKFLOW_NAME_PATTERN = /name:\s*['"]([^'"]+)['"]/;
+
+export interface WorkflowDirInfo {
+  workflowId: string | undefined;
+  scenarioNames: string[];
+}
+
+function readWorkflowName( filePath: string ): string | undefined {
+  const content = readFileSync( filePath, 'utf-8' );
```
catsby: I believe `readFileSync` can throw, and there doesn't seem to be any `try/catch` here or up the chain, so if I understand correctly if something goes sideways here it could cause the CLI to crash?

sdk/cli/src/utils/workflow_dir_parser.ts
```diff
@@ -0,0 +1,43 @@
...
+
+function parseWorkflowId( targetDir: string ): string | undefined {
+  const workflowFile = WORKFLOW_FILE_NAMES
+    .map( name => join( targetDir, name ) )
+    .find( existsSync );
+
+  return workflowFile ? readWorkflowName( workflowFile ) : undefined;
+}
+
+function listScenarioNames( targetDir: string ): string[] {
```
catsby, line 38: same `try/catch` concern here

sdk/cli/src/utils/workflow_dir_parser.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+  fn: async ( input ) => {
+    return input;
+  }
+} );`
+    );
+
+    const result = parseWorkflowDir( tempDir );
+    expect( result.workflowId ).toBe( 'complexWorkflow' );
+  } );
+} );
```
catsby, line 113: can we add a test to verify behavior when parseWorkflowDir returns `undefined`? From `return match?.[1];`

catsby (commented):
defensive suggestions

catsby (approved):

PR #124, opened 2026-04-10
chore(cli): include workflow alias's

bnchrch:
## Overview

- Add `aliases` field to the Swagger `Workflow` schema so the OpenAPI spec and generated clients expose workflow aliases
- Regenerate `openapi.json` and CLI TypeScript types to include `aliases?: string[]` on the `Workflow` interface
- Update CLI `workflow list` to display aliases in table/JSON output and match them when filtering with `--filter`
- Add test coverage for alias display in `list.spec.ts`


sdk/cli/src/commands/workflow/list.ts
```diff
@@ -94,6 +100,7 @@ function formatWorkflowsAsJson( workflows: Workflow[] ): string {
       return {
         name: display.name,
         description: display.description,
+        aliases: display.aliases === 'none' ? [] : display.aliases.split( ', ' ),
```
catsby:
```suggestion
        aliases: w.aliases ?? [],
```

nit: `parseWorkflowForDisplay()` above checks aliases and converts it to a string from the `w` object, and here we convert it back to an array, so just use the `w` thing directly

sdk/cli/src/commands/workflow/list.ts
```diff
@@ -68,12 +73,13 @@ function createWorkflowTable( workflows: Workflow[], detailed: boolean ): string
     const display = parseWorkflowForDisplay( workflow );
 
     if ( detailed ) {
+      const aliases = display.aliases.split( ', ' ).join( '\n' );
```
catsby: same run-around as mentioned (though you may see this one first), we take the `aliases` and join them into a string, just to break them back out to an array here, so just use the source
```suggestion
      const aliases = workflow.aliases?.length ? workflow.aliases.join( '\n' ) : 'none';
```

catsby (approved):
Minor nits, take them or leave them

PR #133, opened 2026-04-15
chore: Migrate some Output plugin Commands to Skills

catsby:
## Summary

Migrate `debug` and `plan` commands to be skills, which I think makes more sense than commands that are directly invoked. They can still be directly invoked, but skills are more likely to be loaded and used in the course of chat with Claude than if left as commands.

### DIFF
--- a/coding_assistants/claude/plugins/outputai/commands/build_workflow.md
+++ b/coding_assistants/claude/plugins/outputai/commands/build_workflow.md
@@ -1,6 +1,6 @@
 ---
 argument-hint: "[workflow-plan-file-path] [workflow-name] [workflow-directory]"
-description: Workflow Implementation Command for Output SDK
+description: Implement an Output SDK workflow from a plan document. Use after plan_workflow has produced a plan and the user is ready to build.
 version: 0.1.3
 model: opus
 ---
@@ -324,10 +324,6 @@ Verify the implementation is ready for use.
 </step>
 
 </process_flow>
-
-<post_flight_check>
-  EXECUTE: Claude Skill: `output-meta-post-flight`
-</post_flight_check>
 
 ---- START ----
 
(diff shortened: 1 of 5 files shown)


coding_assistants/claude/plugins/outputai/commands/build_workflow.md
```diff
@@ -1,6 +1,6 @@
 ---
 argument-hint: "[workflow-plan-file-path] [workflow-name] [workflow-directory]"
-description: Workflow Implementation Command for Output SDK
+description: Implement an Output SDK workflow from a plan document. Use after plan_workflow has produced a plan and the user is ready to build.
```
bnchrch: 💅 This can go to a skill as well.

Since we did this, all skills are commands and commands are deprecated.

coding_assistants/claude/plugins/outputai/skills/plan-workflow/SKILL.md
```diff
@@ -1,6 +1,8 @@
 ---
+name: plan-workflow
 argument-hint: [workflow-description-and-additional-instructions]
 description: Use when the user asks to create, build, generate, scaffold, or plan a new workflow. Orchestrates the full planning process including architecture, steps, prompts, evaluators, and testing strategy using specialized subagents.
+when_to_use: When user says "create a workflow", "build a workflow", "make a workflow", "generate a workflow", "scaffold a workflow", "plan a workflow", or describes workflow functionality they want built for the Output SDK
```
bnchrch: ❗ `when_to_use` is not part of the skill spec. You need to use `description`
https://agentskills.io/specification

bnchrch (changes requested):

coding_assistants/claude/plugins/outputai/skills/plan-workflow/SKILL.md
catsby: good catch, I removed it in 785235e

bnchrch (approved):

PR #134, opened 2026-04-16
feat: add workflow history endpoint

catsby:
### Update: 

We should merge PR #135 and rebase here before review+merge of this PR

## Summary

Adds a new endpoint to retrieve a workflows history. The results are paginated and more or less just pass on what we get from Temporal, even using Temporal's next page token value to use for finding the next set of events.

Example output:

<details>

```
❯ curl -s http://localhost:3001/workflow/Y4anphMwOjwu5m27nqLSq/history | jq
{
  "workflow": {
    "workflowId": "Y4anphMwOjwu5m27nqLSq",
    "runId": "554c2511-1201-4257-9cc9-2dfbf55523d7",
    "status": "completed",
    "startTime": "2026-04-16T17:28:49.071Z",
    "closeTime": "2026-04-16T17:28:50.171Z",
    "historyLength": 17,
    "taskQueue": "main"
  },
  "events": [
    {
      "eventId": "1",
      "eventType": 1,
      "eventTypeName": "WORKFLOW_EXECUTION_STARTED",
      "eventTime": "2026-04-16T17:28:49.071Z",
      "workflowExecutionStartedEventAttributes": {
        "completionCallbacks": [],
        "workflowType": {
          "name": "continue_as_new"
        },
        "taskQueue": {
          "name": "main",
          "kind": "TASK_QUEUE_KIND_NORMAL"
        },
        "workflowExecutionTimeout": {
          "seconds": "86400"
        },
        "workflowRunTimeout": {
          "seconds": "86400"
        },
        "workflowTaskTimeout": {
          "seconds": "10"
        },
        "continuedExecutionRunId": "4ae28eab-2f43-4159-8833-3ca640206fdc",
        "initiator": 1,
        "originalExecutionRunId": "554c2511-1201-4257-9cc9-2dfbf55523d7",
        "firstExecutionRunId": "019d9756-97f1-7938-9a5f-cdc30c7a1b1c",
        "attempt": 1,
        "workflowExecutionExpirationTime": {
          "seconds": "1776446927",
          "nanos": 857000000
        },
        "firstWorkflowTaskBackoff": {
          "nanos": 785768697
        },
  [... more events ...]
    {
      "eventId": "17",
      "eventType": 2,
      "eventTypeName": "WORKFLOW_EXECUTION_COMPLETED",
      "eventTime": "2026-04-16T17:28:50.171Z",
      "workflowExecutionCompletedEventAttributes": {
        "workflowTaskCompletedEventId": {
          "low": 16,
          "high": 0,
          "unsigned": false
        }
      }
    }
  ],
  "nextPageToken": null
}

```

</details> 

*By default*, this does not include the event payloads, as they could be quite large. Use the `includePayloads` URL parameter to get the full payloads (base64 encoded, direct from Temporal). 

At this time I'm opting for `includePayloads` instead of something like `include=payloads` for now. At this time I'm not sure what other fields we would omit by default. If we decide to later have more fields omitted or even support `include` and `exclude`, that's probably a small lift

## Claude Generated Summary:

Below is a more detailed and comprehensive explanation of the pull request, but I'm hiding it because it's very wordy and very generated.

<details> 

## How it works

The endpoint returns paginated Temporal history events with ergonomic serialization:
- Timestamps as ISO 8601 strings
- Event type names as human-readable strings (e.g. `"ACTIVITY_TASK_SCHEDULED"`)
- Step names extracted from activity type names (e.g. `"extractPassages"`)
- Payloads (step inputs/outputs) excluded by default to keep responses lightweight

### Basic request

```
GET /workflow/my-workflow-abc123/history
```

### Response (first page)

```json
{
  "workflow": {
    "workflowId": "my-workflow-abc123",
    "runId": "run-uuid",
    "status": "running",
    "startTime": "2024-04-15T12:00:00.000Z",
    "closeTime": null,
    "historyLength": 42,
    "taskQueue": "default"
  },
  "events": [
    {
      "eventId": "1",
      "eventType": 1,
      "eventTypeName": "WORKFLOW_EXECUTION_STARTED",
      "eventTime": "2024-04-15T12:00:00.000Z",
      "workflowExecutionStartedEventAttributes": {
        "workflowType": { "name": "factChecker" }
      }
    },
    {
      "eventId": "5",
      "eventType": 10,
      "eventTypeName": "ACTIVITY_TASK_SCHEDULED",
      "eventTime": "2024-04-15T12:00:01.000Z",
      "activityTaskScheduledEventAttributes": {
        "activityId": "1",
        "activityType": { "name": "fact-checker#extractPassages" },
        "stepName": "extractPassages"
      }
    },
    {
      "eventId": "6",
      "eventType": 11,
      "eventTypeName": "ACTIVITY_TASK_STARTED",
      "eventTime": "2024-04-15T12:00:01.500Z",
      "activityTaskStartedEventAttributes": {
        "scheduledEventId": "5"
      }
    },
    {
      "eventId": "7",
      "eventType": 12,
      "eventTypeName": "ACTIVITY_TASK_COMPLETED",
      "eventTime": "2024-04-15T12:00:10.000Z",
      "activityTaskCompletedEventAttributes": {
        "scheduledEventId": "5"
      }
    }
  ],
  "runId": "run-uuid",
  "nextPageToken": "abc123..."
}
```

### A note on `:id` vs `runId`

The `:id` in `/workflow/:id/history` is the **workflowId** -- the user-visible execution ID returned by `POST /workflow` (e.g. `my-workflow-abc123`). This is consistent with every other endpoint in the API (`/status`, `/result`, `/trace-log`, `/reset`, etc.).

The `runId` is a separate Temporal concept that identifies a specific run within a workflow. A single workflowId can have multiple runs (after resets or continue-as-new). When not specified, the API resolves to the latest run. When you need a specific run, pass `runId` as a query parameter.

### Pagination

Histories can be large, so results are paginated. Use nextPageToken from each response to fetch the next page. When paginating, you must pass the runId from the first page to ensure all pages come from the same workflow run. The top-level runId is provided on every response for this purpose — you can echo it directly:

    GET /workflow/my-workflow-abc123/history?pageToken=abc123...&runId=run-uuid

Subsequent pages return "workflow": null — the workflow metadata doesn't change, so consumers should cache it from the first page. The top-level runId is preserved on every page.

### Including step inputs and outputs

By default, payloads (step inputs, outputs, and failure details) are excluded to keep responses small. To include them:

```
GET /workflow/my-workflow-abc123/history?includePayloads=true
```

This is useful when inspecting a specific step's data. For live polling, the default (no payloads) is recommended.

### Targeting a specific run

By default, the endpoint returns the latest run. To target a specific run (e.g. one from before a reset):

```
GET /workflow/my-workflow-abc123/history?runId=specific-run-uuid
```

### Query parameters

| Parameter | Default | Description |
|-----------|---------|-------------|
| `runId` | latest run | Target a specific workflow run |
| `pageSize` | 20 (max 50) | Number of events per page |
| `pageToken` | (none) | Pagination cursor from previous response |
| `includePayloads` | false | Include step inputs, outputs, and failure details |

## Building a visualization from history events

The history returns raw events. To build a step-by-step visualization, consumers correlate events into per-activity records:

1. **ACTIVITY_TASK_SCHEDULED** -- a step was queued. Carries `stepName` and `activityId`.
2. **ACTIVITY_TASK_STARTED** -- a step began executing. Links to its SCHEDULED event via `scheduledEventId`.
3. **ACTIVITY_TASK_COMPLETED** -- a step finished. Links via `scheduledEventId`. Includes output when `includePayloads=true`.
4. **ACTIVITY_TASK_FAILED** -- a step failed. Links via `scheduledEventId`. Includes failure details when `includePayloads=true`.

By grouping these events via `scheduledEventId`, a consumer can determine each step's status, duration, and data.

To show human-readable step names, descriptions, and steps that haven't executed yet, combine history data with the workflow catalog (`GET /workflow/catalog/:id`). Match `stepName` from history events to `functionName` in the catalog's activity list.

## Live polling

To show real-time progress, poll the endpoint on an interval. On each poll:

1. Fetch from page 1 (re-fetch the full history to get the latest state)
2. Paginate through all pages to build the complete picture
3. Check `workflow.status` -- stop polling when it reaches a terminal state (`completed`, `failed`, `canceled`, `terminated`, `timed_out`)

</details>


api/src/clients/temporal_client.js
```diff
@@ -67,14 +68,69 @@ const TemporalStatus = {
...
+  WORKFLOW_EXECUTION_TERMINATED: 27,
+  WORKFLOW_EXECUTION_CONTINUED_AS_NEW: 28,
+  START_CHILD_WORKFLOW_EXECUTION_INITIATED: 29,
+  CHILD_WORKFLOW_EXECUTION_STARTED: 31,
+  CHILD_WORKFLOW_EXECUTION_COMPLETED: 32,
+  CHILD_WORKFLOW_EXECUTION_FAILED: 33,
+  CHILD_WORKFLOW_EXECUTION_CANCELED: 34,
+  CHILD_WORKFLOW_EXECUTION_TIMED_OUT: 35,
+  CHILD_WORKFLOW_EXECUTION_TERMINATED: 36,
+  MARKER_RECORDED: 25
```
catsby: note: I'm not 100% on including these here. They're unlikely to change, but it feels odd. I'm still thinking about it

api/src/index.js
```diff
@@ -979,6 +982,74 @@ app.get( '/workflow/:id/runs/:rid/result', resultHandler );
 app.get( '/workflow/:id/trace-log', traceLogHandler );
 app.get( '/workflow/:id/runs/:rid/trace-log', traceLogHandler );
 
+/**
+ * @swagger
+ * /workflow/{id}/history:
```
bnchrch, line 978: ❓ We should also add `workflow/{wid}/run/{rid}/history` right?

bnchrch (commented):
One comment.

Additionally (in this PR or another) we should 

- [ ] update the cli to use this endpoint
- [ ] update claude skills to include knowledge of this

api/src/clients/event_types.js
```diff
@@ -0,0 +1,66 @@
+// Values correspond to temporal.api.enums.v1.EventType protobuf enum.
+export const EventType = {
```
catsby, line 2: This event list is needed so we can pass things on to consumers. It's taken from Temporal, and we have a test to verify we stay in sync with it. The alternative would be to import the Temporal dependency directly and strip the `EVENT_TYPE_` prefix from them. The `@temporalio/proto` package was already a transitive dependency, and this way we can keep it just as a `devDependency` used by the spec that tests conformity

api/src/clients/temporal_client.js
```diff
@@ -143,6 +163,116 @@ export const extractWorkflowInput = history => {
...
+    delete attrs.failure;
+    delete attrs.details;
+    delete attrs.lastCompletionResult;
+    delete attrs.lastFailure;
+  }
+
+  serialized[attrKey] = attrs;
+  return serialized;
+};
+
```
szanata: This belongs on a separate file, with a fixture from Temporal and proper tests.

api/src/clients/temporal_client.js
```diff
@@ -527,6 +657,72 @@ export default {
         return { workflowId, runId: response.runId };
       },
 
+      async getWorkflowHistory( workflowId, { runId, pageSize = 20, pageToken, includePayloads = false } = {} ) {
+        const firstPage = !pageToken;
```
szanata: `isFirstPage` would a clearername for this var.

api/src/clients/temporal_client.js
```diff
@@ -67,14 +72,29 @@ const TemporalStatus = {
...
+const serializeEventTime = eventTime => {
+  if ( eventTime?.seconds === null || eventTime?.seconds === undefined ) {
+    return null;
+  }
+  const seconds = toNumberSafe( eventTime.seconds );
+  const nanos = toNumberSafe( eventTime.nanos );
+  if ( !Number.isFinite( seconds ) || !Number.isFinite( nanos ) ) {
+    return null;
+  }
+  return new Date( ( seconds * 1000 ) + Math.floor( nanos / 1e6 ) ).toISOString();
```
szanata: This code is too conservative and verbose.

According to temporal, and event time is:

```js
{
  "eventTime": {
    "seconds": "1713696930",
    "nanos": 123000000
  }
}
```

So I dont think it is necessary to have so many type casts and tests, just go for the basic:

```js
const serializeEventTime = eventTime => {
  if ( !eventTime?.seconds || !eventTime?.seconds ) {
    return null;
  }
  return new Date( ( +eventTime.seconds * 1000 ) + Math.floor( eventTime.nanos / 1e6 ) ).toISOString();
```

api/src/clients/temporal_client.js
```diff
@@ -67,14 +72,29 @@ const TemporalStatus = {
...
- */
-const EventType = {
-  WORKFLOW_TASK_COMPLETED: 7,
-  ACTIVITY_TASK_SCHEDULED: 10,
-  ACTIVITY_TASK_COMPLETED: 12
+const warnedUnknownEventTypes = new Set();
+
+// Subset of gRPC status codes from @grpc/grpc-js (transitive through @temporalio/client).
+// See https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
+const GRPC_STATUS = { INVALID_ARGUMENT: 3, NOT_FOUND: 5 };
```
szanata: I think this should be PascalCase:
```js
const GrpcStatus = { INVALID_ARGUMENT: 3, NOT_FOUND: 5 };

// so to use as
GrpcStats.INVALID_ARGUMENT;
```

In JS is incommon to have `VAR.PROP` both in all caps.

api/src/handlers/workflow_history.js
```diff
@@ -0,0 +1,38 @@
+import { z } from 'zod';
+import { readPinnedRunId } from './utils.js';
+
+export function createWorkflowHistoryHandler( client ) {
+  const querySchema = z.object( {
+    runId: z.string().optional(),
+    pageSize: z.coerce.number().int().min( 1 ).max( 50 ).default( 20 ),
+    pageToken: z.preprocess(
+      v => v === '' ? undefined : v,
+      z.string().regex( /^[A-Za-z0-9+/]+={0,2}$/, 'Invalid pageToken format' ).optional()
```
szanata: This could be a separate function, and the Regexp could be isolated in a proper named variable.

api/src/handlers/workflow_history.js
```diff
@@ -0,0 +1,38 @@
...
+    ),
+    includePayloads: z.union( [ z.literal( 'true' ), z.literal( 'false' ) ] ).default( 'false' ).transform( v => v === 'true' )
+  } ).refine(
+    data => !data.pageToken || data.runId,
+    { message: 'runId is required when using pageToken', path: [ 'runId' ] }
+  );
+
+  return async ( req, res, next ) => {
+    try {
```
szanata: This try catch is unnecessary. All API events already have a default error handler.

api/src/middleware/error_handler.js
```diff
@@ -22,6 +22,7 @@ const GRPC_STATUS_HTTP = {
 
 const NAMED_ERROR_STATUSES = {
   [ZodError.name]: 400,
+  [InvalidPageTokenError.name]: 400,
```
szanata, line 25: 👍

szanata (commented):
I left some comments of stuff I think should be refined.

Plus, I think this could benefit from having actual fixtures from Temporal SDK responses, since it is a pretty complex endpoint.

api/src/index.js
catsby: I believe I did this in 782bb65

api/src/clients/temporal_client.js
catsby: refactored into it's own file in 1b44039

api/src/clients/temporal_client.js
catsby: updated in 48d99ed4836bb249b225b49356d14a341a340c85

api/src/clients/temporal_client.js
catsby: I refactored this into it's own file and tried to simplify things as suggested, please take a look

catsby: I lied, I forgot to simplify things, but I'm doing that now

catsby: This was refactored in 53ecc41fb33726ffd99d420c19b5f23c11669677

api/src/clients/temporal_client.js
catsby: Changed to `GrpcStatus` in 53ecc41fb33726ffd99d420c19b5f23c11669677

api/src/handlers/workflow_history.js
catsby: Agreed, thanks

api/src/handlers/workflow_history.js
catsby: We could do that yes, but I'm not sure I see the value if it's not going to be shared.

api/src/clients/event_serialization.js
```diff
@@ -0,0 +1,122 @@
...
+            Buffer.from( p.metadata.encoding ).toString() :
+            'unknown';
+          logger.warn( 'Failed to decode event payload', {
+            eventId: event.eventId?.toString(),
+            encoding,
+            error: error.message
+          } );
+          return { _raw: true, encoding };
+        }
+      } );
```
bnchrch: 💅 Could be a named function to improve readability

api/src/clients/event_serialization.js
```diff
@@ -0,0 +1,122 @@
...
+import { logger } from '#logger';
+import { EventTypeName } from './event_types.js';
+
+const warnedUnknownEventTypes = new Set();
+
+export const serializeEventTime = eventTime => {
+  if ( !eventTime?.seconds ) {
+    return null;
+  }
+  return new Date( ( +eventTime.seconds * 1000 ) + Math.floor( eventTime.nanos / 1e6 ) ).toISOString();
```
bnchrch, line 11: 💅 feels like there has to be a better way to parse this?

api/src/clients/temporal_client.js
```diff
@@ -39,15 +42,20 @@ const getCatalog = async ( { client, taskQueue } ) => {
...
+    return 'unspecified';
+  }
+  return ( {
+    RUNNING: 'running',
+    COMPLETED: 'completed',
+    FAILED: 'failed',
+    CANCELED: 'canceled',
+    TERMINATED: 'terminated',
+    TIMED_OUT: 'timed_out',
+    CONTINUED_AS_NEW: 'continued'
```
bnchrch: 💅 Should be its own const outside of this function

api/src/clients/temporal_client.js
```diff
@@ -524,6 +526,78 @@ export default {
         return { workflowId, runId: response.runId };
       },
 
+      async getWorkflowHistory( workflowId, { runId, pageSize = 20, pageToken, includePayloads = false } = {} ) {
+        if ( pageToken && !runId ) {
+          throw new InvalidPageTokenError();
+        }
+        const isFirstPage = !pageToken;
+        const metadata = isFirstPage ? await ( async () => {
```
bnchrch, line 536: 💅 Can this async fn be its own named function? it being inline is making it harder to reason about

api/src/clients/temporal_client.js
```diff
@@ -524,6 +526,78 @@ export default {
...
+              startTime: description.startTime?.toISOString() ?? null,
+              closeTime: description.closeTime?.toISOString() ?? null,
+              historyLength: description.historyLength,
+              taskQueue: description.taskQueue
+            },
+            resolvedRunId: description.runId
+          };
+        } )() : { workflow: null, resolvedRunId: runId };
+
+        const response = await connection.workflowService.getWorkflowExecutionHistory( {
```
bnchrch, line 561: 💅 would move this to its own fn

api/src/handlers/workflow_history.js
```diff
@@ -0,0 +1,34 @@
+import { z } from 'zod';
+import { readPinnedRunId } from './utils.js';
+
+export function createWorkflowHistoryHandler( client ) {
+  const querySchema = z.object( {
+    runId: z.string().optional(),
+    pageSize: z.coerce.number().int().min( 1 ).max( 50 ).default( 20 ),
+    pageToken: z.preprocess(
+      v => v === '' ? undefined : v,
+      z.string().max( 4096 ).regex( /^[A-Za-z0-9+/]+={0,2}$/, 'Invalid pageToken format' ).optional()
```
bnchrch, line 11: 💅 Could use a comment as to what were regexing for?

bnchrch (approved):
Couple nits I think we should address.

Also 

❗ we need a changeset before we merge.

Happy to pre-approve

-----
**Legend**
- ❗Change requested
- ❓Question asked
- 📚 Documentation requested 

api/package.json
```diff
@@ -18,6 +18,7 @@
     "zod": "catalog:"
   },
   "devDependencies": {
+    "@temporalio/proto": "catalog:",
```
szanata, line 21: I think we should not use this.

Image

api/src/clients/temporal_client.js
```diff
@@ -39,15 +42,22 @@ const getCatalog = async ( { client, taskQueue } ) => {
  * @param {string} status - The Temporal status name (e.g., 'RUNNING', 'COMPLETED')
  * @returns {string} User-friendly status string
  */
-const mapWorkflowStatus = status => ( {
+const WORKFLOW_STATUS_MAP = {
```
szanata: This will be a uppercase variable with upper case properties. This is not common in JS.

Either make the var OR the properties uppercase.

The most common flavor in Js is to make only the properties UPPERCASE.

api/src/clients/temporal_client.js
```diff
@@ -39,15 +42,22 @@ const getCatalog = async ( { client, taskQueue } ) => {
...
   CONTINUED_AS_NEW: 'continued'
-}[status] || status.toLowerCase() );
+};
+
+const mapWorkflowStatus = status => {
+  if ( !status ) {
+    return 'unspecified';
+  }
+  return WORKFLOW_STATUS_MAP[status] || status.toLowerCase();
+};
```
szanata, line 60: Why this change? will this field be ever empty?

szanata (changes requested):
Given the complexity of the data serialization, I must insist that we add some fixtures to the project, besides the tests (could be used in the tests as well).

This, plus my other comments.

api/package.json
catsby: `@temporalio/proto` is used exclusively as a dev dependency in a test, to make sure our event types stay in-sync with Temporal's. From my brief search, none of the packages listed there export the event types we need to use. We could remove it and the test, but then we risk drift between our types and Temporals that we would have to reconcile with once it's reported to us or something otherwise breaks. 

Considering it is exclusively a dev dependency used only in the way I describe, I feel it's safe to leave as-is, but I'll remain open to other suggestions anyone has on how to maintain parity with Temporal's event types without it.

szanata: Gotcha... we can keep it.

szanata (approved):
I am pre-approving here to unblock you.

api/src/clients/temporal_client.js
catsby: I changed it to `WorkflowStatusMap` in db46239

PR #136, opened 2026-04-17
feat(cli): add credentials set command

bnchrch:
## Summary

- Add `output credentials set KEY VALUE` command for programmatic credential updates
- Accepts dot-notation paths (e.g. `anthropic.api_key`) and creates nested keys as needed
- Decrypts, updates, and re-encrypts the credentials file
- Supports `--environment` and `--workflow` flags (same as other credentials commands)
- Useful in sandbox/CI environments where interactive `credentials edit` is not available

## Test plan

- [x] `npm run lint` passes
- [x] `set.spec.ts` tests pass (9 tests)
- [ ] Smoke test: `output credentials set anthropic.api_key sk-test && output credentials get anthropic.api_key`

### DIFF
--- a/sdk/cli/src/commands/credentials/set.ts
+++ b/sdk/cli/src/commands/credentials/set.ts
@@ -1,0 +1,79 @@
+import { Args, Command, Flags } from '@oclif/core';
+import { load as parseYaml, dump as stringifyYaml } from 'js-yaml';
+import {
+  decryptCredentials,
+  credentialsExist,
+  writeEncrypted,
+  resolveCredentialsPath
+} from '#services/credentials_service.js';
+
+type CredentialsObject = Record<string, unknown>;
+
+const setNestedValue = ( obj: CredentialsObject, dotPath: string, value: string ): void => {
+  const parts = dotPath.split( '.' );
+  const parent = parts.slice( 0, -1 ).reduce<CredentialsObject>( ( current, key ) => {
+    if ( !current[key] || typeof current[key] !== 'object' ) {
+      current[key] = {};
+    }
+    return current[key] as CredentialsObject;
+  }, obj );
+
+  parent[parts[parts.length - 1]] = value;
+};
+
+export default class CredentialsSet extends Command {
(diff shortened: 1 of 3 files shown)


sdk/cli/src/commands/credentials/set.ts
```diff
@@ -0,0 +1,79 @@
...
+      this.error( 'Cannot specify both --environment and --workflow.' );
+    }
+
+    if ( !credentialsExist( environment, workflow ) ) {
+      this.error(
+        `No credentials file found at ${resolveCredentialsPath( environment, workflow )}. Run "output credentials init" first.`
+      );
+    }
+
+    const plaintext = decryptCredentials( environment, workflow );
```
catsby: I believe `decryptCredentials` can throw, and if it does users will get a raw error and a CLI crash(?) from whatever internal thing threw instead of a nicely handled error message

sdk/cli/src/commands/credentials/set.ts
```diff
@@ -0,0 +1,79 @@
...
+    const data = ( parseYaml( plaintext ) || {} ) as CredentialsObject;
+
+    setNestedValue( data, args.path, args.value );
+
+    writeEncrypted( environment, stringifyYaml( data ), workflow );
```
catsby: same unhandled throw concern here

sdk/cli/src/commands/credentials/set.ts
bnchrch: Good catch — wrapped the decrypt+update+encrypt block in a try/catch that surfaces a friendly `this.error('Failed to update credentials: ...')` message, matching the pattern from `credentials edit`. Added tests for both decrypt and encrypt failure paths.

sdk/cli/src/commands/credentials/set.ts
bnchrch: Addressed in the same try/catch — both `decryptCredentials` and `writeEncrypted` errors are now caught and surfaced cleanly.

catsby (approved):

PR #137, opened 2026-04-17
feat(cli): add non-interactive mode for sandbox environments

bnchrch:
## Summary

- Add `--yes` / `--non-interactive` CLI flags and auto-detect `!process.stdin.isTTY`
- All `@inquirer/prompts` calls go through a new wrapper (`#utils/prompt.js`) that returns defaults when non-interactive
- Interactive refinement loops in `workflow plan` and `workflow generate` skip automatically
- Prevents CLI hangs in headless/sandbox environments

## Test plan

- [x] `npm run lint` passes
- [x] All affected tests pass
- [ ] Smoke test: `echo '' | npx output init --yes test-project` completes without hanging

### DIFF
--- a/sdk/cli/src/utils/interactive.ts
+++ b/sdk/cli/src/utils/interactive.ts
@@ -1,0 +1,7 @@
+const state = { nonInteractive: false };
+
+export const setNonInteractive = ( value: boolean ): void => {
+  state.nonInteractive = value;
+};
+
+export const isInteractive = (): boolean => !state.nonInteractive && !!process.stdin.isTTY;
--- a/sdk/cli/src/utils/prompt.ts
+++ b/sdk/cli/src/utils/prompt.ts
@@ -1,0 +1,31 @@
+import {
+  confirm as inquirerConfirm,
+  input as inquirerInput,
+  password as inquirerPassword
+} from '@inquirer/prompts';
+import { isInteractive } from './interactive.js';
+
+type ConfirmOptions = { message: string; default?: boolean };
+type InputOptions = { message: string; default?: string; validate?: ( value: string ) => boolean | string };
+type PasswordOptions = { message: string; mask?: boolean };
+
+export const confirm = async ( options: ConfirmOptions ): Promise<boolean> => {
+  if ( !isInteractive() ) {
+    return options.default ?? true;
+  }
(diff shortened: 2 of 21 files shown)


sdk/cli/src/utils/prompt.ts
```diff
@@ -0,0 +1,31 @@
...
+    return options.default ?? '';
+  }
+  return inquirerInput( options );
+};
+
+export const password = async ( options: PasswordOptions ): Promise<string> => {
+  if ( !isInteractive() ) {
+    return '';
+  }
+  return inquirerPassword( options );
```
szanata, line 30: In which scenario we would not want auto confirm in non-interactive mode?

bnchrch: the one case that motivated `default: false` was `coding_agents.ts` handling Claude plugin setup failure ("Would you like to proceed without the Claude plugin setup?").

In non-interactive mode the wrapper was returning `false` there, which silently cancelled headless installs with a misleading `UserCancelledError` when the real cause was the plugin command failing.

Just fixed this:
- `confirm()` now always returns `true` non-interactively (matches `--yes` semantics used by npm/apt)
- `handlePluginError` skips the recovery confirm and rethrows the real `ClaudePluginError` in non-interactive mode, so automation sees the actual cause
- Reverted the defensive `project_scaffold.ts` `default: false → true` flip since it's no longer needed

But `input` and `password` still honor their own defaults they're not boolean gates, and since `password()` has no default, secrets are written empty, which is the right behavior for "init credentials empty" in sandbox setups.

szanata (approved):

sdk/cli/src/utils/interactive.ts
```diff
@@ -0,0 +1,7 @@
+const state = { nonInteractive: false };
+
+export const setNonInteractive = ( value: boolean ): void => {
+  state.nonInteractive = value;
+};
+
+export const isInteractive = (): boolean => !state.nonInteractive && !!process.stdin.isTTY;
```
szanata, line 7: Why the check for the env var `process.stdin.isTTY` is here bvut also on `hooks/init.ts`?

bnchrch: good catch

PR #138, opened 2026-04-17
fix(worker): eliminate catalog startup race condition

bnchrch:
## Summary

- Replace `WorkflowIdConflictPolicy.FAIL` with `TERMINATE_EXISTING` for the catalog workflow startup
- Remove the `describe()` + `executeUpdate('complete')` dance — Temporal handles old workflow termination atomically
- Eliminates `WorkflowExecutionAlreadyStartedError` when restarting the worker before the old catalog fully closes

## Test plan

- [x] `npm run lint` passes
- [x] `start_catalog.spec.js` tests pass
- [ ] Smoke test: `npm run start:worker` starts cleanly on repeated restarts

### DIFF
--- a/.changeset/catalog-startup-fix.md
+++ b/.changeset/catalog-startup-fix.md
@@ -1,0 +1,5 @@
+---
+"@outputai/core": patch
+---
+
+Fix catalog workflow startup race condition by using TERMINATE_EXISTING conflict policy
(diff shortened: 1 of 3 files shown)


szanata (changes requested):
If there is an catalog running and another tries to run, it terminates the current one as failure, which is something we don't want in order to not pollute the logs.

The worker used to have this and this was deliberately changed.

PR #140, opened 2026-04-17
feat: add HTTP and gRPC proxy support for sandbox environments

bnchrch:
## Summary

- **Worker fetch proxy**: `bootstrapFetchProxy()` sets undici's `EnvHttpProxyAgent` as global dispatcher when `HTTPS_PROXY`/`HTTP_PROXY` is set — all activity fetch() calls route through it automatically
- **Worker gRPC proxy**: New `TEMPORAL_GRPC_PROXY` env var passed to `NativeConnection.connect()` as native HTTP CONNECT proxy option
- **CLI fetch proxy**: Same undici pattern in `bin/run.js` so CLI commands respect proxy env vars
- **API server**: `@grpc/grpc-js` already reads `grpc_proxy`/`https_proxy` automatically — no changes needed

## Test plan

- [x] `npm run lint` passes
- [x] All proxy tests pass
- [ ] Smoke test: Set `HTTPS_PROXY`, verify fetch calls route through proxy
- [ ] Smoke test: Set `TEMPORAL_GRPC_PROXY`, verify Temporal connection uses it


catsby (commented):
_posted in error_

sdk/cli/src/utils/proxy.ts
```diff
@@ -0,0 +1,12 @@
+import { EnvHttpProxyAgent, setGlobalDispatcher } from 'undici';
+
+export const bootstrapProxy = (): void => {
```
catsby, line 11: since this is exported should we document it?

sdk/cli/src/utils/proxy.ts
```diff
@@ -0,0 +1,12 @@
...
+  const proxyUrl = process.env.HTTPS_PROXY || process.env.https_proxy ||
+    process.env.HTTP_PROXY || process.env.http_proxy;
+
+  if ( !proxyUrl ) {
+    return;
+  }
```
catsby, line 17: Should we validate this is an actual url? How will the `EnvHttpProxyAgent()` thing respond if it's not? The CLI calls `bootstrapProxy()` with no try/catch so I think we'd get a crash or a raw error surfaced?

sdk/core/src/worker/proxy.js
catsby, line 1: This seems to be a copy of the cli/utils/proxy file, except this one has logging and is JS not TS. Is drift a concern between the two of them? If so, should there be a shared `sdk/utils`?

sdk/core/src/worker/configs.js
```diff
@@ -26,7 +26,9 @@ const envVarSchema = z.object( {
   // Whether to send activity heartbeats (enabled by default)
   OUTPUT_ACTIVITY_HEARTBEAT_ENABLED: z.transform( v => v === undefined ? true : isStringboolTrue( v ) ),
   // Time to allow for hooks to flush before shutdown
-  OUTPUT_PROCESS_FAILURE_SHUTDOWN_DELAY: z.preprocess( coalesceEmptyString, z.coerce.number().int().positive().default( 3000 ) )
+  OUTPUT_PROCESS_FAILURE_SHUTDOWN_DELAY: z.preprocess( coalesceEmptyString, z.coerce.number().int().positive().default( 3000 ) ),
+  // HTTP CONNECT proxy for Temporal gRPC connections (e.g. "proxy-host:8080")
+  TEMPORAL_GRPC_PROXY: z.string().optional()
```
catsby: should we validate this to be `host:port`? Given it's "HTTP CONNECT proxy" I could see people putting `http://host:port` or something

sdk/core/src/worker/index.js
```diff
@@ -52,8 +55,8 @@ const callerDir = process.argv[2];
   const catalog = createCatalog( { workflows, activities } );
 
   log.info( 'Connecting Temporal...' );
-  // Enable TLS when connecting to remote Temporal (API key present)
-  const connection = await NativeConnection.connect( { address, tls: Boolean( apiKey ), apiKey } );
+  const proxy = grpcProxy ? { type: 'http-connect', targetHost: grpcProxy } : undefined;
```
catsby, line 58: we should probably log a confirmation when we're using a proxy

sdk/core/src/worker/index.spec.js
```diff
@@ -52,6 +53,9 @@ vi.mock( './interceptors.js', () => ( { initInterceptors: initInterceptorsMock }
 const startCatalogMock = vi.fn().mockResolvedValue( undefined );
 vi.mock( './start_catalog.js', () => ( { startCatalog: startCatalogMock } ) );
 
+const bootstrapFetchProxyMock = vi.fn();
+vi.mock( './proxy.js', () => ( { bootstrapFetchProxy: bootstrapFetchProxyMock } ) );
```
catsby, line 57: This doesn't have any matching `expect( bootstrapFetchProxyMock ).toHaveBeenCalledWith()`

catsby (commented):
Mostly questions, but I do think we need to catch any crashing issues with bad URL formats

.changeset/proxy-support.md
```diff
@@ -0,0 +1,5 @@
+---
+"@outputai/core": patch
+---
```
catsby, line 4: we need to include CLI in this changeset as well, right? Otherwise CLI wont get a new version

sdk/cli/src/utils/proxy.ts
bnchrch: Done — added JSDoc describing what it does, which env vars it reads, and when to call it. 8a10656d

sdk/cli/src/utils/proxy.ts
bnchrch: Good catch — wrapped the value in `new URL()` with a try/catch in both the CLI and worker versions. Malformed proxy URL now logs a warning and returns (no crash, no dispatcher swap). Added a test for it in each file. 8a10656d

sdk/core/src/worker/proxy.js
bnchrch: Thought about this — I'd like to keep the duplication for now. The bodies are ~15 lines, the two packages have different module styles (TS for CLI, JS for core) and different logger conventions (console for CLI since it runs before the logger is initialized, `createChildLogger` for the worker). Extracting into a shared package just to dedupe would pull in a new cross-package dependency for a short, stable snippet. If the logic grows or starts drifting in meaningful ways I'll revisit — happy to open a follow-up if you'd prefer.

sdk/core/src/worker/configs.js
bnchrch: Yeah good call — Temporal's `targetHost` is a bare `host:port` (it's a native HTTP CONNECT option, not a URL), so a scheme would fail at connect time with a confusing error. Added a zod refinement that rejects any value containing `://` with a message pointing users at the right format. 8a10656d

sdk/core/src/worker/index.js
bnchrch: Added `log.info('Using gRPC proxy', { targetHost: grpcProxy })` right before the connect call, guarded by the `proxy` variable so it only fires when a proxy is actually configured. 8a10656d

sdk/core/src/worker/index.spec.js
bnchrch: Added `expect( bootstrapFetchProxyMock ).toHaveBeenCalled()` in the main assertion block. 8a10656d

.changeset/proxy-support.md
bnchrch: Right — CLI's `bin/run.js` imports from the built `dist`, so the CLI itself carries the fetch proxy code and needs a bump. Added `@outputai/cli: patch` to the changeset. 8a10656d

sdk/core/src/worker/proxy.js
```diff
@@ -0,0 +1,32 @@
...
+ * through an HTTP/HTTPS proxy when standard proxy env vars are set
+ * (`HTTPS_PROXY`, `https_proxy`, `HTTP_PROXY`, `http_proxy`). No-op when
+ * none are set. Invalid URLs are logged and skipped so the worker keeps
+ * running.
+ *
+ * Call once at worker startup, before any network activity.
+ */
+export const bootstrapFetchProxy = () => {
+  const proxyUrl = process.env.HTTPS_PROXY || process.env.https_proxy ||
+    process.env.HTTP_PROXY || process.env.http_proxy;
```
szanata, line 17: In which scenarios the env vars would be lower case?

szanata (approved):

sdk/core/src/worker/proxy.js
bnchrch: Lowercase `http_proxy` / `https_proxy` is actually the original Unix convention — curl, wget, git, and most Unix tooling have used lowercase as the primary form for decades. Uppercase came later and is common but not universal. After the 2016 httpoxy CVE (CVE-2016-5385), some environments deliberately prefer lowercase because `HTTP_PROXY` collides with the CGI-derived `HTTP_*` header namespace and can be injected via a request.

More importantly: `undici`'s own `EnvHttpProxyAgent` (which we hand off to via `setGlobalDispatcher`) reads both cases internally. If we only checked uppercase here, undici would still proxy traffic when only `https_proxy` was set, but our validation/logging in `bootstrapFetchProxy` would miss it and we'd silently skip the warn-on-invalid-URL branch. Checking both keeps our validation aligned with what undici actually does.

PR #145, opened 2026-04-20
fix: API Workflow response was inconsistent between sync/async run

szanata:
## Summary

When executing an workflow via the API using `/run`, the "input" field was empty, whereas when getting the result via `/result` it was present.

This PR normalizes both to have the exact same fields and refactors the jsdoc to use `WorkflowResult` type.

## Test plan

[x] Integration tests
[x] Manually run

### DIFF
--- a/api/src/clients/temporal_client.js
+++ b/api/src/clients/temporal_client.js
@@ -35,22 +35,19 @@ const getCatalog = async ( { client, taskQueue } ) => {
 };
 
 /**
- * Map Temporal workflow execution status to user-friendly string
- * @param {string} statusName - The Temporal status name (e.g., 'RUNNING', 'COMPLETED')
+ * Maps Temporal workflow execution status to user-friendly string
+ * @param {string} status - The Temporal status name (e.g., 'RUNNING', 'COMPLETED')
  * @returns {string} User-friendly status string
  */
-const mapWorkflowStatus = statusName => {
-  const statusMap = {
-    RUNNING: 'running',
-    COMPLETED: 'completed',
-    FAILED: 'failed',
-    CANCELED: 'canceled',
-    TERMINATED: 'terminated',
-    TIMED_OUT: 'timed_out',
-    CONTINUED_AS_NEW: 'continued'
-  };
-  return statusMap[statusName] || statusName.toLowerCase();
-};
+const mapWorkflowStatus = status => ( {
+  RUNNING: 'running',
+  COMPLETED: 'completed',
+  FAILED: 'failed',
+  CANCELED: 'canceled',
+  TERMINATED: 'terminated',
+  TIMED_OUT: 'timed_out',
+  CONTINUED_AS_NEW: 'continued'
+}[status] || status.toLowerCase() );
 
 /**
  * Terminal Temporal workflow execution status codes.
@@ -78,7 +75,7 @@ const EventType = {
 };
 
 /**
- * Resolve a step name to the WORKFLOW_TASK_COMPLETED event ID to reset to.
+ * Resolves a step name to the WORKFLOW_TASK_COMPLETED event ID to reset to.
  * Scans the workflow history to find the activity matching the step, then locates
  * the workflow task completed event immediately after that activity finished.
  *
@@ -129,7 +126,7 @@ export const resolveResetEventId = ( events, stepName ) => {
 };
 
 /**
- * Extract the workflow input from a Temporal history object.
+ * Extracts the workflow input from a Temporal history object.
  * The first event is always WorkflowExecutionStarted, which contains the input payloads.
  *
  * @param {object} history - Temporal History object from handle.fetchHistory()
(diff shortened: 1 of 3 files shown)


api/src/clients/temporal_client.js
```diff
@@ -144,19 +144,33 @@ export const extractWorkflowInput = history => {
...
- * @param {string} status - The workflow status
- * @param {Object} [options] - Optional fields
- * @param {string|null} [options.runId] - The specific run id for this execution
- * @param {any} [options.input] - The original workflow input
- * @param {any} [options.output] - The workflow output
+ * A representation of a workflow execution result, including input, output, error and meta information
+ *
+ * @typedef {object} WorkflowResult
+ * @property {string} workflowId - The workflow execution id
+ * @property {'completed'|'failed'} status - The workflow status
```
catsby: Below we define a more comprehensive list: 

```
('completed'|'failed'|'canceled'|'terminated'|'timed_out'|'continued'|'unspecified')
```

is there a reason we only list `completed|failed` here?

api/src/clients/temporal_client.js
```diff
@@ -144,19 +144,33 @@ export const extractWorkflowInput = history => {
...
+ * @property {object|null} trace - Trace information
+ * @property {string|null} error - Error message if failed
+ */
+
+/**
+ * Build a WorkflowResult object
+ *
+ * @param {Object} args - Fields
+ * @param {string} options.workflowId - The workflow execution id
+ * @param {string} options.status - The workflow status
```
catsby: Should the `args` here be `options`, to match the descriptions below? 
```suggestion
 * @param {Object} options - Fields
 * @param {string} options.workflowId - The workflow execution id
 * @param {string} options.status - The workflow status
```

api/src/clients/temporal_client.js
```diff
@@ -144,19 +144,33 @@ export const extractWorkflowInput = history => {
...
+ * @param {string} options.runId - The specific run id for this execution
+ * @param {unknown} [options.output] - The workflow output
  * @param {object} [options.trace] - Trace information
  * @param {string} [options.error] - Error message if failed
- * @returns {Object} Standardized workflow response
+ * @returns {WorkflowResult}
  */
-const buildWorkflowResponse = ( workflowId, status, { runId = null, input = null, output = null, trace = null, error = null } = {} ) =>
-  ( { workflowId, runId, status, input, output, trace, error } );
+const buildWorkflowResult = ( { workflowId, status, input, runId, output = null, trace = null, error = null } ) =>
```
catsby: could the input be null here, or no? Do we need `input = null`?

api/src/clients/temporal_client.js
```diff
@@ -35,22 +35,19 @@ const getCatalog = async ( { client, taskQueue } ) => {
...
-};
+const mapWorkflowStatus = status => ( {
+  RUNNING: 'running',
+  COMPLETED: 'completed',
+  FAILED: 'failed',
+  CANCELED: 'canceled',
+  TERMINATED: 'terminated',
+  TIMED_OUT: 'timed_out',
+  CONTINUED_AS_NEW: 'continued'
+}[status] || status.toLowerCase() );
```
catsby, line 50: I've noticed some inconsistency with the usage of `mapWorkflowStatus`. `getWorkflowResult` (line ~369) uses this map and will translate `CONTINUED_AS_NEW` to `continued` in the response. `getWorkflowStatus` (line ~302) doesn't use the `mapWorkflowStatus` and just returns the Temporal status lowercased. Then `runWorkflow` only returns `'completed'` or `'failed'`. Maybe the `runWorkflow` bit is intentional but I think we should make the other 2 consistent while we're here

api/src/clients/temporal_client.js
```diff
@@ -144,19 +141,37 @@ export const extractWorkflowInput = history => {
...
- * @param {string} workflowId - The workflow execution id
- * @param {string} status - The workflow status
- * @param {Object} [options] - Optional fields
- * @param {string|null} [options.runId] - The specific run id for this execution
- * @param {any} [options.input] - The original workflow input
- * @param {any} [options.output] - The workflow output
- * @param {object} [options.trace] - Trace information
- * @param {string} [options.error] - Error message if failed
- * @returns {Object} Standardized workflow response
+ * @typedef {'canceled'|'completed'|'continued_as_new'|'failed'|'running'|'terminated'|'timed_out'|'unspecified'} WorkflowExecutionStatus
```
catsby: nit: I don't think `unspecified` is valid, at least not coming from Temporal, is it?

szanata: No. Removed

api/src/clients/temporal_client.js
szanata: Not introduced by this PR, but it will be fixed as well.

bnchrch (approved):

catsby (approved):

PR #146, opened 2026-04-20
feat(cli|agent|ci): Add support for migrate command and migration guide

bnchrch:
## Overview
image

- Adds `output migrate` — a CLI command that upgrades a project from one Output framework version to another.
- Adds the Claude slash command `/outputai:migrate` that the CLI wraps; the skill fetches guidance at runtime so it never goes stale.
- Auto-generates `docs.output.ai/changelog` and `docs.output.ai/migrations` from a single source of truth.
- Source of truth is `docs/guides/data/releases.json` — changesets land in it on release, Mintlify snippets are rendered from it.
- Backfills the full v0.1.1 → v0.1.12 release history.

### When a user runs \`output migrate\`

1. The CLI reads their project's \`@outputai/*\` version.
2. The CLI invokes the \`/outputai:migrate\` Claude skill.
3. The skill fetches \`docs.output.ai/migrations\`, finds the section for their FROM → TO range, applies the steps, bumps dependencies, runs their type checker.

### How the migration guide gets generated

Follow the data:

1. **Changeset** — a contributor writes a changeset; for breaking changes they add a \`## Migration\` section to the body.
2. **JSON** — on release, \`ops/bump_release.sh\` appends a new entry (summary + optional migration body) to \`docs/guides/data/releases.json\`, then runs \`pnpm changeset version\`.
3. **Snippets** — \`./ops/regenerate_docs.sh\` reads the JSON and rewrites two Mintlify snippets: \`snippets/changelog.mdx\` and \`snippets/migration_guides.mdx\`.
4. **Pages** — hand-written \`changelog/index.mdx\` and \`migrations/index.mdx\` import those snippets. Mintlify's GitHub App deploys on merge.


szanata (commented):
I am torn on having migration inside the changeset itself. Migration guides can get pretty long. Wouldnt be better to have a each migration under its own page on the guides docs?

Thus each page would be written as guide without the need to extract and transform from changesets.

ops/regenerate_docs.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Regenerates Mintlify MDX pages from docs/guides/data/releases.json.
+#
+# Called locally after hand-editing releases.json, and from the
+# docs_regenerate GitHub workflow when a PR modifies the JSON.
```
szanata: I think this type command can get out of date pretty quick. A file should not know/care where it is called.

szanata (approved):

sdk/cli/src/services/claude_client.ts
```diff
@@ -40,11 +40,24 @@ date: <plan-date>
...
+
+4. Confirm the planned changes with the user before editing files.
+
+5. After you mark all todos as complete, provide a summary of which files changed and whether the type check passed.
 `
 } as const;
 
 const PLAN_COMMAND = 'outputai:plan_workflow';
 const BUILD_COMMAND = 'outputai:build_workflow';
+const MIGRATE_COMMAND = 'output-migrate';
```
catsby, line 63: Do we have a convention for naming commands between `outputai:` and `output-` styles?

sdk/cli/src/commands/migrate.ts
```diff
@@ -0,0 +1,49 @@
...
+      required: false
+    } ),
+    notes: Flags.string( {
+      char: 'n',
+      description: 'Extra guidance passed through to the migration agent.',
+      required: false
+    } )
+  };
+
+  async run(): Promise<void> {
```
catsby, line 33: Do we need some kind of `try/catch` or `async catch( error: Error )` to handle if this throws somewhere?

.changeset/migrate-command.md
```diff
@@ -0,0 +1,14 @@
...
+
+The command detects the current framework version in the project's `package.json`, fetches the matching migration guide from `docs.output.ai/migrations`, applies the changes, bumps dependencies, and runs the project's type checker. If no guide exists for the exact version pair, it falls back to the migrations index and chains the guides that cover the range.
+
+This is paired with two repo-level additions that wire migration content into the docs site:
+
+- A new `/outputai:migrate` Claude Code slash command (shipped via the outputai plugin marketplace) that carries the migration logic. The command fetches guides at runtime instead of embedding version-specific steps in the prompt, so the skill stays small and the docs are the source of truth.
+- A changelog and migration guide generator (`ops/generate_docs_from_changesets.mjs`, wired into `ops/bump_release.sh`) that writes each release to `docs/guides/changelog/` as a Mintlify `<Update>` block and, when a changeset body includes a `## Migration` section, writes a dedicated guide under `docs/guides/migrations/v{from}-to-v{to}.mdx` and registers it in `docs.json`.
```
catsby: I don't see a `ops/generate_docs_from_changesets.mjs` file, was that omitted by mistake, or is this changeset outdated?

docs/guides/scripts/build_releases_json.mjs
```diff
@@ -0,0 +1,98 @@
...
+const scriptDir = path.dirname( fileURLToPath( import.meta.url ) );
+const guidesDir = path.resolve( scriptDir, '..' );
+const repoRoot = path.resolve( guidesDir, '../..' );
+
+const paths = {
+  changesetsDir: path.join( repoRoot, '.changeset' ),
+  releasesJson: path.join( guidesDir, 'data/releases.json' )
+};
+
+function parseFlag( name ) {
```
catsby: Is this something we could use `node:util`'s `parseArgs()`? 

- https://nodejs.org/api/util.html#utilparseargsconfig

docs/guides/scripts/lib/store.mjs
```diff
@@ -0,0 +1,49 @@
...
+    if ( err.code === 'ENOENT' ) {
+      return { ...EMPTY_DATA };
+    }
+    throw err;
+  }
+}
+
+export async function writeReleasesJson( path, data ) {
+  const payload = {
+    generatedAt: new Date().toISOString(),
```
catsby: do we ever use `generatedAt`?

catsby (commented):
A few questions/suggestions

ops/regenerate_docs.sh
bnchrch: Good call — tightened the header in 022986d0 to just describe what the script does.

sdk/cli/src/services/claude_client.ts
bnchrch: Yes — convention is:

- `outputai:<name>` — plugin **commands** (files under `coding_assistants/claude/plugins/outputai/commands/`). The `outputai:` prefix is the Claude Code plugin namespace.
- `output-<kebab-name>` — plugin **skills** (files under `coding_assistants/claude/plugins/outputai/skills/<name>/SKILL.md`). Skills surface as top-level slash commands without the namespace.

Documented inline above the `*_COMMAND` constants in 022986d0 so future readers don't have to infer it.

sdk/cli/src/commands/migrate.ts
bnchrch: Matches the pattern in `plan.ts` and `generate.ts` — oclif's `Command` base class catches unhandled rejections and renders them via `this.error`, and `ClaudeInvocationError` / `UserCancelledError` already produce human-readable messages. If we want nicer per-error-type UX, that's a cross-cutting change across all three commands; probably belongs in a follow-up PR.

.changeset/migrate-command.md
bnchrch: Good catch — changeset was stale. Rewrote it in 022986d0 to describe the current state of the PR (no more reference to the deleted `ops/generate_docs_from_changesets.mjs`).

docs/guides/scripts/build_releases_json.mjs
bnchrch: Great suggestion — swapped to `parseArgs` from `node:util` in 022986d0.

docs/guides/scripts/lib/store.mjs
bnchrch: Nope — written but never read. Dropped from both `store.mjs` and the committed `releases.json` in 022986d0.

bnchrch: removed

catsby (approved):

PR #151, opened 2026-04-21
feat(api,cli): rename task-queue to catalog across workflow commands

bnchrch:
## Summary

Use `catalog` as the public name for the routing/filtering target across the CLI and HTTP API. Internally the value is still a Temporal task queue — only the user-facing name changes. Addresses issue #9 in [OUT-411](https://linear.app/growthx/issue/OUT-411).

## CLI

- `output workflow runs list` gains `--catalog`/`-c` (with `OUTPUT_CATALOG_ID` env fallback) — scopes listed runs to a single worker's catalog/session.
- `output workflow run` and `output workflow start` rename the routing flag to `--catalog`/`-c`. The previous `--task-queue` and `-q` are kept as deprecated oclif aliases (`deprecateAliases: true` emits a warning when used).

## HTTP API

- `GET /workflow/runs?catalog=...` filters runs by catalog.
- `POST /workflow/run` and `POST /workflow/start` accept a `catalog` body field. The previous `taskQueue` body field is still accepted as a deprecated alias; if both are sent, `catalog` wins. The API logs `Deprecated body field { field: 'taskQueue', successor: 'catalog', route }` when `taskQueue` is used.

## Internal

- `temporal_client.runWorkflow` / `startWorkflow` options renamed `taskQueue` → `catalog`; `listWorkflowRuns` already takes `catalog`.
- The wrapper still calls Temporal's `client.workflow.start({ taskQueue: catalog, ... })` because Temporal's API uses that name.
- `resolveWorkflowName(catalogDef, workflowName, catalog)` clarified to distinguish the catalog *definition* (workflows array) from the catalog *name*.

## Files

| File | Change |
|---|---|
| `api/src/clients/temporal_client.js` | Rename option `taskQueue` → `catalog` on `runWorkflow`/`startWorkflow`/`listWorkflowRuns`. Inner Temporal calls keep `taskQueue`. |
| `api/src/index.js` | Add `resolveCatalogField` helper. Zod schemas for `/workflow/run` and `/workflow/start` accept both `catalog` and `taskQueue`. Swagger updated. |
| `api/src/index.spec.js` | New tests: catalog forwarding, taskQueue→catalog forwarding + deprecation warning, catalog wins when both sent. |
| `api/src/clients/temporal_client.spec.js` | Updated `Workflow alias resolved` log assertion (key is now `catalog`). |
| `sdk/cli/src/commands/workflow/run.ts`, `start.ts`, `runs/list.ts` | New `catalog` flag with `aliases: ['task-queue']`, `charAliases: ['q']`, `deprecateAliases: true`. |
| `sdk/cli/src/commands/workflow/run.spec.ts`, `start.spec.ts` | Flag-existence assertions updated. |
| `sdk/cli/src/services/workflow_runs.ts`, `sdk/cli/src/api/generated/api.ts` | Generated/derived from updated openapi. |
| `api/openapi.json`, `docs/guides/openapi.json` | Regenerated. |
| `docs/guides/workflows/running.mdx`, `docs/guides/packages/cli.mdx` | Flag tables updated. |
| `.changeset/catalog-aware-runs-listing.md` | Patch bump for `output-api` and `@outputai/cli`. |

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run build:packages` — regenerates openapi.json, copies to docs/guides, regens cli api.ts
- [x] `./ops/extra_files_error.sh` — passes
- [x] `npx vitest run --exclude '.claude/worktrees/**' …` — 1605 passed (5 new tests vs main)
- [x] End-to-end against `test_workflows` (worker on catalog `sepcat`, dev stack on shifted ports):
  - `output workflow run simple --catalog sepcat --input '...'` → completed
  - `output workflow run simple --catalog main --input '...'` → 503 `CatalogNotAvailableError`
  - `output workflow run simple -c sepcat ...` → completed
  - `output workflow start simple --catalog sepcat ...` → workflowId
  - `output workflow run simple --task-queue sepcat ...` → completed + oclif warning `Warning: The "--task-queue" flag has been deprecated. Use "--catalog | -c" instead.`
  - `output workflow run simple -q sepcat ...` → completed
  - `output workflow runs list -c sepcat` → only sepcat runs
  - `output workflow runs list -c main` → empty
  - `curl -X POST .../workflow/run -d '{"taskQueue":"sepcat", ...}'` → completed + API log `Deprecated body field { field: 'taskQueue', successor: 'catalog', route: '/workflow/run' }`
  - `curl -X POST .../workflow/run -d '{"catalog":"sepcat", ...}'` → completed (no warning)

### DIFF
--- a/api/src/clients/temporal_client.js
+++ b/api/src/clients/temporal_client.js
@@ -629,15 +629,21 @@ export default {
        *
        * @param {Object} [options] - Optional configuration
        * @param {string} [options.workflowType] - Filter by workflow type/name
+       * @param {string} [options.taskQueue] - Filter by Temporal task queue name
        * @param {number} [options.limit=100] - Maximum number of runs to return
        * @returns {WorkflowRunsListResult}
        */
       async listWorkflowRuns( options = {} ) {
-        const { workflowType, limit = 100 } = options;
+        const { workflowType, taskQueue, limit = 100 } = options;
 
-        const query = workflowType ?
-          `WorkflowType = "${workflowType}"` :
-          undefined;
+        const conditions = [];
+        if ( workflowType ) {
+          conditions.push( `WorkflowType = "${workflowType}"` );
+        }
+        if ( taskQueue ) {
+          conditions.push( `TaskQueue = "${taskQueue}"` );
+        }
+        const query = conditions.length > 0 ? conditions.join( ' AND ' ) : undefined;
 
         const executions = await takeFromAsyncIterable(
           client.workflow.list( { query } ),
--- a/sdk/cli/src/commands/workflow/runs/list.ts
+++ b/sdk/cli/src/commands/workflow/runs/list.ts
@@ -67,6 +67,7 @@ export default class WorkflowRunsList extends Command {
     '<%= config.bin %> <%= command.id %>',
     '<%= config.bin %> <%= command.id %> simple',
     '<%= config.bin %> <%= command.id %> simple --limit 10',
+    '<%= config.bin %> <%= command.id %> --catalog my-catalog',
     '<%= config.bin %> <%= command.id %> --format json',
     '<%= config.bin %> <%= command.id %> --format table'
   ];
@@ -79,6 +80,11 @@ export default class WorkflowRunsList extends Command {
   };
 
   static override flags = {
+    catalog: Flags.string( {
+      char: 'c',
+      description: 'Filter runs by catalog (defaults to OUTPUT_CATALOG_ID)',
+      env: 'OUTPUT_CATALOG_ID'
+    } ),
     limit: Flags.integer( {
       char: 'l',
       description: 'Maximum number of runs to return',
@@ -97,6 +103,7 @@ export default class WorkflowRunsList extends Command {
 
     const { runs, count } = await fetchWorkflowRuns( {
       workflowType: args.workflowName,
+      catalog: flags.catalog,
       limit: flags.limit
     } );
 
--- a/.changeset/catalog-aware-runs-listing.md
+++ b/.changeset/catalog-aware-runs-listing.md
@@ -1,0 +1,12 @@
+---
+"@outputai/cli": patch
+"output-api": patch
+---
+
+Use `catalog` as the public name for the routing/filtering target across the CLI and HTTP API:
+
+- `output workflow runs list` gains `--catalog`/`-c` (with `OUTPUT_CATALOG_ID` env fallback) and `GET /workflow/runs` accepts `?catalog=...`, scoping listed runs to a single worker's catalog/session.
+- `output workflow run` and `output workflow start` rename the routing flag to `--catalog`/`-c`. The previous `--task-queue` and `-q` are kept as deprecated aliases (oclif emits a warning when used).
+- `POST /workflow/run` and `POST /workflow/start` accept a `catalog` body field; the previous `taskQueue` field is still accepted as a deprecated alias and the API logs a deprecation warning when it is used.
+
+Internally the value is still a Temporal task queue — only the user-facing name changes.
--- a/api/openapi.json
+++ b/api/openapi.json
@@ -545,9 +545,14 @@
                     "type": "string",
                     "description": "(Optional) The workflowId to use. Must be unique"
                   },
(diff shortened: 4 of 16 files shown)


szanata (approved):

api/src/clients/temporal_client.js
```diff
@@ -629,15 +629,21 @@ export default {
        *
        * @param {Object} [options] - Optional configuration
        * @param {string} [options.workflowType] - Filter by workflow type/name
+       * @param {string} [options.catalog] - Filter by catalog (matches Temporal task queue)
        * @param {number} [options.limit=100] - Maximum number of runs to return
        * @returns {WorkflowRunsListResult}
        */
       async listWorkflowRuns( options = {} ) {
-        const { workflowType, limit = 100 } = options;
+        const { workflowType, catalog, limit = 100 } = options;
```
catsby: Do we need to sanitize `catalog` and `workflowType`? They're straight from input into Temporal query, I don't know if Temporal is susceptible to sql injection like attacks, but still. I know `workflowType` is not introduced in this PR so this is already a pre-existing concern

sdk/cli/src/commands/workflow/runs/list.ts
```diff
@@ -79,6 +80,11 @@ export default class WorkflowRunsList extends Command {
   };
 
   static override flags = {
+    catalog: Flags.string( {
+      char: 'c',
+      description: 'Filter runs by catalog (defaults to OUTPUT_CATALOG_ID)',
+      env: 'OUTPUT_CATALOG_ID'
+    } ),
```
catsby, line 87: For `workflow start|run` we use `OUTPUT_CATALOG_ID` or `-q` for `task-queue`, but here we use `OUTPUT_CATALOG_ID` or `-c` for `catalog`, which seems inconsistent

api/src/clients/temporal_client.js
```diff
@@ -629,15 +629,21 @@ export default {
...
-          `WorkflowType = "${workflowType}"` :
-          undefined;
+        const conditions = [];
+        if ( workflowType ) {
+          conditions.push( `WorkflowType = "${workflowType}"` );
+        }
+        if ( catalog ) {
+          conditions.push( `TaskQueue = "${catalog}"` );
+        }
+        const query = conditions.length > 0 ? conditions.join( ' AND ' ) : undefined;
```
catsby, line 646: should we add more tests to `temporal_client.spec.js`? From what I can tell, it only tests if `workflowType` has a value, but with this we've introduced more scenarios: 

- workflowType only ✅ 
- catalog only - is this possible, practically speaking?
- both joined (joined by an `AND`)
- neither which should result in... all results?

api/src/clients/temporal_client.js
bnchrch: Good catch. Restricted `catalog`, `taskQueue`, and `workflowType` to the same `[a-z0-9_.@-]` regex already enforced on `OUTPUT_CATALOG_ID` in `@outputai/core` (see `sdk/core/src/worker/configs.js`). Any value that reaches the visibility query is now guaranteed not to break out of the quoted literal. Added API tests for the rejection on `/workflow/run` and `/workflow/runs`. Commit f5e83dfe.

sdk/cli/src/commands/workflow/runs/list.ts
bnchrch: Resolved in 3dc59017 — `workflow run` and `workflow start` now use `--catalog`/`-c` as the canonical flag too, with `--task-queue` and `-q` kept as deprecated oclif aliases (`deprecateAliases: true` emits a warning). All three commands are consistent on `-c` for catalog now.

api/src/clients/temporal_client.js
bnchrch: Added in f5e83dfe. The `listWorkflowRuns` describe block now covers all four query branches: `workflowType` only (already covered), `taskQueue` only, both (AND-joined), and neither (no query → all runs). Re catalog only — yes, that case is realistic: in the sandbox/multi-session scenario this PR is built for, an agent listing its own catalog without filtering by workflow type is the common path.

catsby (approved):

PR #158, opened 2026-04-23
fix(trace): Improving trace performance and fixing memory usage [OUT-423] [OUT-235]

szanata:
### Fix trace local processor (perf/memory)

In order to prevent `RangeError: Invalid string length` and improve resource usages while processing local traces, some improvements were added:
- Removed reading the `.trace` file, parsing it, building the tree, and writing the JSON file on every span. Now it only appends on new entries, and only does the "read → build → write" when the span is the root workflow end or error.
  That removes the O(n²) work and repeated JSON.stringify(..., 2) allocations, which is where we the `RangeError` was happening.
- Added `safeFormatJSON()` function to format the JSON string written to the final trace file. This function will serialize to a plain string first, and if its size is manageable (_<50mb_), it will then try to format with indentation. If a `RangeError` happens in this step, it falls back to the plain string again.

## Test plan

[x] Unit test
[x] e2e test
[x] Manually invoking workflows

### DIFF
--- a/.changeset/clever-rats-build.md
+++ b/.changeset/clever-rats-build.md
@@ -1,0 +1,5 @@
+---
+"@outputai/core": patch
+---
+
+Optimizing local trace to use less memory and avoid "RangeError: Invalid string length"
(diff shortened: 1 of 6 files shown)


bnchrch (approved):

szanata:
https://linear.app/growthx/issue/OUT-423/local-trace-processor-hits-invalid-string-length

PR #165, opened 2026-04-24
feat(cli): offer to initialize git in generated project

bnchrch:
## Summary

- `output init` now prompts "Initialize a git repository?" (default yes) after `npm install` and runs `git init` in the new project folder when accepted
- Adds `--skip-git` flag, mirroring the existing `--skip-env` pattern for non-interactive / scripted use
- Reuses the existing `executeCommandWithMessages` wrapper so a missing `git` binary or non-zero exit downgrades to a warning instead of failing the whole scaffold
- `.gitignore` is already shipped by the template processor, so the fresh repo immediately ignores `node_modules/`, `.env`, credential `.key` files, etc.

Ref: [OUT-396](https://linear.app/growthx/issue/OUT-396/offer-to-initialize-git-in-generated-project)

## Test plan

- [x] `npm run lint` passes
- [x] `tsc --noEmit` on `@outputai/cli` passes
- [x] Smoke test `--skip-git`: `output init demo --skip-env --skip-git` — no `.git` directory created, no prompt shown
- [x] Smoke test default-accept (non-interactive): `output init demo --skip-env` — `.git` directory created, `git status` clean of ignored paths
- [ ] Manual interactive run: accept the prompt, then decline, confirm both paths behave correctly

## Out of scope

- Auto-cd'ing into the new folder (intentionally deferred — a child process can't change the parent shell's cwd without a shell wrapper, and the ticket's primary ask is the git init offer)

### DIFF
--- a/.changeset/long-meals-cover.md
+++ b/.changeset/long-meals-cover.md
@@ -2,4 +2,5 @@
 "@outputai/cli": patch
 ---
 
-Offer to initialize a git repository when running `output init`. Adds a `--skip-git` flag to opt out in non-interactive / scripted use.
+- Offer to initialize a git repository when running `output init`. Adds a `--skip-git` flag to opt out in non-interactive / scripted use.
+- Fix `--yes` / `--non-interactive` being rejected as "Nonexistent flag" by oclif's per-command parser when passed to any command.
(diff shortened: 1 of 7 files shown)


szanata (approved):
Good ideia

PR #167, opened 2026-04-24
fix: Use lazy getters to defer processing ENV until after reading .env file

catsby:
## Summary

`config.ts` reads the `process.env` at ES module evaluation time. But `.env` isn't loaded yet at that point, so all env-var-based config set in `.env` files are silently ignored. Overriding default values only works when exported as shell environment variables.

The result is if you set `OUTPUT_API_URL` and `OUTPUT_API_AUTH_TOKEN` in the `.env` file to point to a deployed Output API and try something like `output workflow list`, you'd get an error because it couldn't connect to `localhost:3001` unless it was running locally, in which case you'd get results for the local instance and not the one specified in `OUTPUT_API_URL`

### DIFF
--- a/.changeset/soft-teeth-mix.md
+++ b/.changeset/soft-teeth-mix.md
@@ -1,0 +1,5 @@
+---
+"@outputai/cli": patch
+---
+
+Fix issue where values in .env files were silently ignored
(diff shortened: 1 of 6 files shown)


bnchrch (approved):

szanata (commented):
This is a clever solution, but did you consider loading the config with a config.load() during startup? Or in this case this isnt possible?

szanata (approved):

PR #168, opened 2026-04-24
fix(cli): accept --yes and --non-interactive flags without parser error

bnchrch:
## Problem

`output init --yes` (and any other command invoked with `--yes` or `--non-interactive`) fails with:

```
 ›   Error: Nonexistent flag: --yes
 ›   See more help with --help
```

The init hook at [sdk/cli/src/hooks/init.ts](sdk/cli/src/hooks/init.ts) inspects `process.argv` for `--yes` / `--non-interactive` and flips the global non-interactive state. But those flags aren't declared on any oclif command, so oclif's per-command strict parser rejects them a moment later.

Introduced in [#137](https://github.com/growthxai/output/pull/137) — the manual smoke test checkbox (`echo '' | npx output init --yes test-project`) was never completed, which is why this slipped through.

## Solution

After the init hook detects either flag, strip it from `process.argv` so oclif's command-level parser never sees it. Small, isolated 2-line change with no API surface impact.

## Test plan

- [x] `npm run lint` passes
- [x] `vitest run src/hooks/init.spec.ts` passes (6 tests, 3 new covering argv stripping + flip-and-no-flag cases)
- [x] Smoke test: `output init --yes --help` and `output --yes init --help` both succeed locally on the linked dev build (previously errored with "Nonexistent flag: --yes")

## Out of scope

- Declaring `--yes` / `--non-interactive` as proper globalFlags (larger refactor). This patch is the minimal fix to unblock the existing documented usage.

### DIFF
--- a/.changeset/two-horses-cry.md
+++ b/.changeset/two-horses-cry.md
@@ -1,0 +1,5 @@
+---
+"@outputai/cli": patch
+---
+
+Fix `--yes` / `--non-interactive` flags being rejected as "Nonexistent flag" by oclif's per-command parser. The init hook now strips these global flags from `process.argv` after flipping the non-interactive state, so they reach the hook but not the command-level flag validator.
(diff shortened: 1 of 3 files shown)


bnchrch:
Folded into #165 — keeping both changes in one PR.

PR #169, opened 2026-04-24
feat(core,http): Trace event attributes and http addRequestCost [OUT-428]

szanata:
### Add event attributes
Adding a new tracing feature `Tracing.addEventAttribute()` to add "attributes" to trace events. This is a new object field for each trace event in the trace file, that accepts any kind of values:

```js
[
  {
    "id": "generateText-1777069221681",
    "kind": "llm",
    "name": "generateText",
    "startedAt": 1777069221681,
    "endedAt": 1777069225372,
    "input": {
      ...
    },
    "output": {
      ...
    },
    "children": [],
    "attributes": {
      "flavor": "vanilla" // <-- ATTRIBUTE
    }
  }
],
```

### Add HTTP request cost
Adding a new method to the `@outputai/http` module `addRequestCost()`, that allows user to add a cost attribute to an HTTP trace event using the response (both fetch's `Response` or ky's `KyResponse`):

```js
import { fetch } from '@outputai/http';

const response = await fetch( https://foo );
addRequestCost( response, {
  total: 32,
  components: [
    { name: 'api', value: 24 },
    { name: 'caching', value: 12 } 
  ]
} );
```

The cost format must respect:
```ts
export type RequestCost = {
  total: number,
  components?: [
    {
      name: string,
      value: number
    }
  ]
```

## Refactoring LLM cost

Refactoring `@outputai/llm` to attach LLM costs using `Tracing.addEventAttribute()`, with the same format used by the HTTP module.

## Test plan

- [x] Unit tests
- [x] e2e tests
- [x] Manual tests

### DIFF
--- a/sdk/core/src/activity_integration/tracing.d.ts
+++ b/sdk/core/src/activity_integration/tracing.d.ts
@@ -1,39 +1,37 @@
 /**
- * Record the start of an event for the current workflow.
+ * Creates a new event.
  *
- * @param args - Event information
- * @param args.id - A unique id for the Event, must be the same across all phases: start, end, error.
+ * @param args
+ * @param args.id - A unique id for the Event.
  * @param args.kind - The kind of Event, like HTTP, DiskWrite, DBOp, etc.
- * @param args.name - The human friendly name of the Event: query, request, create.
- * @param args.details - Arbitrary metadata associated with this phase (e.g., payloads, summaries).
+ * @param args.name - The human-friendly name of the Event: query, request, create.
+ * @param args.details - Arbitrary data to add to this event, it will be used as the "input" field.
  */
 export declare function addEventStart( args: { id: string; kind: string; name: string; details: unknown } ): void;
 
 /**
- * Record the end of an event.
+ * Concludes an event.
  *
- * @param args - Event information
- * @param args.id - Identifier matching the event's start phase.
- * @param args.details - Arbitrary metadata associated with this phase (e.g., results, response body).
+ * @param args
+ * @param args.id - The id of the event to conclude.
+ * @param args.details - Arbitrary data to add to this event, it will be used as the "output" field.
  */
 export declare function addEventEnd( args: { id: string; details: unknown } ): void;
 
 /**
- * Record the error in an event.
+ * Concludes an event with an error.
  *
- * @param args - Event metadata for the error phase.
- * @param args.id - Identifier matching the event's start phase.
- * @param args.details - Arbitrary metadata associated with this phase, possible error info.
+ * @param args
+ * @param args.id - The id of the event to conclude.
+ * @param args.details - Arbitrary data to add to this event, it will be used as the "error" field.
  */
 export declare function addEventError( args: { id: string; details: unknown } ): void;
 
 /**
- * Add an attribute to an event.
+ * Adds an attribute to an event.
  *
- * Use the same id as the start phase to correlate phases.
- *
- * @param args - Arguments for the attribute phase.
- * @param args.eventId - The event id
+ * @param args
+ * @param args.eventId - The id of the event to attach the attribute to.
  * @param args.name - The attribute name
  * @param args.value - The attribute value
  */
--- a/sdk/core/src/activity_integration/tracing.d.ts
+++ b/sdk/core/src/activity_integration/tracing.d.ts
@@ -36,3 +36,10 @@ export declare function addEventError( args: { id: string; details: unknown } ):
  * @param args.value - The attribute value
  */
 export declare function addEventAttribute( args: { eventId: string; name: string, value: unknown } ): void;
+
+/**
+ * Known attributes.
+ */
+export declare const Attribute: {
+  COST: 'cost';
+};
--- a/sdk/core/src/activity_integration/tracing.d.ts
+++ b/sdk/core/src/activity_integration/tracing.d.ts
@@ -32,7 +32,7 @@ export declare function addEventError( args: { id: string; details: unknown } ):
  *
  * Use the same id as the start phase to correlate phases.
  *
- * @param args - Event metadata for the error phase.
+ * @param args - Arguments for the attribute phase.
  * @param args.eventId - The event id
  * @param args.name - The attribute name
  * @param args.value - The attribute value
--- a/sdk/core/src/activity_integration/tracing.d.ts
+++ b/sdk/core/src/activity_integration/tracing.d.ts
@@ -1,5 +1,5 @@
 /**
- * Record the start of an event on the default trace for the current workflow.
+ * Record the start of an event for the current workflow.
  *
  * @param args - Event information
  * @param args.id - A unique id for the Event, must be the same across all phases: start, end, error.
@@ -10,9 +10,7 @@
 export declare function addEventStart( args: { id: string; kind: string; name: string; details: unknown } ): void;
 
 /**
- * Record the end of an event on the default trace for the current workflow.
- *
- * Use the same id as the start phase to correlate phases.
+ * Record the end of an event.
  *
  * @param args - Event information
  * @param args.id - Identifier matching the event's start phase.
@@ -21,12 +19,22 @@ export declare function addEventStart( args: { id: string; kind: string; name: s
 export declare function addEventEnd( args: { id: string; details: unknown } ): void;
 
 /**
- * Record an error for an event on the default trace for the current workflow.
- *
- * Use the same id as the start phase to correlate phases.
+ * Record the error in an event.
  *
  * @param args - Event metadata for the error phase.
  * @param args.id - Identifier matching the event's start phase.
  * @param args.details - Arbitrary metadata associated with this phase, possible error info.
  */
 export declare function addEventError( args: { id: string; details: unknown } ): void;
+
+/**
+ * Add an attribute to an event.
+ *
+ * Use the same id as the start phase to correlate phases.
+ *
+ * @param args - Event metadata for the error phase.
+ * @param args.eventId - The event id
+ * @param args.name - The attribute name
+ * @param args.value - The attribute value
+ */
+export declare function addEventAttribute( args: { eventId: string; name: string, value: unknown } ): void;
--- a/sdk/core/src/activity_integration/tracing.js
+++ b/sdk/core/src/activity_integration/tracing.js
@@ -1,46 +1,46 @@
-import { addEventPhaseWithContext, EventPhase } from '#tracing';
+import { addEventActionWithContext, EventAction } from '#tracing';
 
 /**
- * Adds the start phase of a new event at the default trace for the current workflow.
+ * Creates a new event.
  *
  * @param {object} args
- * @param {string} args.id - A unique id for the Event, must be the same across all phases: start, end, error.
+ * @param {string} args.id - A unique id for the Event.
  * @param {string} args.kind - The kind of Event, like HTTP, DiskWrite, DBOp, etc.
- * @param {string} args.name - The human friendly name of the Event: query, request, create.
- * @param {object} args.details - All details attached to this Event Phase. DB queried records, HTTP response body.
+ * @param {string} args.name - The human-friendly name of the Event: query, request, create.
+ * @param {object} args.details - Arbitrary data to add to this event, it will be used as the "input" field.
  * @returns {void}
  */
 export const addEventStart = ( { id, kind, name, details } ) =>
-  addEventPhaseWithContext( EventPhase.START, { kind, name, details, id } );
+  addEventActionWithContext( EventAction.START, { kind, name, details, id } );
 
 /**
- * Adds the end phase for an event using its id.
+ * Concludes an event.
  *
  * @param {object} args
- * @param {string} args.id - A unique id for the Event, must be the same across all phases: start, end, error.
- * @param {object} args.details - All details attached to this Event Phase. DB queried records, HTTP response body.
+ * @param {string} args.id - The id of the event to conclude.
+ * @param {object} args.details - Arbitrary data to add to this event, it will be used as the "output" field.
  * @returns {void}
  */
-export const addEventEnd = ( { id, details } ) => addEventPhaseWithContext( EventPhase.END, { id, details } );
+export const addEventEnd = ( { id, details } ) => addEventActionWithContext( EventAction.END, { id, details } );
 
 /**
- * Adds the error phase for an event using its id.
+ * Concludes an event with an error.
  *
  * @param {object} args
- * @param {string} args.id - A unique id for the Event, must be the same across all phases: start, end, error.
- * @param {object} args.details - All details attached to this Event Phase. DB queried records, HTTP response body.
+ * @param {string} args.id - The id of the event to conclude.
+ * @param {object} args.details - Arbitrary data to add to this event, it will be used as the "error" field.
  * @returns {void}
  */
-export const addEventError = ( { id, details } ) => addEventPhaseWithContext( EventPhase.ERROR, { id, details } );
+export const addEventError = ( { id, details } ) => addEventActionWithContext( EventAction.ERROR, { id, details } );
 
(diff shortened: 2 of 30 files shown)


sdk/core/src/activity_integration/tracing.d.ts
```diff
@@ -21,12 +19,22 @@ export declare function addEventStart( args: { id: string; kind: string; name: s
...
  * @param args.details - Arbitrary metadata associated with this phase, possible error info.
  */
 export declare function addEventError( args: { id: string; details: unknown } ): void;
+
+/**
+ * Add an attribute to an event.
+ *
+ * Use the same id as the start phase to correlate phases.
+ *
+ * @param args - Event metadata for the error phase.
```
catsby:
```suggestion
 * @param args - Arguments for the attribute phase.
```

mistaken copy-paste I think?

sdk/core/src/tracing/internal_interface.js
```diff
@@ -32,40 +35,46 @@ export { addEventPhaseWithContext };
...
  * @returns {void}
  */
-export const addEventEnd = options => addEventPhase( 'end', options );
+export const addEventEnd = options => addEventPhase( EventPhase.END, options );
 
 /**
  * Internal use only
  *
- * Adds the error phase at an event as error at the default trace for the current workflow.
+ * Adds the error phase of an event, matching buy its id.
```
catsby:
```suggestion
 * Adds the error phase of an event, matching by its id.
```

sdk/core/src/tracing/internal_interface.js
```diff
@@ -32,40 +35,46 @@ export { addEventPhaseWithContext };
...
  *
- * It needs to use the same id of the start phase.
+ * @param {object} args
+ * @param {string} args.id - A unique id for the Event, must be the same across all phases: start, end, error.
+ * @param {any} args.details - All details attached to this Event Phase. DB queried records, HTTP response body.
+ * @param {string} args.parentId - The parent Event, used to build a three.
```
catsby:
```suggestion
 * @param {string} args.parentId - The parent Event, used to build a tree.
```

sdk/core/src/tracing/internal_interface.js
```diff
@@ -32,40 +35,46 @@ export { addEventPhaseWithContext };
...
+ * Internal use only
+ *
+ * Adds an attribute to an event using its id.
  *
  * @param {object} args
  * @param {string} args.id - A unique id for the Event, must be the same across all phases: start, end, error.
- * @param {string} args.kind - The kind of Event, like HTTP, DiskWrite, DBOp, etc.
- * @param {string} args.name - The human friendly name of the Event: query, request, create.
  * @param {any} args.details - All details attached to this Event Phase. DB queried records, HTTP response body.
  * @param {string} args.parentId - The parent Event, used to build a three.
```
catsby: Is `args.parentId` relevant, used, or valid here? We're not building trees of these event attributes are we?

sdk/core/src/activity_integration/tracing.d.ts
```diff
@@ -21,12 +19,22 @@ export declare function addEventStart( args: { id: string; kind: string; name: s
...
+ * @param args.eventId - The event id
```
catsby: The other events just use `args.id` but here we use `eventId`... I think `eventId` is actually more appropriate for all except maybe `addEventStart()`... thoughts? I'm not sure if I'm asking for any change(s) here yet, I just noticed that and felt we should be consistent, but I'm not yet sure which way to go, all use `id` or all use `eventId` (with maybe `addEventStart()` excluded)

sdk/http/src/cost.ts
```diff
@@ -0,0 +1,22 @@
...
+  total: number,
+  components?: [
+    {
+      name: string,
+      value: number
+    }
+  ]
+};
+
+export const addRequestCost = async ( response: KyResponse | Response, cost: RequestCost ) : Promise<void> => {
```
catsby: If we're exporting this for public use we should probably have JSDocs here right?

sdk/core/src/tracing/trace_consts.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Each possible phase a of a trace event
```
catsby:
```suggestion
 * Each possible phase of a trace event
```

sdk/core/src/tracing/trace_consts.js
catsby, line 1: Does this file need a matching `.d.ts` file ?

sdk/http/src/cost.ts
catsby: also does this need to be async?

catsby (commented):
Mostly great, just left some comments

sdk/core/src/tracing/internal_interface.js
szanata: Good catch

sdk/core/src/activity_integration/tracing.d.ts
szanata: I changed to `eventId` on purpose. It is more clear. I plan to refactor the others, but on different PR.

sdk/core/src/tracing/trace_consts.js
szanata: It is not public.

sdk/http/src/cost.ts
szanata: Fixed

bnchrch (commented):
🥇 Pointer for reviewers: enter through `sdk/core/src/tracing/trace_consts.js` + `build_trace_tree.js` for how attributes land in the trace tree. Then `sdk/http/src/fetch/index.ts:51` + `sdk/http/src/cost.ts` for how the request id symbol gets tagged on a response and read back out by `addRequestCost`.

📌 Heads up for reviewers: the LLM `cost.components` shape changed from `{ input: { value }, output: { value }, ... }` to `[{ name, value }, ...]`. If anything outside these packages reads those keys it will break. WDYT on calling that out in the changeset?

🤔 Left a couple of bigger 'should we do this differently' thoughts on `addRequestCost` and the `ADD_ATTR` phase name. Totally fine to disagree on those. The line-level stuff stands on its own either way.

sdk/core/src/tracing/internal_interface.js
```diff
@@ -10,12 +11,14 @@ export { init, getDestinations };
...
+export { EventPhase };
+
 /**
  * Trace nomenclature
  *
  * Trace - The collection of Events;
- * Event - Any entry in the Trace file, must have the two phases START and END or ERROR;
- * Phase - An specific part of an Event, either START or the conclusive END or ERROR;
+ * Event - Any entry in the Trace file, must have the two phases START and END/ERROR, plus any number of ADD_ATTR phases;
+ * Phase - An specific part of an Event, either START, END, ERROR or ADD_ATTR;
```
bnchrch: 🤔 📚 I keep tripping over `ADD_ATTR` being called a 'phase' here.

Phases in my head are start / end / error, the temporal lifecycle of one event. An attribute isn't really a point in time, it's metadata you hang on the event whenever. So this line:

> Phase - An specific part of an Event, either START, END, ERROR or ADD_ATTR

mixes lifecycle with operation.

Two options:
1. Rename internally so attributes aren't a 'phase' (maybe an `op` / `kind` concept that sits above phase)
2. Just update this doc block to say something like 'phases are START/END/ERROR; attributes are metadata recorded against an event id'

Fine with option 2. This is a nit (💅), wire format is good either way.

sdk/http/src/cost.ts
```diff
@@ -0,0 +1,27 @@
+import { KyResponse } from 'ky';
```
bnchrch, line 1: ❓ 🔮 I wonder if this whole helper wants to be `addRequestAttribute(response, name, value)` instead of `addRequestCost(response, cost)`.

Cost is one thing users will want to tag, but latency, retries, region, cache-hit status, rate-limit remaining, they're all the same shape of problem: 'here's an http response, please tag this metadata onto the trace event behind it.'

```ts
export const addRequestAttribute = ( response: KyResponse | Response, name: string, value: unknown ): void => {
  const eventId = Reflect.get( response, config.requestIdSymbol ) as string;
  if ( !eventId ) {
    console.warn( `addRequestAttribute(): The 'response' argument did not originate from @outputai/http, no '${name}' was added.` );
    return;
  }
  Tracing.addEventAttribute( { eventId, name, value } );
};
```

and then cost is just a convention:

```ts
addRequestAttribute( response, 'cost', { total: 32, components: [...] } );
```

Does that make sense? I know it softens the typed shape for cost specifically, but the `RequestCost` type can still live beside it.

sdk/http/src/cost.ts
```diff
@@ -0,0 +1,27 @@
...
+import { config } from './config.js';
+import { Tracing } from '@outputai/core/sdk_activity_integration';
+
+export type RequestCost = {
```
bnchrch, line 5: ❓ 🔮 The LLM cost comes back in this exact same shape (see `sdk/llm/src/cost/index.js`), but `RequestCost` lives here in `@outputai/http` and the LLM side rebuilds it by convention. I think they're gonna drift.

WDYT about hoisting this type into core (or a tiny shared module):

```ts
export type Cost = {
  total: number;
  components?: Array<{ name: string; value: number }>;
};
```

then both `@outputai/http` and `@outputai/llm` import the same type. That way the trace contract is enforced instead of implied.

sdk/http/src/cost.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @param cost - The cost information
+ * @returns
+ */
+export const addRequestCost = ( response: KyResponse | Response, cost: RequestCost ) : void => {
+  const eventId = Reflect.get( response, config.requestIdSymbol ) as string;
+  if ( !eventId ) {
+    console.warn( 'addRequestCost(): The "response" argument did not originate from @outputai/http, no costs were added.' );
+    return;
+  }
+  Tracing.addEventAttribute( { eventId, name: 'cost', value: cost } );
```
bnchrch: 💅 `'cost'` is also used over in `sdk/llm/src/trace_utils.js:18`. Worth pulling the attribute names into a shared const so they can't silently drift:

```ts
// somewhere in core
export const TraceAttribute = { COST: 'cost' } as const;
```

then both sites use `TraceAttribute.COST`.

sdk/http/src/fetch/index.ts
```diff
@@ -46,6 +47,9 @@ export const fetch = async ( input: RequestInfo | Request, init?: RequestInit )
 
   try {
     const response = await undici.fetch( request );
+
+    Object.defineProperty( response, config.requestIdSymbol, { value: requestId, enumerable: false, configurable: false, writable: false } );
```
bnchrch: 💅 This `Object.defineProperty` is doing a lot on one line, and it's the whole reason `addRequestCost` works. A small named helper reads better and gives us a natural home for a jsdoc explaining *why*:

```ts
// in cost.ts or a small internals file
export const tagResponseWithRequestId = ( response: Response, requestId: string ): void => {
  Object.defineProperty( response, config.requestIdSymbol, {
    value: requestId,
    enumerable: false,
    configurable: false,
    writable: false
  } );
};
```

then here it's just `tagResponseWithRequestId( response, requestId );`.

sdk/core/src/activity_integration/tracing.js
```diff
@@ -10,28 +10,37 @@ import { addEventPhaseWithContext } from '#tracing';
...
+/**
+ * Adds an attribute to an event using the event id
+ *
+ * @param {object} args
+ * @param {string} args.eventId - The id of the event to attach the attribute
+ * @param {string} args.name - The attribute name
+ * @param {unknown} args.value - The attribute value
+ * @returns {void}
+ */
+export const addEventAttribute = ( { eventId, name, value } ) =>
```
bnchrch, line 45: 💅 ❓ Is there a reason this is `eventId` when the siblings (`addEventStart/End/Error`) all take `id`?

```js
export const addEventAttribute = ( { id, name, value } ) =>
  addEventPhaseWithContext( EventPhase.ADD_ATTR, { id, details: { name, value } } );
```

Reading `Tracing.addEventAttribute({ eventId, ... })` right next to `Tracing.addEventEnd({ id, ... })` is a small stutter. WDYT?

sdk/core/src/tracing/internal_interface.js
```diff
@@ -28,44 +31,47 @@ export { addEventPhaseWithContext };
...
+ * Internal use only
  *
- * It needs to use the same id of the start phase.
+ * Adds an attribute to an event using its id.
  *
  * @param {object} args
  * @param {string} args.id - A unique id for the Event, must be the same across all phases: start, end, error.
- * @param {string} args.kind - The kind of Event, like HTTP, DiskWrite, DBOp, etc.
- * @param {string} args.name - The human friendly name of the Event: query, request, create.
  * @param {any} args.details - All details attached to this Event Phase. DB queried records, HTTP response body.
```
bnchrch: 📚 This jsdoc is copy-pasted from `addEventEnd`/`addEventError`. For the attribute path the contract is `details: { name: string, value: unknown }`, not 'DB queried records, HTTP response body.' Worth spelling that out so readers don't go hunting for an http body.

sdk/llm/src/trace_utils.js
```diff
@@ -15,7 +15,8 @@ export const endTraceWithSuccess = async ( traceId, modelId, response, extraDeta
   const { text: result, totalUsage: usage, providerMetadata } = response;
   const cost = await calculateLLMCallCost( { usage, modelId } );
   emitEvent( 'llm:call_cost', { modelId, cost, usage } );
-  Tracing.addEventEnd( { id: traceId, details: { result, usage, cost, providerMetadata, ...extraDetails } } );
+  Tracing.addEventAttribute( { eventId: traceId, name: 'cost', value: cost } );
```
bnchrch: 💅 Same magic string as `sdk/http/src/cost.ts:26`. If we pull attribute names into a shared const (see my comment over there), this becomes `name: TraceAttribute.COST`.

sdk/core/src/tracing/internal_interface.js
szanata: I renamed "phase" to "action, which makes more sense.

sdk/http/src/cost.ts
szanata: It is not a bad idea, but I would prefer to keep the feature to exactly what was being asked for. If there is a need for more information in the future, we can easily revisit it.

sdk/core/src/activity_integration/tracing.js
szanata: Yes, because in `Tracing.addEventEnd({ id, ... })` it is not clear if the id is the attribute id of the id of the event. I plan to refactor other methods down the road as they are less ambiguous.

sdk/http/src/fetch/index.ts
szanata: I totally agree, I just isolated it and added tests and comments.

sdk/core/src/tracing/internal_interface.js
szanata: I have fixed all the comments!

sdk/http/src/cost.ts
szanata: I don't disagree. But I prefer to way for the tracing refactor to introduce exportable types.

sdk/http/src/cost.ts
szanata: Done: both `.js` and `.ts` consts.

sdk/llm/src/trace_utils.js
szanata: Done.

sdk/http/src/cost.ts
szanata: I have added the docs.

bnchrch (approved):

catsby (approved):

PR #173, opened 2026-04-28
feat: Emitting HTTP cost events / Adding costs to LLM response [OUT-418] [OUT-415]

szanata:
## Summary

- **HTTP:** `addRequestCost()` now emits `http:request_cost` with `{ requestId, url, cost }`, aligned with cost events from **LLM**;
- **LLM (generate):** text results wrapper now exposes `.cost`, from `calculateLLMCallCost()`;
- **LLM (stream):** the **`onFinish`** callback receives the SDK finish object wrapped in a Proxy so `.cost` is also available there (the returned `StreamTextResult` is unchanged);
- **Refactor:** replace `trace_utils` / `response_utils` with `utils/trace.js`, `utils/response_wrappers.js`, and move `message` / `source_extraction` under `utils/`; add focused unit tests.
- **Types:** update `index.d.ts` for `LLMCallCost`, augmented `GenerateTextResult`, and wrapped stream `onFinish` types.

## Test plan

- [x] Manual test;
- [x] Unit tests;
- [x] LLM review;
- [x] e2e tests;


sdk/llm/src/utils/trace.js
```diff
@@ -0,0 +1,18 @@
...
+};
+
+export const endTraceWithError = ( { traceId, error } ) => {
+  Tracing.addEventError( { id: traceId, details: error } );
+};
+
+export const endTraceWithSuccess = ( { traceId, modelId, response, cost, ...extra } ) => {
+  const { totalUsage: usage, text: result, providerMetadata } = response;
+  Tracing.addEventAttribute( { eventId: traceId, name: Tracing.Attribute.COST, value: cost } );
+  Tracing.addEventEnd( { id: traceId, details: { result, cost, usage, providerMetadata, ...extra } } );
```
bnchrch: ❗ ❓ Heads up, cost is now landing in three places from this one function:

1. `addEventAttribute({ name: COST, value: cost })` (line 15)
2. inside the END action `details: { ..., cost, ... }` (this line)
3. `emitEvent('llm:call_cost', { cost, ... })` (line 17)

On #169 we deliberately moved cost OUT of `details` so the attribute was the single source of truth for the trace tree. Putting it back here re-introduces the duplication.

Is that intentional (e.g. some downstream consumer needs to see cost in `details` rather than via attribute), or did it slip back in during the refactor?

If intentional: WDYT about a one-line comment here pointing at the rationale, so the next person doesn't try to remove it again.
If not: drop `cost` from the spread.

```suggestion
  Tracing.addEventEnd( { id: traceId, details: { result, usage, providerMetadata, ...extra } } );
```

sdk/llm/src/utils/response_wrappers.js
```diff
@@ -0,0 +1,74 @@
...
+ * onError: Concludes the trace event
+ *
+ * @param {object} args
+ * @param {string} args.traceId - id created by the startTrace
+ * @param {string} args.modelId - id of the model used
+ * @param {Function} args.onFinish - User added callback
+ * @param {Function} args.onError - User added callback
+ * @returns {object} Proxied response
+ */
+export const wrapStreamResponse = ( { traceId, modelId, onFinish: userOnFinish, onError: userOnError } ) => ( {
```
bnchrch: 💅 ❓ Parallel name, non-parallel shape. `wrapTextResponse(response)` returns a wrapped response. `wrapStreamResponse(...)` returns `{ onFinish, onError }` callbacks to splat into the AI SDK call. They don't really do the same thing.

Fun fact: this used to be called `traceStreamCallbacks` before #169, which I think reads more honestly. Maybe pair them as `wrapTextResponse` + `traceStreamCallbacks`, or `wrapTextResponse` + `wrapStreamCallbacks` if you want to keep `wrap*`?

Not blocking, just a small stutter every time I read both names next to each other.

sdk/llm/src/utils/source_extraction.js
```diff
@@ -47,3 +47,17 @@ export function extractSourcesFromSteps( steps ) {
     return [];
   }
 }
+
+/**
+ * Merge sources used tools and sources from AI SDK response into a single list
+ *
+ * Deduplicate by url (first in stays).
```
bnchrch: ❗ 📚 The doc and the behavior disagree.

This line says **"Deduplicate by url (first in stays)"**, but the implementation puts `sourcesFromTools` first and `sourcesFromResponse` last in the concat:

```js
sourcesFromTools.concat( ...sourcesFromResponse ).map( s => [ s.url, s ] )
```

and `new Map(entries)` makes *later* entries win, so response sources actually clobber tool sources. The test in `response_wrappers.spec.js` confirms this: when both have the same url, `wrapped.sources[0].title` is `'from-response'`.

Either flip the doc to match ("response sources win") or flip the concat order so tool sources stay. The behavior in tests reads correct to me, so I'd just fix the doc:

```suggestion
 * Deduplicate by url (response sources win when urls overlap).
```

sdk/llm/src/utils/response_wrappers.js
```diff
@@ -0,0 +1,74 @@
...
+  async onFinish( response ) {
+    const cost = await calculateLLMCallCost( { modelId, usage: response.totalUsage } );
+
+    endTraceWithSuccess( { traceId, modelId, response, cost } );
```
bnchrch, line 56: ❓ The text path passes `sourcesFromTools` through to `endTraceWithSuccess` (line 22), but the stream path doesn't. So streams never get tool-source extraction in their trace details, while `generateText` does.

Is that intentional (streams don't get steps in the same shape?), or a parity gap? If intentional, a one-liner comment here would save the next reader the same question.

sdk/llm/src/utils/response_wrappers.js
```diff
@@ -0,0 +1,74 @@
...
+
+    userOnFinish?.( new Proxy( response, {
+      get( target, prop, receiver ) {
+        if ( prop === 'cost' ) {
```
bnchrch, line 63: 💅 The `if ( prop === 'cost' ) return cost; ... return Reflect.get(...)` pattern is duplicated in both wrappers (line 29 and here). Worth extracting:

```js
const withProxiedCost = ( target, cost ) => new Proxy( target, {
  get( t, prop, receiver ) {
    if ( prop === 'cost' ) return cost;
    return Reflect.get( t, prop, receiver );
  }
} );
```

Bonus: while you're in here, would it make sense for the stream proxy to also expose `.result` for symmetry with the text proxy? Stream finish events have `text`, so `.result → text` would be consistent.

sdk/http/src/cost.ts
```diff
@@ -24,4 +24,5 @@ export const addRequestCost = ( response: KyResponse | Response, cost: RequestCo
     return;
   }
   Tracing.addEventAttribute( { eventId, name: Tracing.Attribute.COST, value: cost } );
+  emitEvent( 'http:request_cost', { requestId: eventId, url: response.url, cost } );
```
bnchrch: 💅 Same value, two names. Locally it's `eventId`, in the emitted event it gets renamed to `requestId`. They are the same thing. The mid-line rename reads as if they're different concepts.

Pick one and stay with it:

```suggestion
  emitEvent( 'http:request_cost', { eventId, url: response.url, cost } );
```

Or rename the local up top to `requestId`, since that's where it originated (in `fetch/index.ts`).

bnchrch: 💅 🔮 Both `'http:request_cost'` here and `'llm:call_cost'` over in `sdk/llm/src/utils/trace.js:17` are magic strings, and the two names don't share a structure.

WDYT about introducing a simple hierarchy now, before more event names land:

```
<concern>:<category>:<subcategory>

cost:llm:call
cost:http:request
```

That way consumers can route or filter by `concern` (cost, latency, retries, …) without parsing each event name. Could land as constants in core:

```ts
export const Events = {
  COST_LLM_CALL: 'cost:llm:call',
  COST_HTTP_REQUEST: 'cost:http:request'
} as const;
```

or just a documented convention. Cheap moment to set the pattern before more event names land.

sdk/llm/src/index.d.ts
```diff
@@ -205,6 +221,29 @@ export type ExtractedSource = {
   title: string;
 };
 
+/**
+ * Cost breakdown from the cost module (`calculateLLMCallCost`). `total` is null when pricing data is missing or calculation fails.
+ */
+export type LLMCallCost = {
+  total: number | null;
```
bnchrch, line 228: ❓ 📚 `total: number | null` is honest, but it's a UX footgun: every caller that does `cost.total + something` has to remember to null-check.

A couple of options:

1. Keep the null and make the JSDoc on `GenerateTextResult.cost` (line 259-ish) much louder about the null case + show a safe-access example
2. Treat `null` as `0` here and document "0 means we couldn't compute it" (less honest, more ergonomic)

Leaning option 1 since you've got the `message?` field for the diagnostic. But worth deciding now before consumers start writing `.total + .total` math everywhere.

sdk/llm/src/agent.js
```diff
@@ -52,48 +52,48 @@ export class Agent extends AIToolLoopAgent {
...
+    this.#prompt = prompt;
+    this.#modelId = loadedPrompt.config.model;
+    this.#initialMessages = allMessages.filter( isRole( ROLE.USER ) );
+    this.#store = conversationStore ?? null;
   }
 
-  async _preSendHook( userMessages ) {
-    const priorMessages = this._store ? await this._store.getMessages() : [];
-    return [ ...this._initialMessages, ...priorMessages, ...userMessages ];
+  async #fetchMessages( userMessages ) {
```
bnchrch, line 61: 💎 Real private fields and intent-revealing names. `#fetchMessages` and `#storeMessages` say what they do. The old `_preSendHook` / `_postSendHook` hid the intent behind a lifecycle vocabulary that didn't really apply. Like to see this.

sdk/llm/src/utils/response_wrappers.js
szanata: This is as it was before my changes, and to be honest, I think it is right.

sdk/http/src/cost.ts
szanata: This variable has two meanings:
- `requestId` its origin
- `traceId` its usage on the trace

But when emitting the event `requestId` is the facet that has meaning for the user. `traceId` is internal only.

sdk/llm/src/utils/response_wrappers.js
szanata: I agree on adding `.result` shortcut as a bonus, since it was lacking before. I prefer to not make a proxy wrapping function since there are still discrepancies between the functions.

sdk/llm/src/utils/trace.js
szanata: Yes, makes sense. I just removed it.

sdk/llm/src/utils/response_wrappers.js
szanata: Renamed to `wrapStreamOnFinishResponse` and removed the `onError` handling from it, making it single purpose.

sdk/llm/src/utils/source_extraction.js
szanata: Fixed, it was only the comment that was wrong.

sdk/http/src/cost.ts
szanata: Updated to the suggested names, but without const.

sdk/llm/src/index.d.ts
szanata: I prefer to keep `null` because `null` is different than `0`:
- `null` something went wrong
- `0` might have been free?

I think the current JSDoc is clear enough, it already says that total might be null and this is also reinforced by the TS typing below it.

docs/guides/costs/cost-events.mdx
```diff
@@ -1,26 +1,29 @@
...
+
+Use either or both to log spend to your observability stack, trigger alerts, or aggregate cost per workflow over time.
 
 ## Setup
 
 Cost events use the same [hooks system](/operations/error-hooks) as error hooks:
 
 1. Create a hook file and import `on` from `@outputai/core/hooks`.
-2. Register a handler for `llm:call_cost`.
+2. Register a handler for `cost:llm:request`, `cost:http:request`, or both.
```
catsby, line 18: `llm:call_cost` to `cost:llm:request` feels like a breaking change, but the changeset lists it as `patch`. I feel like we should upgrade that to major (?) and we should document a migration path

szanata: Makes sense. I ignore it assuming people would not be using this, but I am probably wrong

szanata: Done. It is now v0.3.0, which is a breaking change since when using 0.x, any minor is considered a "major"

bnchrch (approved):
LGTM! Thanks for the changes!

Obligatory gif

Image

catsby (approved):

PR #174, opened 2026-04-28
feat(cli): rebuild output dev as a full-featured INK TUI

bnchrch:
## Overview
image

- Replaces the thin status-only INK panel in `output dev` with a full tabbed terminal UI: **Workflows · Recent Runs · Services · Help**.
- Adds an in-TUI scenario picker, custom-input editor, scrollable JSON expand modal, and live `docker compose logs` tail with restart hotkeys.
- Bumps `@outputai/cli` to `0.3.0-dev.tui.0` and pins the bundled `docker-compose-dev.yml` API image default to the matching dev tag (the TUI relies on the new `runId` field on `/workflow/runs`).

## Demo

📹 [Loom walkthrough](https://www.loom.com/share/3b1810f8cec148d68429c8cfd01fc9d7)

## Problem

`output dev` previously printed a static service-health table and a flat list of recent runs. Anything beyond glancing at it required a second terminal — running workflows, watching specific runs, looking at output, tailing logs, restarting services. The shutdown banner ("Stopping services…") was the only interactive surface.

## Solution

A single INK app rendered in the alternate screen buffer with a fixed-height frame, four tabs, and a small set of modals layered on top.

**Tabs**
- **Workflows** — catalog list with descriptions; below the rule a magenta heading shows the selected workflow plus its run stats and recent runs.
- **Recent Runs** — runs sorted with active/failed first; below the rule a magenta `<workflow_type> : <run_id>` heading and the run's `Input` / `Output` JSON pane (`←/→` toggles). `enter` drills into a step view.
- **Services** — Docker status list above a live `docker compose logs -f` tail with `r/R` restart, `p` pause, `c` clear, `o` open URL.
- **Help** — paginated sections, default `Run from CLI`.

**Cross-cutting**
- Two-tone OUTPUT logo + counters (running / failed / services-down / workflow & run totals).
- Live JSON syntax highlighting via a hand-rolled tokenizer + colored `<Text>`.
- `e` opens the current input/output in a fullscreen scrollable modal (`↑/↓` line, `pgup/pgdn` page) — replaced the original `$EDITOR` shell-out which was racing with INK's render and corrupting frames.
- Custom workflow input is edited inside the TUI with live JSON validity feedback.
- Arrow-only navigation across the app; no `j/k/h/l`.
- Slight padding around the app so the logo isn't flush with the corner.

**API/infra**
- The runs list endpoint already returns `runId` per row (existing API code), but the published `outputai/api:0.2.0` image predates that. The bundled compose now defaults to `outputai/api:0.3.0-dev.tui.0`. Use `output dev --image-pull-policy never` after building the image locally.

## Notes for reviewers

- The frame is wrapped in `<Box height={stdout.rows} overflow="hidden">` plus alt-screen entry/exit. Together they keep `log-update` from scrolling old frames into terminal scrollback when the rendered tree exceeds visible rows.
- Old shell-out helpers (`view_in_pager.ts`, `edit_in_editor.ts`) were tried and removed — they all hit the same INK + altscreen race once polling fired during the spawned process. The fullscreen modal + in-TUI editor sidesteps the race entirely.
- `views/dev.tsx` and `views/workflow/list.tsx` are deleted; everything new lives under `sdk/cli/src/views/dev/`.

## Test plan

- [ ] `pnpm run build` from `sdk/cli/`
- [ ] `docker build . -f ops/api.Dockerfile -t outputai/api:0.3.0-dev.tui.0`
- [ ] `output dev --image-pull-policy never` from a project with workflows
- [ ] Tab through Workflows / Recent Runs / Services / Help; each renders cleanly
- [ ] Select a workflow → press `r` → run a scenario; new run appears in Recent Runs within ~2s
- [ ] Custom input via the run modal → live validity indicator → submits valid JSON
- [ ] Recent Runs → highlight a run → `←/→` toggles Input / Output (colored JSON); `e` opens fullscreen modal; `esc` returns cleanly
- [ ] `enter` on a run → step list renders, `↑/↓` navigates, `←/→` cycles Input · Output · Meta
- [ ] Services tab → service rows + live log tail; `r` restarts one service, status flips through `starting → healthy`
- [ ] `Ctrl+C` exits altscreen, runs `docker compose down`, returns to a clean shell


sdk/cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@outputai/cli",
-  "version": "0.2.0",
+  "version": "0.3.0-dev.tui.0",
```
szanata: This should not be changed by hand, it might break our release flow.

szanata (commented):
Could you upgrade ink to v7?

sdk/cli/package.json
bnchrch: Oh sorry this was an artifact of me confirming the correct output was installed. reverting

bnchrch:
@szanata updated to v7

sdk/cli/src/views/dev/hooks/use_poll.ts
```diff
@@ -0,0 +1,135 @@
...
+): void => {
+  const onRunsRef = useRef( onRuns );
+  onRunsRef.current = onRuns;
+
+  usePoll( enabled, POLL_INTERVAL_MS, async () => {
+    try {
+      const { runs } = await fetchWorkflowRuns( { limit: 100 } );
+      onRunsRef.current( runs );
+    } catch {
+      // API may not be ready yet
```
catsby, line 131: What happens if the API is crashed?

sdk/cli/src/views/dev/hooks/use_run_detail.ts
```diff
@@ -0,0 +1,215 @@
...
+      await getWorkflowIdRunsRidResult( workflowId, runId ) :
+      await getWorkflowIdResult( workflowId );
+    if ( response.status !== 200 ) {
+      return null;
+    }
+    return response.data as WorkflowResultResponse;
+  } catch {
+    return null;
+  }
+};
```
catsby, line 132: `fetchResult()` and `fetchTrace()` both just do nothing if they don't get a `200` response or otherwise fail? Should we do some kind of toast style message with some kind of warning?

sdk/cli/src/commands/dev/index.ts
```diff
@@ -90,6 +82,43 @@ export default class Dev extends Command {
...
+
+    process.on( 'exit', exitAltScreenOnce );
+
+    const handleSignal = async () => {
+      await cleanup();
+      instanceRef.current?.unmount();
+    };
+
+    process.on( 'SIGINT', handleSignal );
+    process.on( 'SIGTERM', handleSignal );
```
catsby, line 142: `handleSignal` is async and should be awaited? Or make it synchronous, does it need to be async?

sdk/cli/src/views/dev/hooks/use_poll.ts
bnchrch: good question — silent retry is intentional here. at a 2s tick, toasting on every flake would just spam. when the api is actually down it surfaces in the Services tab via the docker health poll, which is the right place to see it. open to adding a one-shot warning toast if N ticks fail in a row, but leaving as-is for now.

sdk/cli/src/views/dev/hooks/use_run_detail.ts
bnchrch: yeah this is a real gap — fixed in d6929741. while a run is going the api legitimately 404s on `/result` and `/trace-log`, so we cant toast on every miss. now those two helpers swallow only 404 and re-throw anything else. the existing `.catch` in `useRunDetail` puts the message into `detail.error` and also pushes a one-shot error toast (deduped per run+message so a flapping api at the 2s poll cadence doesnt stack toasts).

sdk/cli/src/commands/dev/index.ts
bnchrch: right — `process.on` doesnt await the listener, so the previous `async () => {}` floated the cleanup promise and a rejection would surface as an unhandled rejection. fixed in d6929741: handler is sync, calls cleanup with explicit `.catch` (logs the failure) and `.finally` (always unmounts ink).

sdk/cli/src/views/dev/hooks/use_run_detail.ts
bnchrch: walking that back in 985d4f37 — the toast wasnt actually useful. result/trace fetches are best-effort and many statuses (in-progress, failed, canceled) legitimately wont have a full payload at any given moment. silent swallow is the right behavior, no error state needed.

sdk/cli/src/commands/dev/index.ts
```diff
@@ -90,6 +82,48 @@ export default class Dev extends Command {
...
+
+    process.on( 'exit', exitAltScreenOnce );
+
+    // `process.on` doesn't await the handler, so the cleanup promise would
+    // float and any rejection would surface as an unhandled rejection.
+    // Wrap the async work in a sync registration that explicitly logs
+    // failures and always unmounts Ink afterwards.
+    const handleSignal = (): void => {
+      cleanup()
+        .catch( err => console.error( 'Cleanup failed:', getErrorMessage( err ) ) )
```
catsby: I think at this point INK still owns all the things/screens, so this gets swallowed. 

Maybe call `exitAltScreenOnce()` first?

```js
const handleSignal = (): void => {
  cleanup()
    .catch( err => {
      exitAltScreenOnce();
      console.error( 'Cleanup failed:', getErrorMessage( err ) );
    } )
    .finally( () => instanceRef.current?.unmount() );
};
```

catsby (approved):
I'll give ✅ however, we need to resolve the API dependency first. I think we need a new release of Output with the new history/run ID endpoints, so we can update this PR to reference it, then we can merge and release. Happy to be wrong though

bnchrch:
@catsby I dont think we need to resolve the api dependency first?

So all this PR doesnt change the api. The API changes are all in main, just unpublished.

Which means theres a few ways to test

## 1. Run `npm run dev`

This builds a fresh api image and runs the cli. Everything should work

## 2. Build the api image, then run `output dev --image-pull-policy never`
If you
1. build the api image
2. and build and link the cli from source

You can run `output dev --image-pull-policy never` in a new `output init` project no problem and have it all work

## 3. Run the release dev workflow from this branch
You can run dev release here and use the dev version of the cli, and api to init a new project with the api and cli changes

sdk/cli/src/commands/dev/index.ts
bnchrch: good catch — applied in 719affc9. exiting the alt-screen first means the error actually lands on the users real terminal instead of being painted into a buffer ink still owns. `exitAltScreenOnce` is idempotent so the on-exit handler skipping after this call is fine.

szanata:
All docker images are online and running but some icons are not green:
image

sdk/cli/src/views/dev/utils/json_editor.tsx
```diff
@@ -0,0 +1,171 @@
...
+        <Box marginTop={1}>
+          <Text color="yellow">{submitMessage}</Text>
+        </Box>
+      )}
+
+      <Box marginTop={1}>
+        <Text bold>ctrl+s</Text><Text dimColor> submit  </Text>
+        <Text bold>esc</Text><Text dimColor> cancel  </Text>
+        <Text bold>↑↓←→</Text><Text dimColor> move  </Text>
+        <Text bold>tab</Text><Text dimColor> indent</Text>
```
szanata: Are those spaces inside Text tag necessary? If they are made on purpose, would be bette rto use `/nbsp`?

test_workflows/package.json
```diff
@@ -9,7 +9,7 @@
     "output:worker:build": "rm -rf dist/* && tsc && output-copy-assets",
     "output:worker:start": "output-worker",
     "output:worker": "npm run output:worker:install && npm run output:worker:build && npm run output:worker:start",
-    "output:worker:watch": "npx nodemon --watch src --watch package.json --ext ts,js,json,prompt,md --ignore 'dist/**' --ignore '**/*.test.*' --exec 'npm run output:worker'"
+    "output:worker:watch": "npm run output:worker:install && npx nodemon --watch src --watch package.json --ext ts,js,json,prompt,md --ignore 'dist/**' --ignore '**/*.test.*' --exec 'npm run output:worker'"
```
szanata: Why the install was added here?

szanata (commented):
I left some comments

test_workflows/package.json
bnchrch: leftover from when the worker container shipped empty and `nodemon` wasnt resolvable until install ran. that build now ships with deps installed, so the prefix isnt needed. removed in 7d2e855d.

bnchrch:
fixed in 7d2e855d — `isServiceHealthy` already treats `running` + `health=none` as healthy. now the icon matches that semantics, so services without a healthcheck (e.g. redis) render green when up instead of the neutral blue dot.

sdk/cli/src/views/dev/utils/json_editor.tsx
bnchrch: No theyre not! Removed

PR #175, opened 2026-04-29
fix(cli,credentials): clean error messages for credentials.key failures [OUT-431]

marcosmarxm:
## Summary

Running any `output` CLI command with `credential:` env refs used to dump a raw stack trace when credentials couldn't be loaded. This PR turns both failure modes from [OUT-431](https://linear.app/growthx/issue/OUT-431) into a single clean error line + exit 1:

1. **Missing `credentials.key`** → caught at the bin layer; prints the existing `MissingKeyError` message without a stack trace.
2. **Wrong `credentials.key`** → AES-GCM auth tag failure inside `@noble/ciphers` is now wrapped into a new `InvalidCredentialsKeyError` that names the credentials file, then handled at the bin layer the same way.

Unrelated errors still propagate.

### Before (wrong key)
```
file:///.../@noble/ciphers/aes.js:820
    throw new Error('aes/gcm: invalid ghash tag');
Error: aes/gcm: invalid ghash tag
    at Object.decrypt (...)
    [...8 more frames...]
```

### After (wrong key)
```
Error: Failed to decrypt /path/to/config/credentials.yml.enc. The credentials key does not match the one used to encrypt this file. Check OUTPUT_CREDENTIALS_KEY env var or config/credentials.key.
```

### Before (missing key)
```
file:///.../encrypted_yaml_provider.js:19
    throw new MissingKeyError(environment);
MissingKeyError: No credentials key found. Set OUTPUT_CREDENTIALS_KEY env var or create config/credentials.key.
    at resolveKey (...)
    [...8 more frames...]
```

### After (missing key)
```
Error: No credentials key found. Set OUTPUT_CREDENTIALS_KEY env var or create config/credentials.key.
```

## Changes

- `sdk/credentials/src/errors.ts` — adds `InvalidCredentialsKeyError`; sets `error.name` on credential error classes
- `sdk/credentials/src/encrypted_yaml_provider.ts` — wraps `decryptYaml` to translate the cipher's GCM tag error into `InvalidCredentialsKeyError`
- `sdk/credentials/src/index.ts` — exports `InvalidCredentialsKeyError`
- `sdk/cli/src/utils/credentials_loader.ts` *(new)* — catches `MissingKeyError`/`InvalidCredentialsKeyError` from `resolveCredentialRefs()`, prints the message, exits 1; rethrows everything else
- `sdk/cli/bin/run.js` — uses the new wrapper

## Test plan

- [x] New unit tests in `credentials_loader.spec.ts` (5 tests) cover happy path, both error types, env-specific message variants, and rethrow-on-unexpected-error
- [x] New test in `encrypted_yaml_provider.spec.ts` verifies a wrong key throws `InvalidCredentialsKeyError` with the credentials path
- [x] Existing credentials package tests still pass
- [x] Lint passes
- [x] Manual: run `npx output --version` in a project with `credential:` env refs and (a) no `credentials.key`, (b) wrong `credentials.key` — confirm clean message + exit 1

### DIFF
--- a/sdk/cli/bin/run.js
+++ b/sdk/cli/bin/run.js
@@ -3,11 +3,11 @@
 import { execute } from '@oclif/core';
 import { loadEnvironment } from '../dist/utils/env_loader.js';
 import { bootstrapProxy } from '../dist/utils/proxy.js';
-import { resolveCredentialRefs } from '@outputai/credentials';
+import { loadCredentialRefs } from '../dist/utils/credentials_loader.js';
 
 // Load environment variables from .env files before executing CLI
 loadEnvironment();
 bootstrapProxy();
-resolveCredentialRefs();
+loadCredentialRefs();
 
 await execute( { dir: import.meta.url } );
--- a/sdk/credentials/src/encrypted_yaml_provider.ts
+++ b/sdk/credentials/src/encrypted_yaml_provider.ts
@@ -1,7 +1,7 @@
 import { readFileSync, existsSync } from 'node:fs';
 import { load as parseYaml } from 'js-yaml';
 import { decrypt } from './encryption.js';
-import { InvalidCredentialsKeyError, MissingKeyError } from './errors.js';
+import { InvalidCredentialsKeyError, MalformedCredentialsKeyError, MissingKeyError } from './errors.js';
 import {
   resolveKeyEnvVar,
   resolveCredentialsPath,
@@ -52,8 +52,17 @@ const resolveWorkflowKey = ( workflowName: string, workflowDir: string ): string
 const safeDecrypt = ( ciphertext: string, key: string, credPath: string ): string => {
   try {
     return decrypt( ciphertext, key );
-  } catch {
-    throw new InvalidCredentialsKeyError( credPath );
+  } catch ( error ) {
+    if ( error instanceof RangeError ) {
+      throw new MalformedCredentialsKeyError( credPath, error.message );
+    }
+    if ( error instanceof Error && error.message.includes( 'aes/gcm' ) ) {
+      throw new InvalidCredentialsKeyError( credPath );
+    }
+    throw new InvalidCredentialsKeyError(
+      credPath,
+      error instanceof Error ? error.message : String( error )
+    );
   }
 };
 
--- a/sdk/credentials/src/encrypted_yaml_provider.ts
+++ b/sdk/credentials/src/encrypted_yaml_provider.ts
@@ -1,7 +1,7 @@
 import { readFileSync, existsSync } from 'node:fs';
 import { load as parseYaml } from 'js-yaml';
 import { decrypt } from './encryption.js';
-import { MissingKeyError } from './errors.js';
+import { InvalidCredentialsKeyError, MissingKeyError } from './errors.js';
 import {
   resolveKeyEnvVar,
   resolveCredentialsPath,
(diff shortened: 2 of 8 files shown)


sdk/credentials/src/encrypted_yaml_provider.ts
```diff
@@ -49,8 +49,16 @@ const resolveWorkflowKey = ( workflowName: string, workflowDir: string ): string
...
-  ( parseYaml( decrypt( readFileSync( credPath, 'utf8' ).trim(), key ) ) as Record<string, unknown> ) || {};
+const decryptYaml = ( credPath: string, key: string ): Record<string, unknown> => {
+  try {
+    return ( parseYaml( decrypt( readFileSync( credPath, 'utf8' ).trim(), key ) ) as Record<string, unknown> ) || {};
+  } catch ( error: unknown ) {
+    if ( error instanceof Error && error.message.includes( 'aes/gcm' ) ) {
+      throw new InvalidCredentialsKeyError( credPath );
+    }
+    throw error;
+  }
```
catsby, line 66: I think we should break these up in to different checks, to distinguish between failure to decrypt and failure to parse. For example, while this catches failure to decrypt with a bad key (which gives the `aes/gcm` error), if you truncate a key so it's not the right length, or malform the key (replace any char with an empty space ` `), you get different errors that don't get caught here:

Invalid key length:

```
❯ output credentials show
file:///[...]node_modules/.pnpm/@noble+ciphers@2.2.0/node_modules/@noble/ciphers/utils.js:332
        throw new RangeError('hex string expected, got unpadded hex of length ' + hl);
              ^

RangeError: hex string expected, got unpadded hex of length 55
```

Key with ` ` in the middle:

```
~/go/github.com/growthxai/output-examples main*
❯ output credentials show
file:///[...]node_modules/.pnpm/@noble+ciphers@2.2.0/node_modules/@noble/ciphers/utils.js:339
            throw new RangeError('hex string expected, got non-hex character "' + char + '" at index ' + hi);
                  ^

RangeError: hex string expected, got non-hex character " 2" at index 10
```

sdk/cli/bin/run.js
```diff
@@ -3,11 +3,11 @@
...
 import { loadEnvironment } from '../dist/utils/env_loader.js';
 import { bootstrapProxy } from '../dist/utils/proxy.js';
-import { resolveCredentialRefs } from '@outputai/credentials';
+import { loadCredentialRefs } from '../dist/utils/credentials_loader.js';
 
 // Load environment variables from .env files before executing CLI
 loadEnvironment();
 bootstrapProxy();
-resolveCredentialRefs();
+loadCredentialRefs();
```
catsby, line 11: we may want to move this to somewhere in [cli/hooks/init.ts](https://github.com/growthxai/output/blob/7bd2d4ed840c1fedce37a50b2e4846f701fe8ef5/sdk/cli/src/hooks/init.ts#L19), and skip it if the user is running `--version` or `--help`. Right now, users can't even run `--version` without an error if their key is wrong

catsby (commented):
Looks great so far! I have just a few requests

sdk/credentials/src/encrypted_yaml_provider.ts
marcosmarxm: Good catch — fixed in 47cf0e0. Narrowed the try-block to wrap only `decrypt()` so any throw from it (GCM tag failure, truncated key, non-hex chars) becomes `InvalidCredentialsKeyError`. `parseYaml` errors stay outside the try and propagate. Added unit tests for both failure modes you identified.

marcosmarxm: The try block now wraps only decrypt() so YAML parse errors propagate as themselves. `RangeError` from `hexToBytes` (truncated or non-hex key) maps to a new `MalformedCredentialsKeyError` that points at format issues like whitespace and truncation; AES-GCM auth tag failure keeps the precise `InvalidCredentialsKeyError` "does not match" message; any other unknown decrypt failure produces `InvalidCredentialsKeyError` with the underlying error text and softer "may not match, or file may be corrupted" wording so we never lie about the cause and never dump a stack. Unit tests cover all three paths plus the truncated and non-hex variants you flagged.

sdk/cli/bin/run.js
marcosmarxm: So a few ways we could go:
1. just skip for `--version` and `--help` like you suggested. Tiny change. Catch is output credentials init still blows up when the key's broken, so the user can't  fix things from the CLI.
2. same flags, plus skip for init, version, update and some credentials actions so a busted key doesn't lock you out of the commands you'd use to fix it. Few more lines in the  hook.
3. flip it to an allow-list and only load creds for `workflow` and `credentials` actions. Cleanest in theory, but every new command author has to remember to opt in.

catsby (approved):

PR #177, opened 2026-04-29
feat(core): Adding new workflow lifecycle hooks [OUT-433]

szanata:
## Summary

- Added new hook functions `onWorkflowStart`, `onWorkflowEnd`, `onWorkflowError`:
  - `onWorkflowStart()`: Triggers when a workflow starts, receives the run id and workflow name;
  - `onWorkflowEnd()`: Triggers when a workflow ends (no error), receives the run id, workflow name and duration (elapsed time);
  - `onWorkflowError()`: Triggers when a workflow throws an error, receives the run id, workflow name, duration and error thrown;
  - Important: These three hooks are not triggered by the internal "$catalog" workflow lifecycle;
- Renamed `onBeforeStart()` hook to `onBeforeWorkerStart()`;
- Fixed possible issue where a broken handler attached to `onBeforeStart()` could interrupt the worker process.

## Test plan

[x] e2e
[x] manual test


bnchrch (approved):
Approved!

Image

sdk/core/src/hooks/index.js
```diff
@@ -1,34 +1,50 @@
...
+ *
+ * @param {Function} fn
+ * @param {any} args - Args to invoke the function with
+ * @param {string} hookName - hookName to identify this hook function in the logs
+ */
+const safeInvoke = async ( fn, args, hookName ) => {
+  try {
+    await fn( args );
+  } catch ( error ) {
+    log.error( `${hookName} hook error`, { error } );
```
catsby: I think we need to shape the error here, or the [dev logger](https://github.com/growthxai/output/blob/883624761cbcbdf3de341399366dcfd6e43ccff0/api/src/logger.js#L14) may not format what we expect here. Other usages of `log.error` shape the object more:

- https://github.com/growthxai/output/blob/883624761cbcbdf3de341399366dcfd6e43ccff0/sdk/core/src/worker/index.js#L93
- https://github.com/growthxai/output/blob/883624761cbcbdf3de341399366dcfd6e43ccff0/sdk/core/src/tracing/trace_engine.js#L55

sdk/core/src/hooks/index.js
```diff
@@ -1,34 +1,50 @@
...
+    safeInvoke( handler, { source: 'workflow', workflowName: name, error }, 'onError' ) );
   messageBus.on( BusEventType.RUNTIME_ERROR, async ( { error } ) =>
-    invokeHandler( { source: 'runtime', error } ) );
+    safeInvoke( handler, { source: 'runtime', error }, 'onError' ) );
 };
 
-export const onBeforeStart = handler => messageBus.on( BusEventType.WORKER_BEFORE_START, handler );
+/** Listen to worker before start events */
+export const onBeforeWorkerStart = handler => messageBus.on( BusEventType.WORKER_BEFORE_START, () =>
+  safeInvoke( handler, undefined, 'onBeforeWorkerStart' ) );
```
catsby, line 34: The docs in `index.d.ts` say the `handler` will run before the workflow will, but `safeInvoke()` is `async` so wont this return a promise and move on? Do we need to `await` here?

catsby: Should we add a spec file similar to [log_hooks.spec.js](https://github.com/growthxai/output/blob/883624761cbcbdf3de341399366dcfd6e43ccff0/sdk/core/src/worker/log_hooks.spec.js) to verify these hooks?

sdk/core/src/hooks/index.js
```diff
@@ -1,34 +1,50 @@
...
+export const onError = handler => {
   messageBus.on( BusEventType.ACTIVITY_ERROR, async ( { name, workflowName, error } ) =>
-    invokeHandler( { source: 'activity', activityName: name, workflowName, error } ) );
+    safeInvoke( handler, { source: 'activity', activityName: name, workflowName, error }, 'onError' ) );
   messageBus.on( BusEventType.WORKFLOW_ERROR, async ( { name, error } ) =>
-    invokeHandler( { source: 'workflow', workflowName: name, error } ) );
+    safeInvoke( handler, { source: 'workflow', workflowName: name, error }, 'onError' ) );
   messageBus.on( BusEventType.RUNTIME_ERROR, async ( { error } ) =>
-    invokeHandler( { source: 'runtime', error } ) );
+    safeInvoke( handler, { source: 'runtime', error }, 'onError' ) );
```
catsby, line 29: `onError()` is called with `id` and `duration` as well, should we include those in these `safeInfoke()` calls?

sdk/core/src/hooks/index.js
```diff
@@ -1,34 +1,50 @@
...
+    log.error( `${hookName} hook error`, { error } );
+  }
+};
 
+/** Triggers on any errors: workflow, activity and runtime */
+export const onError = handler => {
   messageBus.on( BusEventType.ACTIVITY_ERROR, async ( { name, workflowName, error } ) =>
-    invokeHandler( { source: 'activity', activityName: name, workflowName, error } ) );
+    safeInvoke( handler, { source: 'activity', activityName: name, workflowName, error }, 'onError' ) );
   messageBus.on( BusEventType.WORKFLOW_ERROR, async ( { name, error } ) =>
```
catsby: should we distinguish between `WORKFLOW_ERROR` and a generic catch-all error type? Here, `onError` and `onWorkflowError` both listen on `BusEventType.WORKFLOW_ERROR`, but the have a different output shape:

- `onError`: `handler, { source: 'workflow', workflowName: name, error }`
- `onWorkflowError`: `handler, { id, name, duration, error }`

If someone wants to wire up Sentry or PagerDuty or something, they should probably only use 1 of these, otherwise things will fire twice, right?

sdk/core/src/hooks/index.js
szanata: For consistency with the other sources, I don't think so.

sdk/core/src/hooks/index.js
szanata: Yes, in that case it would fire twice. But I think it is ok, if they want sentry => `onError`, if they want to check failed workflows `onWorkflowError()`

sdk/core/src/hooks/index.js
szanata: Emit is always async, it is fire and forget. so doesnt matter if we wait the handler of not.

sdk/core/src/hooks/index.js
szanata: Good eye. I will fix this.

sdk/core/src/hooks/index.js
szanata: > When the EventEmitter object emits an event, all of the functions attached to that specific event are called synchronously. Any values returned by the called listeners are ignored and discarded.

https://nodejs.org/dist/latest/docs/api/events.html

szanata: I added tests thou. Thanks

catsby: > is fire and forget

Yes, that reiterates my point. As I mentioned previously, the docs in `index.d.ts` say:

```
/**
 * Register a handler to be invoked once, before the worker starts processing tasks.
 * Runs before Worker.create().
```

Our documentation suggests the hook(s) will be ran before the worker is created, but we can guarantee that here can we? If any of the hook handlers are async they may not finish by the time `Worker.create()` fires, right? 

This might be a simple fix to just update the wording in the docs to match the behavior

sdk/core/src/hooks/index.js
catsby: If users want Sentry to receive errors, they can use `onError()` as a catch-all, but any workflow errors reported that way will only receive `source: 'workflow', workflowName: name, error`. They wont receive the workflow ID, they'll only get the name to go off of. If they want the ID, they need to also use `onWorkflowError()`, which gives them duplication, right?

sdk/core/src/hooks/index.js
```diff
@@ -1,34 +1,50 @@
...
+/**
+ * Invokes an external hook handler function with a try catch around it
+ *
+ * @param {Function} fn
+ * @param {any} args - Args to invoke the function with
+ * @param {string} hookName - hookName to identify this hook function in the logs
+ */
+const safeInvoke = async ( fn, args, hookName ) => {
+  try {
+    await fn( args );
```
catsby, line 16: If a user's hook throws, should we emit a runtime error? Otherwise the only insight to failing/throwing hooks is monitoring logs for `hook error` lines

sdk/core/src/hooks/index.spec.js
```diff
@@ -0,0 +1,150 @@
...
+        workflowName: 'wf',
+        error: err
+      } );
+    } );
+
+    it( 'logs and does not rethrow when handler rejects', async () => {
+      const handler = vi.fn().mockRejectedValue( new Error( 'boom' ) );
+      onError( handler );
+
+      await onHandlers[BusEventType.RUNTIME_ERROR]( { error: new Error( 'rt' ) } );
```
catsby: should we assert that the handler gets called, or what kind of shape we get? It's not clear what this test really verifies

sdk/core/src/hooks/index.js
szanata: I don't think so, as this is not a runtime error on our side. Also, we enter a circular logic with that, handler `throws error -> emit runtime error -> handler. throws error -> ...` I think that the logs suffice.

sdk/core/src/hooks/index.spec.js
szanata: The test verified that no error is thrown. I added an assertion for the handler call.

szanata:
@catsby I added `activityId` and `workflowId` to `onError()` handler payload

sdk/core/src/hooks/index.js
bnchrch: This is a good point. We will want the workflow id + run id for any error we can

szanata: I have added workflowId, activityId.

catsby (approved):

PR #178, opened 2026-04-29
Re-export ai.jsonSchema from outputai/llm

catsby:
## Summary

Other projects are importing `ai.jsonSchema` for things, even though `@outputai/llm` exports `ai`, we don't directly export `jsonSchama`, so this will be helpful so users won't need to add `ai@v6` as an extra dependency when they already have a version we ship with.

### DIFF
--- a/.changeset/twenty-ways-invite.md
+++ b/.changeset/twenty-ways-invite.md
@@ -1,0 +1,5 @@
+---
+"@outputai/llm": patch
+---
+
+re-export ai.jsonSchema for downstream use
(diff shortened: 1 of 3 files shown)


szanata (changes requested):
Our pattern is to not use caret operators on packages. We bind to specific versions for stability and repeatability.

catsby:
@szanata 👍 I removed those changes so now it's only exporting `jsonSchema`

szanata (approved):
Looks good to me!

PR #179, opened 2026-04-30
feat(core): Adding support to use workflows installed via npm [OUT-404]

szanata:
## Summary

Adding support to run workflows imported from installed packages:
```js
import { someWorkflow } from 'some-package';
import { workflow } from '@outputai/core';

export default workflow( {
  name: 'myWorkflow',
  fn: async ( ) => {
    await someWorkflow();
  }
} );

```

When the worker starts, it will look for all packages that have `output.workflows.expose: true` config flag in their `package.json`. For each package, it will load all workflows, activities and shared activities (steps/evaluators), the same way it loads for the local code.

If the same workflow is resolved from multiple dependencies (via symlinks) they are de-duplicated.

Workflow and shared activities can still collided with local ones if they have the same name. I evaluated the idea of creating sub namespaces for them, but I took the stance to first check if this is indeed a problem in the real world, as workflow packages will likely have their own name schema.

In order to create a project that exposes workflows, it just need to:
1. Follow the standard convetion for Output projects: `workflow.js`, `steps.js`, etc;
2. Exports the workflows, this can be a barrel export, or named exports;
3. Add this to the `package.json`:
```json
{
   ...
   "output": {
     "workflows": {
       "expose": true
     }
   }
}
```
4. Add a build step that builds it to `.js` with `.d.ts` files.

#### External workflows
image

#### External activities
image

## Test plan

[x] Manual
[x] Unit
[x] e2e

### DIFF
--- a/test_workflows/package.json
+++ b/test_workflows/package.json
@@ -12,6 +12,7 @@
     "output:worker:watch": "npx nodemon --watch src --watch package.json --ext ts,js,json,prompt,md --ignore 'dist/**' --ignore '**/*.test.*' --exec 'npm run output:worker'"
   },
   "dependencies": {
+    "@growthxlabs/workflows_catalog": "0.0.3",
     "@outputai/core": "workspace:^",
     "@outputai/credentials": "workspace:^",
     "@outputai/evals": "workspace:^",
--- a/test_workflows/package.json
+++ b/test_workflows/package.json
@@ -21,7 +21,7 @@
     "@temporalio/workflow": "catalog:",
     "nodemon": "3.1.14"
   },
-  "output": {
+  "@outputai/config": {
     "hookFiles": [
       "./src/hooks.js"
     ]
--- a/test_workflows/package.json
+++ b/test_workflows/package.json
@@ -21,7 +21,7 @@
     "@temporalio/workflow": "catalog:",
     "nodemon": "3.1.14"
   },
-  "@outputai/config": {
+  "outputai": {
     "hookFiles": [
       "./src/hooks.js"
     ]
--- a/test_workflows/package.json
+++ b/test_workflows/package.json
@@ -12,7 +12,7 @@
     "output:worker:watch": "npx nodemon --watch src --watch package.json --ext ts,js,json,prompt,md --ignore 'dist/**' --ignore '**/*.test.*' --exec 'npm run output:worker'"
   },
   "dependencies": {
-    "@growthxlabs/workflows_catalog": "0.0.3",
+    "@growthxlabs/workflows_catalog": "0.0.12",
     "@outputai/core": "workspace:^",
     "@outputai/credentials": "workspace:^",
     "@outputai/evals": "workspace:^",
--- a/.changeset/young-breads-drive.md
+++ b/.changeset/young-breads-drive.md
@@ -1,0 +1,5 @@
+---
+"@outputai/core": patch
+---
+
+Add support for discovering and running workflows from installed npm packages.
--- a/.changeset/young-breads-drive.md
+++ b/.changeset/young-breads-drive.md
@@ -1,5 +1,7 @@
 ---
-"@outputai/core": patch
+"@outputai/core": minor
 ---
 
 Add support for discovering and running workflows from installed npm packages.
+
+Rename the Output.ai settings property in `package.json` from `output` to `@outputai/config`.
--- a/docs/guides/costs/cost-events.mdx
+++ b/docs/guides/costs/cost-events.mdx
@@ -16,7 +16,7 @@ Cost events use the same [hooks system](/operations/error-hooks) as error hooks:
 
 1. Create a hook file and import `on` from `@outputai/core/hooks`.
 2. Register a handler for `cost:llm:request`, `cost:http:request`, or both.
-3. Add the file path to `@outputai/config.hookFiles` in `package.json`.
+3. Add the file path to `outputai.hookFiles` in `package.json`.
 
 See [Error Hooks — Setup](/operations/error-hooks#setup) for the hook file registration pattern.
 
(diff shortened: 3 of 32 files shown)


bnchrch (commented):
First off the wiring through the webpack pipeline is great, and the tests cover all the right paths.

What nags at me is that this doesnt address how users of the framework bring their own shared workflows. With the whitelist hardcoded in `@outputai/core`, someone building `@myorg/some-workflows` cant use the framework without us cutting a release.

The simplest fix would be to lift the whitelist out of the SDK and into the consumer's `package.json`, matching the `package.json.output.hookFiles` pattern we already have:
```json
{
  "output": {
    "workflowPackages": ["@growthxlabs/workflows_catalog", "@myorg/some-workflows"]
  }
}
```

That works, but it still puts the work on every consumer.

What if we extended the existing `output-workflow-bundle` export condition in `bundler_options.js` instead? Packages would self-declare:
```json
{
  "exports": {
    ".": {
      "output-workflow-bundle": "./src/index.js",
      "import": "./src/index.js"
    }
  }
}
```

Then:
- the `exclude` predicate in `bundler_options.js` checks the owning package's `package.json` for that condition (cached)
- the worker loader walks `node_modules` looking for that condition instead of a hardcoded scope
- `isExternalWorkflowPackageSpecifier` resolves the specifier and asks the same question

No SDK list, no consumer config, package authors opt in by their own declaration. WDYT?

test_workflows/package.json
```diff
@@ -12,6 +12,7 @@
     "output:worker:watch": "npx nodemon --watch src --watch package.json --ext ts,js,json,prompt,md --ignore 'dist/**' --ignore '**/*.test.*' --exec 'npm run output:worker'"
   },
   "dependencies": {
+    "@growthxlabs/workflows_catalog": "0.0.12",
```
szanata, line 15: This package exists and it is used to test the feature. I will remove it from here later, when we have an official public workflows pages

bnchrch (commented):
deleted

bnchrch (commented):
deleted

bnchrch (commented):
This is way better! I love that the config is on the workflow publisher, not the installer. I couldnt leave inline comments due to github having a bug but heres a summary of a few things I found:

---

💅 **Config namespace** (`loader_tools.js:140`)

the config key is `output.workflows.expose` but our published scope is `@outputai/*`. WDYT about using `outputai.workflows.expose` instead? "output" is pretty generic and could collide with other tooling configs. Since this is a new public contract for third-party authors, feels like the right time to nail the namespace.

---

❗ **Shared activity collisions silently overwrite** (`loader.js:100`)

shared activities from external packages use the same namespace as local shared activities. If a package ships `normalizeTranscript` and the consumer also has a local `normalizeTranscript`, the later one silently wins. No error, no log.

Workflow names already have `validateWorkflowNames()` for exactly this. Could we add the same check for shared activities? Or at minimum a warning log when an overwrite happens?

---

❓ **Namespace imports silently fall through** (`npm_workflow_export_resolve.js:309`)

what happens when someone does `import * as catalog from '@acme/workflows'`? Looks like it falls through silently (returns `{ type: 'none' }`). At runtime they'd get a confusing error when they try to call `catalog.myWorkflow()`. Should we throw a clear error here instead? Something like "namespace imports from workflow packages aren't supported, use named imports"?

---

📚 **AST export-chain contract needs documenting** (`npm_workflow_export_resolve.js`)

the export-chain resolver has a specific contract: it only follows static `export { X } from` and `export * from` chains with relative paths. Dynamic re-exports, CJS interop, computed paths all silently fail to resolve. Package authors won't know this unless we tell them. Could we add a note in the docs about "your entry point must use static ESM re-exports"? The caveats section of the new docs page would be a good spot.

---

🤔 **JSON parse errors silently swallowed** (`loader_tools.js:141`)

`packageExposesWorkflows` swallows JSON parse errors and returns false. So if someone has a malformed package.json, their package just silently disappears from discovery. Same in the npm resolver. Not saying we should throw here. But maybe a debug-level log? "Skipping package X: failed to parse package.json"? Would save someone 30 minutes of debugging.

---

🔮 **Docs should mention the zero-wiring API path** (`external-workflow-packages.mdx`)

the docs focus on the "import and call as child workflow" pattern. But these external workflows are also auto-registered and callable via the API directly by name, right? For a lot of use cases (shared internal automations, workflow catalogs) that's the main path. People just install, restart the worker, and call via API. Might be worth a section on that. "You can also call external workflows directly via the API without writing any wrapper code."

szanata:
@bnchrch Thanks for the review, all the pointed issues were handled:
- _Config namespace_: https://github.com/growthxai/output/pull/179/commits/0bcefa834de6190c31772b4c406ca71580bd8ad7 - Renamed to `@outputai/config`;
- _Shared activity collisions silently overwrite_: https://github.com/growthxai/output/pull/179/commits/0422dbf34f0610deb503d412fa032f5207125d60 - Refactored to throw errors on duplicated activities + refactoring workflow validation to follow the same logic (inline vs discrete step) and removed case intensive validation as workflow names are case sensitive;
- _Namespace imports silently fall through_: https://github.com/growthxai/output/pull/179/commits/555ce0d22beec5757e0ace4daed2c683f92c83fc - Refactored to throw errors on unsupported import formats;
- _AST export-chain contract needs documenting_: https://github.com/growthxai/output/pull/179/commits/a2ea35b59cbf297a6ed70094e9cf05748e9c911e - Added docs;
- _JSON parse errors silently swallowed_: I didn't added the message because broken `package.json` can't be published to npm, and local packages (imported from `file://` protocol) will throw error during installation, so it is very unlikely that a `package.json` will ever have a invalid JSON syntax and if it has, the project will most likely not even start anyway;
- _Docs should mention the zero-wiring API path_: https://github.com/growthxai/output/pull/179/commits/4b2b7f11f943d4cbcda82ea89623f8be481fd6a1 - Added docs.

bnchrch (approved):
Looks great. Happy to approve.

One optional thing: 

**I recommend we use `outputai` instead of `@outputai/config` as the package.json key.**

The `@` and `/` makes it look like an npm package name, which implies we have a package called `@outputai/config` (we don't). Its also just unconventional. Most tools use a simple key like `eslintConfig`, `prettier`, `jest`. Something like:

```json
{
  "outputai": {
    "workflows": {
      "expose": true
    }
  }
}
```

reads way more naturally than:

```json
{
  "@outputai/config": {
    "workflows": {
      "expose": true
    }
  }
}
```

Not blocking the merge on it but I do think we should change

PR #183, opened 2026-05-01
fix: quote SKILL.md description containing colon

bryan-anthropic:
The frontmatter description in `coding_assistants/claude/plugins/outputai/skills/output-credentials-env-vars/SKILL.md` contains an unquoted colon (`credential:` mid-sentence), which YAML parses as a nested mapping rather than a plain scalar. Strict YAML validators reject the file with:

> YAML Parse error: Unexpected token

Wrapping the description in double quotes makes it parse cleanly while preserving the exact text. All other SKILL.md and agent frontmatter in the plugin parses without changes.

### Diff

```diff
-description: Wire encrypted credentials to environment variables using the credential: convention. ...
+description: "Wire encrypted credentials to environment variables using the credential: convention. ..."
```

### Verification

```ruby
require 'yaml'
content = File.read('skills/output-credentials-env-vars/SKILL.md')
YAML.safe_load(content.split('---', 3)[1])
# => parses cleanly, all fields preserved
```


marcosmarxm:
Thanks @bryan-anthropic !

marcosmarxm (approved):

PR #186, opened 2026-05-04
feat(claude): add model find and update skill

bnchrch:
## Problem

Model recommendations were scattered across ~10 skill files, 5 agents, and the CLI scaffolder, and they had drifted out of sync — Opus refs sat at 4-5/4-6 while 4.7 was current, `gpt-4o` was listed alongside `gpt-5` in different skills, and the workflow_prompt_writer agent's tier matrix disagreed with output-meta-pre-flight on the OpenAI default. The CLI scaffolded every new prompt as `claude-haiku-4-5`, so all 14 production `.prompt` files in `test_workflows/` ended up Haiku regardless of fit.

## Solution

image

- **`output-dev-model-selection`** *(new skill)* — owns model selection. `scripts/snapshot.sh` runs at skill-load time via the `!` fence and injects JSON of the top-10 most recently released models per provider, gateway payloads preserved untouched. Skill body teaches priority (reasoning/balance/speed/cost), provider mapping, family selection, ID translation. Defaults to stable / GA — skips `preview`/`alpha`/`beta` IDs unless explicitly requested.
- **`output-dev-upgrade-prompt-models`** *(new skill)* — bulk in-place model bump scoped to the existing family (Haiku stays Haiku; never escalates tiers or swaps providers). Reuses the snapshot from the selection skill.
- **9 existing skills + `workflow_prompt_writer` agent** — stripped opinionated model recommendations, kept concrete examples with `# current as of 2026-05-04 — run output-dev-model-selection for the latest` comments. `output-meta-pre-flight` no longer hardcodes `gpt-5`. `output-meta-project-context` drops the static provider/model table.
- **CLI scaffold templates** — default bumped from `claude-haiku-4-5` → `claude-sonnet-4-6` in 3 files (workflow scaffold, blog_evaluator example, workflow README).
- **`SESSION_START_CONTEXT.md`** — skill-count corrected (was stale at 33, now 49).


sdk/cli/src/templates/workflow/prompts/example@v1.prompt.template
```diff
@@ -1,6 +1,7 @@
 ---
 provider: anthropic
-model: claude-haiku-4-5
+# current as of 2026-05-04 — run output-dev-model-selection for the latest
```
catsby, line 3: The date `2026-05-04` itself isn't templated, so every project generated gets that date in there until we update it here. I don't believe the skills themselves are instructed to update this date when they update the models, are they? Maybe just remove the date, or make sure the skills update the prompts to include the new "current as of" comment(s)

coding_assistants/claude/plugins/outputai/skills/output-dev-upgrade-prompt-models/SKILL.md
```diff
@@ -0,0 +1,122 @@
...
+| `gemini-*-pro*` | `google-pro` |
+
+If a model doesn't match any pattern, skip the file and log a warning. Do not guess.
+
+### Step 4 — Look up latest in family
+
+Fetch the snapshot once at the start of the run and reuse it for every prompt file. The pipeline is the same one [`output-dev-model-selection`](../output-dev-model-selection/SKILL.md) runs at skill-load time:
+
+```bash
+SNAPSHOT=$(curl -s https://ai-gateway.vercel.sh/v1/models | jq '
```
catsby: if this were to fail in some way (network issue, etc), what kind of diff is presented to the users? Or does the update get `""` for the snapshot and then conclude that there's no update needed?

coding_assistants/claude/plugins/outputai/skills/output-dev-model-selection/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+# Picking a Model for an Output SDK Prompt
+
+This skill is the single source of truth for model selection across Output SDK skills and agents. Other skills link here instead of pinning specific model IDs, because model rosters drift faster than docs.
+
+## Live model snapshot
+
+The block below runs at skill-load time and emits JSON containing the **10 most recently released** models per provider, with each model's payload preserved verbatim from the Vercel AI Gateway.
+
+```!
+curl -s https://ai-gateway.vercel.sh/v1/models | jq '
```
catsby: what's the behavior of this still if this `curl` fails? Network issues, the shape changed, or the user doesn't have `jq` installed?

test_workflows/.claude/settings.json
```diff
@@ -1,13 +1,14 @@
...
+    "outputai@outputai": true
+  },
   "extraKnownMarketplaces": {
     "outputai": {
       "source": {
-        "source": "github",
-        "repo": "growthxai/output"
+        "source": "file",
+        "repo": "growthxai/output",
+        "path": "../.claude-plugin/marketplace.json"
```
catsby, line 9:
```suggestion
        "source": "file",
        "path": "../.claude-plugin/marketplace.json"
```

I don't think `repo` is valid when the source is `file`. It's likely harmless but technically not necessary I think

sdk/cli/src/templates/workflow/prompts/example@v1.prompt.template
bnchrch: Yup, good catch. Went with option B. Taught `output-dev-upgrade-prompt-models` to rewrite any `# current as of YYYY-MM-DD ...` line to today’s date in the same edit that bumps the `model:` line, so the model and the date refresh together and the convention stays coherent.

Fixed in 9f6cf97a.

coding_assistants/claude/plugins/outputai/skills/output-meta-project-context/SKILL.md
catsby: The `SESSION_START_CONTEXT.md` file says there are `49 skills`, the front matter of this file also says `49`, but we say `34` here and tables below have 36 total rows. 

Correcting them to be consistent is one thing, but I wonder if we gain much by including the count?

catsby (approved):
Minor inputs, no blockers

coding_assistants/claude/plugins/outputai/skills/output-dev-upgrade-prompt-models/SKILL.md
bnchrch: 🔮 Good call. Added an explicit guard in Step 4: if the snapshot fetch returns nothing, the skill bails loudly before Step 5 rather than silently reporting "no upgrades needed" against an empty dataset.

Fixed in 558dbda9.

test_workflows/.claude/settings.json
bnchrch: 💅 Good catch, dropped the `repo` key. Fixed in 558dbda9.

coding_assistants/claude/plugins/outputai/skills/output-meta-project-context/SKILL.md
bnchrch: 💎 Yup, the count was lying in three places already. Took your second suggestion and just dropped the parenthetical counts from the frontmatter and every section header in this file. They were going to keep drifting every time we add a skill. Names alone are sufficient inventory.

Fixed in 558dbda9.

PR #188, opened 2026-05-05
fix(llm): prevent tag injection from template variables in prompts

bnchrch:
## Problem

A prompt with multiple `<system>` blocks fails at the provider level (Anthropic rejects with "multiple system messages are not allowed"). Today this can happen accidentally:

- `loadPrompt` Liquid-renders the entire raw file, then `parsePrompt` runs a global non-greedy regex (`<(system|user|assistant|tool)>...</\1>`) over the rendered string.
- That regex can't tell developer-authored tags apart from tag-shaped substrings produced by `{{ variable }}` interpolation.
- Realistic trigger: a workflow that evaluates a webpage / chat transcript / prompt-engineering tutorial whose content already contains things like `</user>` followed by `<system>...</system>`. The injected `<system>` block is parsed as a second system message and the call is rejected.

## Solution

- Register an internal Liquid filter `__var_safe` that replaces `<` / `>` from variable output with `` / `` sentinels (Unicode Private Use Area — never appears in real content).
- Rewrite the raw template before render to append `| __var_safe` to every `{{ ... }}`, skipping `{% raw %}` regions. Control-flow tags (`{% if %}`, `{% for %}`) still emit real `<` / `>`, so dynamic block generation continues to work — only variable interpolation is armed.
- After `parsePrompt`, restore sentinels to real angle brackets in each parsed message body and in any string leaf of the parsed config (so frontmatter Liquid like `model: {{ id }}` keeps round-tripping).

The parser itself is unchanged.

## Test plan

- [x] `npm test` — full `@outputai/llm` suite passes (190 tests), including two new regression cases under `loadPrompt - tag injection from template variables`:
  - variable that closes `</user>` early and opens a new `<system>` block must not produce a second system message
  - tag-shaped substrings inside a variable must round-trip as inert text in the message body
- [x] `npm run lint` passes for the touched files
- [ ] Reviewer smoke: render any existing workflow prompt with a variable that contains `<system>` text and confirm the message body shows the literal tags

## Notes for reviewers

- **Sentinels (`` / ``) over HTML entities (`&lt;` / `&gt;`)** — entities would survive into the LLM input and change semantics. Private Use Area chars are inert during parsing and reverted before the messages leave `loadPrompt`.
- **Filter is appended (last in the chain)** rather than prepended so user-authored filters like `| upcase` run on the original value first; only the final string is escaped.
- **`{% raw %}` regions are explicitly skipped** by splitting on raw blocks before the rewrite. Without this, an authored `{% raw %}{{ x }}{% endraw %}` would emit the literal text `{{ x | __var_safe }}` instead of `{{ x }}`.

### DIFF
--- a/sdk/llm/src/prompt_loader.js
+++ b/sdk/llm/src/prompt_loader.js
@@ -7,7 +7,7 @@ import { FatalError } from '@outputai/core';
 
 const VAR_SAFE_FILTER = '__var_safe';
 
-const escapeXML = value =>
+export const escapeXML = value =>
   value === null || value === undefined ? '' : encodeXML( String( value ) );
 
 const liquid = new Liquid();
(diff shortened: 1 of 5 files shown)


szanata (approved):

sdk/llm/src/prompt_loader.js
```diff
@@ -1,14 +1,54 @@
...
+  return segments.map( ( segment, i ) => {
+    if ( i % 2 === 1 ) {
+      return segment;
+    }
+    return segment.replace(
+      /\{\{\s*([\s\S]+?)\s*\}\}/g,
+      ( _, expr ) => `{{ ${expr.trim()} | ${VAR_SAFE_FILTER} }}`
+    );
+  } ).join( '' );
+};
```
szanata: This could be a single .replace() doesnt need to split.

PR #193, opened 2026-05-05
feat: API Endpoint for Workflow History via Server Side Events (SSE)

catsby:
## Summary

This is a follow-up PR to https://github.com/growthxai/output/pull/134 where we introduced a workflow history endpoint so users could get the history information of a workflow. That endpoint was not live though and only gave information it had at that time - this PR adds server side events, providing live updates of events as they happen.

### Test plan
- [x] manual with scripts
- [x] unit tests


bnchrch (commented):
SSE history endpoint review. One thing GitHub won't let me pin inline, so it's here:

❗ **`api/src/index.js:1597` — shutdown hangs once a stream is open.** `server.close()` waits for every open connection to drain, and an SSE response never drains on its own (generator parked in a `waitNewEvent` long poll, 15s keepalive holding the socket). So one client on a running workflow means `await closeServer()` never resolves, `client.close()` never runs, and we exit only on SIGKILL. `POST /workflow/run` survives a long response because it has a `timeout`; the stream has no cap. Fix is `server.closeAllConnections()` (Node 18+) in `shutdown()`, ideally also ending the streams we track.

Rest of the comments are inline.

api/src/handlers/workflow_history_stream.js
```diff
@@ -0,0 +1,110 @@
...
+      } else if ( ctrl.signal.aborted ) {
+        // Real error masked by client disconnect — log at info so it stays observable.
+        logger.info( 'SSE stream error suppressed by client disconnect', {
+          workflowId, runId, error: error.constructor.name, message: error.message
+        } );
+      } else {
+        logger.error( 'SSE stream error', {
+          workflowId, runId, error: error.constructor.name, message: error.message, stack: error.stack
+        } );
+        const payload = { error: error.constructor.name, message: error.message, workflowId, runId };
```
bnchrch: ❗ `runId` comes from the query (line 18), which is only set on the `/runs/:rid/` route. On plain `/workflow/:id/history/stream` it's `undefined`, so:
- the `server_error` payload drops `runId`, and our OpenAPI/docs promise `{ error, message, workflowId, runId }`
- all four `logger.info/error` lines log `runId: undefined`, so a disconnect/error can't be tied back to a run

We already resolved it, it's sitting in `firstChunk.workflow.runId`. Grab it once:
```js
const { value: firstChunk, done } = await stream.next();
...
const resolvedRunId = firstChunk.workflow.runId;
```
and use `resolvedRunId` in the payload + logs.

🧪 The post-flush test only checks the error class and the `event: server_error` prefix, so it stays green with `runId` missing. Worth asserting `runId` there so the contract is actually covered.

api/src/clients/temporal/workflow/stream_history.js
```diff
@@ -0,0 +1,222 @@
...
+ * @param {object} options
+ * @param {string} [options.runId] - Specific run to target, defaults to latest
+ * @param {boolean} [options.includePayloads=false] - Decode input/output payloads
+ * @param {number} [options.lastEventId] - Resume after this event id (reconnect)
+ * @param {AbortSignal} [options.abortSignal] - Cancels in-flight gRPC calls
+ */
+export const streamHistory = async function *( { client, connection }, workflowId, options = {} ) {
+  const { runId, includePayloads = false, lastEventId, abortSignal } = options ?? {};
+  const handle = client.workflow.getHandle( workflowId, runId );
+  const description = await connection.withAbortSignal( abortSignal, () => handle.describe() ).catch( error => {
```
bnchrch: ❗ Every other gRPC call folds cancellation into a clean exit (`fetchPage` returns `null` on `isGrpcCancelledError`). This catch doesn't, it annotates and rethrows. So:
- client opens the stream, drops before the first chunk
- `req.on('close')` fires `ctrl.abort()`, `withAbortSignal` cancels the describe, it rejects CANCELLED
- that rejection escapes the first `stream.next()` (above the try, before `flushHeaders`) into Express, and `errorHandler` does `res.status(500).json()` on a socket that's already gone, plus a `logger.error`

So a normal disconnect logs as a server error. Mind giving describe the same treatment as `fetchPage`?
```js
.catch( error => {
  if ( isGrpcCancelledError( error ) ) {
    throw error; // benign abort, don't annotate/log as a real error
  }
  if ( error?.code === GrpcStatus.NOT_FOUND ) { ... }
  error.workflowId = workflowId;
  throw error;
} );
```
(or check `ctrl.signal.aborted` around the first `next()` in the handler and bail quietly.)

api/src/clients/temporal/workflow/stream_history.js
```diff
@@ -0,0 +1,222 @@
...
+  } );
+
+  while ( true ) {
+    const response = await fetchPage( true );
+
+    if ( response === null ) {
+      return;
+    }
+
+    state.nextPageToken = response.nextPageToken?.length ? response.nextPageToken : undefined;
```
bnchrch, line 112: ❗ The comment here nails the cause, but the cost is bigger than the dedupe. When the token resets to `undefined`, the next fetch starts at `FirstEventID=1` and we re-read + re-decode + re-filter the entire history every idle cycle. `filterEventId` stops us re-*sending* it, not re-*doing* it, and it scales with history length × open streams.

The SDK's own follow loop keeps the returned token instead of dropping it (`@temporalio/client` `workflow-client.js`: `req.nextPageToken = res.nextPageToken`). Can we do the same here, hold the last continuation token and keep polling from it rather than resetting to `undefined`?

api/src/middleware/error_handler.js
```diff
@@ -45,6 +46,14 @@ const grpcHttpStatus = err =>
   ( err ? directGrpcHttpStatus( err ) ?? grpcHttpStatus( err.cause ) : undefined );
 
 export default function errorHandler( error, req, res, _next ) {
+  if ( res.headersSent ) {
```
bnchrch, line 55: ❗ The morgan logger reads `res.locals.error` on finish to attach `errorType`/`errorMessage` (`http_logger.js:51`). This early return skips that, so any post-flush error that reaches here drops out of the access log (main always set it first). One-liner:
```js
if ( res.headersSent ) {
  res.locals.error = error;
  logger.error( `Post-flush error (headers already sent): ...`, { ... } );
  return;
}
```
🤔 Also, this is fixing an SSE problem in the global error handler, where it's dead for every other route today. Would it sit better in the stream handler's own catch? Not blocking, just feels like the wrong altitude.

api/src/handlers/workflow_history_stream.js
```diff
@@ -0,0 +1,110 @@
...
+        res.write( `event: server_error\ndata: ${JSON.stringify( payload )}\n\n` );
```
bnchrch: 💅 The keepalive `res.write` has a try/catch for the synchronous `ERR_STREAM_DESTROYED` case. This `server_error` write (and the `history`/`done` ones) don't, so in that same disconnect race they can throw and slip out post-flush. It's only harmless because the new `headersSent` guard catches it. A tiny writer helper makes them consistent:
```js
const safeWrite = chunk => {
  if ( res.writableEnded ) return;
  try { res.write( chunk ); } catch { ctrl.abort(); }
};
```
Not blocking, the guard saves us today.

api/src/index.js
```diff
@@ -105,7 +106,7 @@ app.use( ( req, res, next ) => {
 app.use( ( req, res, next ) => {
   res.header( 'Access-Control-Allow-Origin', '*' );
   res.header( 'Access-Control-Allow-Methods', 'GET, POST, PUT, DELETE, OPTIONS' );
-  res.header( 'Access-Control-Allow-Headers', 'Origin, X-Requested-With, Content-Type, Accept, Authorization' );
+  res.header( 'Access-Control-Allow-Headers', 'Origin, X-Requested-With, Content-Type, Accept, Authorization, Last-Event-ID' );
```
bnchrch, line 109: ❓ Heads up on the headline use case (browser live-progress UI). In prod the auth gate (`:102`) wants a Basic `Authorization` header, but:
- EventSource has no way to set request headers
- `Access-Control-Allow-Origin: *` rules out credentialed/cookie CORS
- there's no query-token fallback

So a cross-origin browser gets 401. Only same-origin browsers or non-browser clients (that set `Authorization` themselves) can use it. Adding `Last-Event-ID` to the allow-list is fine but doesn't change that. Not a security problem, the endpoint stays locked. But if browser SSE is the goal we'd want a `?token=` accepted on these routes, or at least a docs note. Was prod browser usage in scope, or is this dev/back-end-proxy only?

api/src/handlers/workflow_history_stream.js
```diff
@@ -0,0 +1,110 @@
...
+    try {
+      for await ( const chunk of stream ) {
+        if ( chunk.type === 'history' ) {
+          res.write( `id: ${chunk.lastEventId}\nevent: history\ndata: ${JSON.stringify( chunk.events )}\n\n` );
+        } else if ( chunk.type === 'done' ) {
+          const payload = { reason: chunk.reason };
+          if ( chunk.newRunId ) {
+            payload.newRunId = chunk.newRunId;
+          }
+          res.write( `event: done\ndata: ${JSON.stringify( payload )}\n\n` );
```
bnchrch: ❓ We send `done` then `res.end()`, but native EventSource reconnects on any close unless the client calls `.close()`. For an already-finished workflow that's a describe + history fetch every ~3s, forever, per lingering client. The docs example does call `es.close()`, but nothing here enforces it. Do we want to lean on that being documented as required? There isn't a clean protocol-level "stop", so this may just be a docs callout, curious what you think.

api/src/handlers/workflow_history_stream.js
```diff
@@ -0,0 +1,110 @@
...
+  } );
+
+  return async ( req, res ) => {
+    const workflowId = req.params.id;
+    const pathRunId = readPinnedRunId( req );
+    const { runId, includePayloads, lastEventId: queryLastEventId } = querySchema.parse(
+      pathRunId ? { ...req.query, runId: pathRunId } : req.query
+    );
+
+    const rawHeaderId = req.headers['last-event-id'];
```
bnchrch, line 32: 💅 If a proxy ever sends the header twice, Node joins it to `"15, 15"`, `safeParse` fails, and we silently replay from event 1. Edge case, but taking the last token would dodge the surprise. Total nit, feel free to ignore.

api/src/clients/temporal/workflow/stream_history.js
```diff
@@ -0,0 +1,222 @@
...
+    isWorkflowClosed( workflowStatus ) &&
+    lastEventId >= historyLength &&
+    !STATUS_HAS_NEW_RUN_ID.has( workflowStatus )
+  ) {
+    // Fast-path: status excluded above guarantees no newRunId is possible here.
+    yield doneChunk( EventTypeName[STATUS_TO_TERMINAL_EVENT_TYPE[workflowStatus]] );
+    return;
+  }
+
+  const normalizeEventType = event => (
```
bnchrch: 💅 This is the same `typeof event.eventType === 'object' ? Number(...) : ...` coercion that's inside `serializeEvent` (`event_serialization.js`). Worth exporting one helper. Beyond DRY: terminal detection runs on raw events and serialization runs on its own copy, so if the wire type ever shifts they could disagree about what counts as terminal.

api/src/clients/temporal/workflow/stream_history.js
```diff
@@ -0,0 +1,222 @@
...
+  } );
+
+  const resolvedRunId = description.runId;
+  if ( !resolvedRunId ) {
+    throw new Error( `Temporal did not report a runId for workflow "${workflowId}"` );
+  }
+  const workflowStatus = description.status.code;
+  const historyLength = description.historyLength;
+
+  const workflow = {
```
bnchrch: 💅 The describe → `{ workflowId, runId, status, startTime, closeTime, historyLength, taskQueue }` mapping plus the NOT_FOUND translation is lifted straight from `get_history.js`. A shared `describeWorkflow(context, workflowId, { runId, abortSignal }) => { workflow, resolvedRunId }` keeps the shape and the not-found wording in one place, and `get_history` picks up `withAbortSignal` for free.

api/src/clients/temporal/workflow/stream_history.js
```diff
@@ -0,0 +1,222 @@
...
+  [WorkflowStatus.CANCELED]: EventType.WORKFLOW_EXECUTION_CANCELED,
+  [WorkflowStatus.TERMINATED]: EventType.WORKFLOW_EXECUTION_TERMINATED,
+  [WorkflowStatus.CONTINUED_AS_NEW]: EventType.WORKFLOW_EXECUTION_CONTINUED_AS_NEW,
+  [WorkflowStatus.TIMED_OUT]: EventType.WORKFLOW_EXECUTION_TIMED_OUT
+};
+
+// Statuses whose terminal history event can carry `newExecutionRunId`. The fast-path
+// (skip history fetch when reconnect is past historyLength) cannot serve these because
+// `describe()` does not expose `newExecutionRunId` -- only the terminal event does.
+const STATUS_HAS_NEW_RUN_ID = new Set( [
```
bnchrch: 💅 `STATUS_TO_TERMINAL_EVENT_TYPE` and `STATUS_HAS_NEW_RUN_ID` only exist to serve the reconnect-after-close fast-path, which saves a single history fetch on a cold path. The streaming path already finds the terminal event in history and only needs `TERMINAL_EVENT_TYPES` + `NEW_RUN_ID_ATTRS`.

🔮 The trap: add a terminal status later (Temporal already has PAUSED/UNPAUSED at 58/59), wire up the other maps, forget `STATUS_HAS_NEW_RUN_ID`, and the fast-path silently emits `done` with no `newRunId`. No test catches it. Is the saved fetch worth four hand-synced maps, or could we drop the fast-path and always read the terminal event?

scripts/test-history.sh
szanata, line 1: These files doesnt follow the pattern of using `_` in file names in this repo.

api/src/handlers/workflow_history_stream.js
```diff
@@ -0,0 +1,110 @@
...
+      includePayloads,
+      lastEventId,
+      abortSignal: ctrl.signal
+    } );
+
+    const { value: firstChunk, done } = await stream.next();
+    if ( done || firstChunk?.type !== 'workflow' ) {
+      throw Object.assign(
+        new Error( `streamHistory did not yield workflow metadata as first chunk (workflowId: ${workflowId})` ),
+        { workflowId }
```
szanata: This is a bit weird. Why not create a custom error class that accepts workflowId in its constructor?

api/src/handlers/workflow_history_stream.js
```diff
@@ -0,0 +1,110 @@
...
+      for await ( const chunk of stream ) {
+        if ( chunk.type === 'history' ) {
+          res.write( `id: ${chunk.lastEventId}\nevent: history\ndata: ${JSON.stringify( chunk.events )}\n\n` );
+        } else if ( chunk.type === 'done' ) {
+          const payload = { reason: chunk.reason };
+          if ( chunk.newRunId ) {
+            payload.newRunId = chunk.newRunId;
+          }
+          res.write( `event: done\ndata: ${JSON.stringify( payload )}\n\n` );
+          break;
```
szanata, line 141: isnt `chunk.type === done` the last chunk already?

api/src/handlers/workflow_history_stream.js
```diff
@@ -0,0 +1,110 @@
...
+        }
+      }
+    } catch ( error ) {
+      if ( isGrpcCancelledError( error ) ) {
+        // Pure cancellation from gRPC client teardown: nothing to log.
+      } else if ( ctrl.signal.aborted ) {
+        // Real error masked by client disconnect — log at info so it stays observable.
+        logger.info( 'SSE stream error suppressed by client disconnect', {
+          workflowId, runId, error: error.constructor.name, message: error.message
```
szanata: Is the stack really necessary on these logs? they arent unexpected, and the last call-site is already know. I think this is noisy.

api/src/middleware/error_handler.js
szanata: I think this doesnt belong here as well

api/src/middleware/error_handler.js
```diff
@@ -45,6 +46,14 @@ const grpcHttpStatus = err =>
...
+    logger.error( `Post-flush error (headers already sent): ${error.constructor.name}: ${error.message}`, {
+      requestId: req?.id,
+      stack: isProduction ? undefined : error.stack
```
szanata: We explicitly configured our logger in prod to have error stack:
```js
  winston.format.errors( { stack: true } ),
```

So why are we overwriting it here?

Also, for this specific log, I dont think we need stack at all, since it is a "expected" failure.

api/src/middleware/error_handler.js
```diff
@@ -45,6 +46,14 @@ const grpcHttpStatus = err =>
...
+    } );
+    return;
```
szanata: this early return in the chain of events of the Express middleware reads wrong. I agree with Ben's comment. This should be handled in the endpoint itself.

scripts/test-history-stream.sh
szanata, line 1: I think these test files are out of place. They should be placed under `api/scripts/`, not in the root folder.

But I think they will get out of date and not functional very soon.

My suggestions would be to convert this to a workflow in `test_workflows/`, you can have a workflow that triggers another workflow and calls this endpoint and streams the result to the console/file.

When putting there, it is easier to test and keep up to date, as that is actual compiled code.

catsby: the scripts were only meant for a human to verify the stream/history endpoint, I had them locally for testing/verification. I'll just remove them

catsby: removed in 249ef88

scripts/test-history.sh
catsby: removed in 249ef88

api/src/handlers/workflow_history_stream.js
catsby: Good catch, fixed in 44bd8d8

api/src/clients/temporal/workflow/stream_history.js
catsby: 👍 also fixed in 44bd8d8

api/src/clients/temporal/workflow/stream_history.js
catsby: I originally changed this as requested in 44bd8d8 but then reverted it in 4ff4a50 after some testing. I found the case you mentioned never happened in practice - I added some logging to check and `nextPageToken` was never _not_ populated (never `undefined` etc), it kept advancing. The code comment that said "An empty long-poll can reset nextPageToken to undefined" seems to be unsubstantiated. 

> Can we do the same here, hold the last continuation token and keep polling from it rather than resetting to undefined?

The SDK code mentioned (`req.nextPageToken = res.nextPageToken`) does basically the same thing. `nextPageToken` comes from gRPC, it's basically `[]bytes`, but it's still a `truthy` thing. Here we check the length and we change it to `undefined`, which is `falsy` and we use it later in checks to see if we're actually done or not and we can drain it (checked around line 206)

api/src/handlers/workflow_history_stream.js
catsby: Technically yes, but using `break` here makes the termination explicit to any person reading it. The code would exit the `for await` on the next loop regardless, but this makes it explicit and saves a loop. The explicit exit is the point though, not the saving of one eval of `chunk of stream`

api/src/index.js
catsby: I don't believe browser based SSE is the goal

api/src/handlers/workflow_history_stream.js
catsby: I added a `204 no content` short-circuit here instead

api/src/handlers/workflow_history_stream.js
catsby: fixed

api/src/clients/temporal/workflow/stream_history.js
catsby: I refactored this to clean it up

api/src/clients/temporal/workflow/stream_history.js
catsby: removed these extra maps and made simpler

api/src/handlers/workflow_history_stream.js
catsby: refactored to be it's own error class

api/src/handlers/workflow_history_stream.js
catsby: removed the stack

api/src/middleware/error_handler.js
catsby: removed the stack

api/src/handlers/workflow_history_stream.js
catsby: refactored to included this

api/src/middleware/error_handler.js
catsby: refactored to address this

api/src/clients/temporal/workflow/stream_history.js
catsby: refactored to fix this

bnchrch (approved):

szanata (approved):
Solid 👍

PR #195, opened 2026-05-07
fix(cli): Scenarios not loading in the RunModal when wf name differs from dir name

szanata:
Scenarios were not loading in the CLI's RunModal when workflow name differs from folder name.

## Summary

Fix dev run modal scenario discovery for workflows whose names do not match their folder paths.

Use the loaded workflow catalog path to find the local `scenarios/` directory before falling back to name-based lookup.

Apply the same path-aware resolution when running, duplicating, and saving scenarios.

### Example Fixed
A workflow named writing_editor located at:
`src/workflows/writing/editor/workflow.ts`

with scenarios in:
`src/workflows/writing/editor/scenarios/`

previously showed no scenarios because the modal looked for:
`src/workflows/writing_editor/scenarios/`

Now the modal resolves the real workflow directory and lists those scenarios correctly.

## Test plan

[x] unit
[x] e2e
[x] manual


docs/guides/TODO_pages.md
bnchrch, line 1: We should probably just delete this page

bnchrch (approved):

PR #197, opened 2026-05-07
fix(cli): Surfacing docker compose startup failures in output dev

szanata:
## Summary

Fixes `output dev` appearing to hang until the health timeout when `docker compose up` fails before creating containers.

Previously, the dev command started Compose and then relied on `docker compose ps` polling from the Ink UI. If Compose exited early, there were no containers to poll, so the UI kept waiting and eventually timed out without showing the real Compose error.

## Changes

- Captures and drains `docker compose up` stdout/stderr while the Ink UI owns the terminal.
- Reports Compose process failures immediately through the dev UI instead of waiting for the service health timeout.
- Includes recent Compose output in the surfaced error, so failures such as invalid config or host port conflicts are visible.
- Makes `docker compose ps` use the same `--project-directory` as `docker compose up`, so status polling checks the same Compose project context that was started.
- Tightens service health checks so only `running` containers can be considered healthy. Containers in `created` state no longer pass just because they have no healthcheck.
- Refactors `startDockerCompose` to accept an options object and return the `ChildProcess` directly. The Docker service now owns output capture and injects captured output into `onError` / `onExit` callbacks.

## Why

The root bug was observability around early Compose startup failure. `output dev` could fail to create any containers, but the CLI only showed a generic timeout because the actual Compose process exit and output were not surfaced to the user.

The fix keeps the TUI clean while still preserving the important startup diagnostics. Compose output is not written directly over Ink, but recent output is retained and shown when Compose exits with an error.

## Tests plan

Added regression coverage for:

- Compose exit failures surfacing recent output through `output dev`.
- Cleanup-triggered Compose exits being ignored.
- `startDockerCompose` wiring process handlers and passing captured stdout/stderr to callbacks.
- Status polling using `--project-directory`.
- `created` containers not counting as healthy.

Plus, tested and passing:
- manual testing (executing workflows);
- unit tests (all);
- e2e test

### DIFF
--- a/sdk/cli/src/services/docker.ts
+++ b/sdk/cli/src/services/docker.ts
@@ -16,6 +16,7 @@ export const SERVICE_HEALTH = {
 
 export const SERVICE_STATE = {
   RUNNING: 'running',
+  CREATED: 'created',
   EXITED: 'exited'
 } as const;
 
@@ -144,7 +145,13 @@ export function parseServiceStatus( jsonOutput: string ): ServiceStatus[] {
 export async function getServiceStatus( dockerComposePath: string ): Promise<ServiceStatus[]> {
   const result = execFileSync(
     'docker',
-    [ 'compose', '-f', dockerComposePath, '--project-name', config.dockerServiceName, 'ps', '--all', '--format', 'json' ],
+    [
+      'compose',
+      '-f', dockerComposePath,
+      '--project-directory', process.cwd(),
+      '--project-name', config.dockerServiceName,
+      'ps', '--all', '--format', 'json'
+    ],
     { encoding: 'utf-8', cwd: process.cwd() }
   );
 
@@ -152,7 +159,7 @@ export async function getServiceStatus( dockerComposePath: string ): Promise<Ser
 }
 
 export function isServiceHealthy( service: ServiceStatus ): boolean {
-  return service.state !== SERVICE_STATE.EXITED &&
+  return service.state === SERVICE_STATE.RUNNING &&
     ( service.health === SERVICE_HEALTH.HEALTHY || service.health === SERVICE_HEALTH.NONE );
 }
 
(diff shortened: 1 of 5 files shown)


sdk/cli/src/services/docker.ts
```diff
@@ -16,6 +16,7 @@ export const SERVICE_HEALTH = {
 
 export const SERVICE_STATE = {
   RUNNING: 'running',
+  CREATED: 'created',
```
catsby, line 19: Is this used somewhere I'm not finding? I see the spec using `state: 'created'`, not this enum though

catsby (approved):

sdk/cli/src/services/docker.ts
szanata: It is just to document this other state and make it clear that it exists. We can remove, but I saw no harm in keeping it.

sdk/cli/src/services/docker.ts
```diff
@@ -144,15 +145,21 @@ export function parseServiceStatus( jsonOutput: string ): ServiceStatus[] {
...
+    ],
     { encoding: 'utf-8', cwd: process.cwd() }
   );
 
   return parseServiceStatus( result );
 }
 
 export function isServiceHealthy( service: ServiceStatus ): boolean {
-  return service.state !== SERVICE_STATE.EXITED &&
+  return service.state === SERVICE_STATE.RUNNING &&
```
bnchrch, line 162: This may treat `IN_PROGRESS` as unhealthy.

bnchrch (approved):

sdk/cli/src/services/docker.ts
szanata: I am considering that healthy is a service up and running, anything else is unhealthy. This should not be confused with "failed".

PR #203, opened 2026-05-13
fix: Handle long trace tree objects IO [OUT-442]

szanata:
## Summary

After generating the trace tree (the complete trace for a workflow run), the content was stringified before beign written to file/ uploaded to s3. This was throwing `"RangeError: Invalid string length"` when the string was violating Node's limit.

This solution replace the in-memory serialization (`JSON.stringify()`) with a stream solution, that stringifying while streaming the result, using far less memory.

## Test plan

[x] unit
[x] manual
[x] e2e


bnchrch (approved):
Good fix. Streaming is the right approach for traces that can blow past Node's string limit.

💅 `big-json` brings in ~20 transitive deps for a one-liner that delegates to `json-stream-stringify`. `createStringifyStream` is literally just `new JsonStreamStringify(opts.body)`. `json-stream-stringify` has zero dependencies and 2.9M weekly downloads (vs 185K for big-json). The swap is trivial:

```js
// big-json (current)
import { createStringifyStream } from 'big-json';
createStringifyStream({ body: content })

// json-stream-stringify (proposed)
import JsonStreamStringify from 'json-stream-stringify';
new JsonStreamStringify(content)
```

Same result. ~20 fewer packages in the lockfile.

szanata:
@bnchrch +1 good call.

PR #206, opened 2026-05-14
feat: Adding costs/llm_usage breakdown and total to workflow results [OUT-449]

szanata:
## Summary

Add a new properties `attributes: []` and `aggregations: {}` to the workflow execution result from `@outputai/core` and in the API.

The object contains the value of each attribute added via `Trace.addEventAttribute()`, plus aggregate cost/usage information based on the attributes:

#### LLMUsage
```js
{
  type: 'llm:usage',
  modelId: string,
  usage: [
    {
      type: string,
      ppm: number,
      amount: number,
      total: number
    }
  ],
  total: number,
  tokensUsed: number
}
```

#### HTTPRequestCount
```js
{
  type: 'http:request:count',
  url: string,
  requestId: string
}
```

#### HTTPRequestCost
```js
{
  type: 'http:request:cost',
  url: string,
  requestId: string,
  total: number
}
```

### Aggregations shape

```js
{
  "cost": {
    "total": number
  },
  "tokens": {
    "total": number,
    "<type>": number // where <type> is each type added to LLMUsage.usage[]
  },
  "httpRequests": {
    "total": 5
  }
}
```

### Examples
<details>

<summary>Workflow result with costs from LLM calls in activities</summary>

```json
{
  "workflowId": "qFDWmuY3D2ZBJzxse6U-0",
  "runId": "019e289e-4e23-7615-8998-027b8aad0e6f",
  "status": "completed",
  "input": "<<omitted>>",
  "output": "<<omitted>>",
  "trace": "<<omitted>>",
  "attributes": [
    {
      "activityId": "2",
      "activityName": "prompt#explainTopic",
      "date": 1778797927462,
      "type": "llm:usage",
      "modelId": "claude-haiku-4-5",
      "usage": [
        {
          "type": "input",
          "ppm": 1,
          "amount": 38,
          "total": 0.000038
        },
        {
          "type": "input_cached",
          "ppm": 0.1,
          "amount": 0,
          "total": 0
        },
        {
          "type": "output",
          "ppm": 5,
          "amount": 231,
          "total": 0.001155
        }
      ],
      "total": 0.001193,
      "tokensUsed": 269
    },
    {
      "activityId": "3",
      "activityName": "prompt#generateCookingInstruction",
      "date": 1778797930387,
      "type": "llm:usage",
      "modelId": "claude-haiku-4-5",
      "usage": [
        {
          "type": "input",
          "ppm": 1,
          "amount": 321,
          "total": 0.000321
        },
        {
          "type": "input_cached",
          "ppm": 0.1,
          "amount": 0,
          "total": 0
        },
        {
          "type": "output",
          "ppm": 5,
          "amount": 186,
          "total": 0.00093
        }
      ],
      "total": 0.001251,
      "tokensUsed": 507
    },
    {
      "activityId": "4",
      "activityName": "prompt#generateDrawingInstructions",
      "date": 1778797935482,
      "type": "llm:usage",
      "modelId": "claude-haiku-4-5",
      "usage": [
        {
          "type": "input",
          "ppm": 1,
          "amount": 311,
          "total": 0.000311
        },
        {
          "type": "input_cached",
          "ppm": 0.1,
          "amount": 0,
          "total": 0
        },
        {
          "type": "output",
          "ppm": 5,
          "amount": 453,
          "total": 0.002265
        }
      ],
      "total": 0.002576,
      "tokensUsed": 764
    },
    {
      "activityId": "5",
      "activityName": "prompt#generateChoice",
      "date": 1778797936661,
      "type": "llm:usage",
      "modelId": "claude-haiku-4-5",
      "usage": [
        {
          "type": "input",
          "ppm": 1,
          "amount": 217,
          "total": 0.000217
        },
        {
          "type": "input_cached",
          "ppm": 0.1,
          "amount": 0,
          "total": 0
        },
        {
          "type": "output",
          "ppm": 5,
          "amount": 9,
          "total": 0.000045
        }
      ],
      "total": 0.000262,
      "tokensUsed": 226
    }
  ],
  "aggregations": {
    "cost": {
      "total": 0.005282
    },
    "tokens": {
      "total": 1766,
      "input": 887,
      "input_cached": 0,
      "output": 879
    },
    "httpRequests": {
      "total": 0
    }
  },
  "error": null
}
```

</details>

<details>

<summary>Workflow result with calls and costs from HTTP requests</summary>

```json
{
  "workflowId": "0X2122Xo4zIhtMzfFKPdy",
  "runId": "019e2897-ae7a-7333-b364-214164ad6b3a",
  "status": "completed",
  "input": "<<omitted>>",
  "output": "<<omitted>>",
  "trace": "<<omitted>>",
  "attributes": [
    {
      "activityId": "2",
      "activityName": "http#listClientsStep",
      "date": 1778797489839,
      "type": "http:request:count",
      "url": "https://httpbin.io/anything/clients/",
      "requestId": "e3630b1f-27ce-47f6-a6d1-ab97ed743e02"
    },
    {
      "activityId": "3",
      "activityName": "http#createClientStep",
      "date": 1778797490395,
      "type": "http:request:count",
      "url": "https://httpbin.io/anything/clients/",
      "requestId": "4a87b754-42bd-4fe5-908f-6ec3429f4c1c"
    },
    {
      "activityId": "4",
      "activityName": "http#exportClientsStep",
      "date": 1778797490522,
      "type": "http:request:count",
      "url": "https://httpbin.io/anything/clients/export",
      "requestId": "ac4f6e90-c9d3-4609-bcb4-5ea74732955e"
    },
    {
      "activityId": "5",
      "activityName": "http#listContractsStep",
      "date": 1778797490641,
      "type": "http:request:count",
      "url": "https://httpbin.io/anything/contracts/",
      "requestId": "5ca2abe5-947c-426e-9890-deb1c408cbc0"
    },
    {
      "activityId": "5",
      "activityName": "http#listContractsStep",
      "date": 1778797490747,
      "type": "http:request:cost",
      "url": "https://httpbin.io/anything/contracts/",
      "requestId": "5ca2abe5-947c-426e-9890-deb1c408cbc0",
      "total": 120
    },
    {
      "activityId": "6",
      "activityName": "http#createContractStep",
      "date": 1778797490764,
      "type": "http:request:count",
      "url": "https://httpbin.io/anything/contracts/",
      "requestId": "566ec19e-53a7-440c-82ff-a53dbd45225e"
    }
  ],
  "aggregations": {
    "cost": {
      "total": 120
    },
    "tokens": {
      "total": 0
    },
    "httpRequests": {
      "total": 5
    }
  },
  "error": null
}
```

</details>

### Important
- On activity failures, attributes are preserved;
- On activity retry, attributes from previous attempts are preserved;
- Child workflow attributes are concatenated to the parent attributes;
- Failed child workflow attributes are also concatenated to the parent attributes;
- Failed workflows still have attributes/aggregations (attached to error, like trace information);
- The final aggregation is calculate at the end of the workflow;
 
## Test plan

[x] unit
[x] e2e
[x] manual

### DIFF
--- a/sdk/core/src/interface/aggregations.js
+++ b/sdk/core/src/interface/aggregations.js
@@ -5,16 +5,17 @@ export const aggregateAttributes = attributes => ( {
   cost: {
     total: attributes
       .filter( a => [ Attribute.HTTPRequestCost.TYPE, Attribute.LLMUsage.TYPE ].includes( a.type ) )
-      .reduce( ( sum, a ) => Decimal( a.total ).add( sum ).toNumber(), 0 )
+      .reduce( ( sum, a ) => sum.add( a.total ), Decimal( 0 ) ).toNumber()
   },
   tokens: {
     total: attributes
       .filter( a => Attribute.LLMUsage.TYPE === a.type )
-      .reduce( ( sum, a ) => Decimal( a.tokensUsed ).add( sum ).toNumber(), 0 ),
-    ...attributes
+      .reduce( ( sum, a ) => sum.add( a.tokensUsed ), Decimal( 0 ) ).toNumber(),
+    ...Object.entries( attributes
       .filter( a => Attribute.LLMUsage.TYPE === a.type )
       .flatMap( a => a.usage )
-      .reduce( ( obj, a ) => Object.assign( obj, { [a.type]: Decimal( obj[a.type] ?? 0 ).add( a.amount ).toNumber() } ), {} )
+      .reduce( ( obj, a ) => Object.assign( obj, { [a.type]: ( obj[a.type] ?? Decimal( 0 ) ).add( a.amount ) } ), {} ) )
+      .reduce( ( obj, [ k, v ] ) => Object.assign( obj, { [k]: v.toNumber() } ), {} ) // convert all values to number
 
   },
   httpRequests: {
--- a/sdk/core/src/interface/aggregations.js
+++ b/sdk/core/src/interface/aggregations.js
@@ -1,0 +1,23 @@
+import { Attribute } from '#trace_attribute';
+import Decimal from 'decimal.js';
+
+export const aggregateAttributes = attributes => ( {
+  cost: {
+    total: attributes
+      .filter( a => [ Attribute.HTTPRequestCost.TYPE, Attribute.LLMUsage.TYPE ].includes( a.type ) )
+      .reduce( ( sum, a ) => Decimal( a.total ).add( sum ).toNumber(), 0 )
+  },
+  tokens: {
+    total: attributes
+      .filter( a => Attribute.LLMUsage.TYPE === a.type )
+      .reduce( ( sum, a ) => Decimal( a.tokensUsed ).add( sum ).toNumber(), 0 ),
+    ...attributes
+      .filter( a => Attribute.LLMUsage.TYPE === a.type )
+      .flatMap( a => a.usage )
+      .reduce( ( obj, a ) => Object.assign( obj, { [a.type]: Decimal( obj[a.type] ?? 0 ).add( a.amount ).toNumber() } ), {} )
+
+  },
+  httpRequests: {
+    total: attributes.filter( a => Attribute.HTTPRequestCount.TYPE === a.type ).length
+  }
+} );
--- a/sdk/core/src/interface/workflow.js
+++ b/sdk/core/src/interface/workflow.js
@@ -3,7 +3,7 @@ import { proxyActivities, inWorkflowContext, executeChild, workflowInfo, uuid4, 
 import { defineSignal, setHandler } from '@temporalio/workflow';
 import { validateWorkflow } from './validations/static.js';
 import { validateWithSchema } from './validations/runtime.js';
-import { SHARED_STEP_PREFIX, ACTIVITY_GET_TRACE_DESTINATIONS, METADATA_ACCESS_SYMBOL } from '#consts';
+import { SHARED_STEP_PREFIX, ACTIVITY_GET_TRACE_DESTINATIONS, METADATA_ACCESS_SYMBOL, Signal } from '#consts';
 import { deepMerge, setMetadata, toUrlSafeBase64 } from '#utils';
 import { FatalError, ValidationError } from '#errors';
 import { Context } from './workflow_context.js';
@@ -76,7 +76,7 @@ export function workflow( { name, description, inputSchema, outputSchema, fn, op
     const traceDestinations = isRoot ? ( await steps[ACTIVITY_GET_TRACE_DESTINATIONS]( executionContext ) ) : null;
 
     const attributes = [];
-    setHandler( defineSignal( 'add_attribute' ), e => attributes.push( e ) );
+    setHandler( defineSignal( Signal.ADD_ATTRIBUTE ), e => attributes.push( e ) );
 
     try {
       // validation comes after setting memo to have that info already set for interceptor even if validations fail
--- a/sdk/core/src/interface/workflow.js
+++ b/sdk/core/src/interface/workflow.js
@@ -7,7 +7,7 @@ import { SHARED_STEP_PREFIX, ACTIVITY_GET_TRACE_DESTINATIONS, METADATA_ACCESS_SY
 import { deepMerge, setMetadata, toUrlSafeBase64 } from '#utils';
 import { FatalError, ValidationError } from '#errors';
 import { Context } from './workflow_context.js';
-import { formatCost } from './events.js';
+import { aggregateAttributes } from './aggregations.js';
 
 const defaultOptions = {
   activityOptions: {
@@ -23,6 +23,9 @@ const defaultOptions = {
   },
   disableTrace: false
 };
+
+export const extractErrorDetail = ( e, key ) =>
+  e ? ( e.details?.find?.( d => d[key] )?.[key] ?? extractErrorDetail( e.cause, key ) ) : null;
 
 export function workflow( { name, description, inputSchema, outputSchema, fn, options = {}, aliases = [] } ) {
   validateWorkflow( { name, description, inputSchema, outputSchema, fn, options, aliases } );
@@ -71,10 +74,9 @@ export function workflow( { name, description, inputSchema, outputSchema, fn, op
 
     // Run the internal activity to retrieve the workflow trace destinations (only for root workflows, not nested)
     const traceDestinations = isRoot ? ( await steps[ACTIVITY_GET_TRACE_DESTINATIONS]( executionContext ) ) : null;
-    const traceObject = { trace: { destinations: traceDestinations } };
 
-    const costEvents = [];
-    setHandler( defineSignal( 'add_event:cost' ), e => costEvents.push( e ) );
+    const attributes = [];
+    setHandler( defineSignal( 'add_attribute' ), e => attributes.push( e ) );
 
     try {
       // validation comes after setting memo to have that info already set for interceptor even if validations fail
@@ -96,17 +98,25 @@ export function workflow( { name, description, inputSchema, outputSchema, fn, op
          * @param {import('@temporalio/workflow').ActivityOptions} extra.options
          * @returns {Promise<unknown>}
          */
-        startWorkflow: async ( childName, input, extra = {} ) =>
-          executeChild( childName, {
-            args: input ? [ input ] : [],
-            workflowId: `${workflowId}-${toUrlSafeBase64( uuid4() )}`,
-            parentClosePolicy: ParentClosePolicy[extra?.detached ? 'ABANDON' : 'TERMINATE'],
-            memo: {
-              executionContext,
-              parentId: workflowId,
-              ...( extra?.options?.activityOptions && { activityOptions: deepMerge( activityOptions, extra.options.activityOptions ) } )
-            }
-          } )
+        startWorkflow: async ( childName, input, extra = {} ) => {
+          try {
+            const result = await executeChild( childName, {
+              args: input ? [ input ] : [],
+              workflowId: `${workflowId}-${toUrlSafeBase64( uuid4() )}`,
+              parentClosePolicy: ParentClosePolicy[extra?.detached ? 'ABANDON' : 'TERMINATE'],
+              memo: {
+                executionContext,
+                parentId: workflowId,
+                ...( extra?.options?.activityOptions && { activityOptions: deepMerge( activityOptions, extra.options.activityOptions ) } )
+              }
+            } );
+            attributes.push( ...( result.attributes ?? [] ) );
+            return result.output;
+          } catch ( error ) {
+            attributes.push( ...( extractErrorDetail( error, 'attributes' ) ?? [] ) );
+            throw error;
+          }
+        }
       };
 
       const output = await fn.call( dispatchers, input, context );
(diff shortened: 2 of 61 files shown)


bnchrch (commented):
Big-picture pause before the line-level stuff. The two things I want to push on are (1) the final shape of `result.attributes` and (2) how we register a new summable concept. They're tangled but the same fix solves both. Please read this in a friendly tone, none of this is "you did it wrong," it's all "knowing we're about to need tokens next, where can we land?"

### The shape today

We've got `result.cost = { events, total }` at the top level, plus per-LLM-event tokens smuggled into `cost.components[i].tokens` and `cost.info.tokens`. That smuggling is the tell. Tokens want to be a thing of their own, but the shape only has a slot for cost. The next concept (latency, retries, third-party API spend, whatever) hits the same wall.

### What I'd love to see instead

One namespace for summable concepts, uniform skeleton, subcategories live in the event name itself:

```js
{
  output, trace, error,
  attributes: {
    cost: {
      total: 5.23,
      components: {
        llm:  { total: 4.73, events: [...] },
        http: { total: 0.50, events: [...] }
      }
    },
    token_usage: {
      total: { input: 1234, output: 567, cached: 0 },
      components: {
        llm: { total: { input: 1234, output: 567, cached: 0 }, events: [...] }
      }
    }
  }
}
```

Same skeleton for every concept. `total` is whatever the aggregate function returns (a number for cost, a small object for token_usage). `components[sub]` is the per-source split. Crucially: **the producer's choice of event name is the components key**. The HTTP SDK attaches `cost:http`, the LLM SDK attaches `cost:llm` and `token_usage:llm`, a future DB layer attaches `cost:db`. Nothing in the framework needs to know what sources exist.

### Registration

Today, one summable concept = 7 hardcoded touch points across 3 files. The `Attribute.COST` const, the `if (name === Attribute.COST)` in `addEventActionWithContext`, the signal name string, the `defineSignal` + `setHandler` in `workflow.js`, the accumulator array, `formatCost`, and the result wiring. Doing this again for tokens means doubling all of it.

What if registration looked like this, colocated with the SDK that produces the concept:

```js
// sdk/llm/src/cost/register.js
registerSummableAttribute( {
  prefix: 'cost',
  initial: 0,
  aggregate: ( events, init ) => events.reduce( ( s, e ) => s + ( e.value.total ?? 0 ), init )
} );

registerSummableAttribute( {
  prefix: 'token_usage',
  initial: { input: 0, output: 0, cached: 0 },
  aggregate: ( events, init ) => events.reduce( ( a, e ) => ( {
    input:  a.input  + ( e.value.inputTokens ?? 0 ),
    output: a.output + ( e.value.outputTokens ?? 0 ),
    cached: a.cached + ( e.value.cachedInputTokens ?? 0 )
  } ), init )
} );
```

Producers attach the full prefixed event name:

```js
// LLM SDK
Tracing.addEventAttribute( { eventId, name: 'cost:llm',        value: cost } );
Tracing.addEventAttribute( { eventId, name: 'token_usage:llm', value: usage } );

// HTTP SDK
Tracing.addEventAttribute( { eventId, name: 'cost:http', value: cost } );
```

The trace engine stops naming attributes. It just matches the registered prefix and forwards:

```js
// trace_engine.js
if ( action === EventAction.ADD_ATTR ) {
  const { name, value } = options.details;
  if ( matchPrefix( name ) ) {
    workflowHandle.signal( SIGNAL_SUMMABLE_EVENT, {
      name, value,
      activityId: parentId,
      activityName: parentName,
      date: Date.now()
    } );
  }
}
```

And `workflow.js` becomes one signal handler that buckets by event name, plus a builder at completion that groups by prefix and subcategory:

```js
const eventsByName = new Map();
setHandler( defineSignal( SIGNAL_SUMMABLE_EVENT ), e => {
  if ( !eventsByName.has( e.name ) ) eventsByName.set( e.name, [] );
  eventsByName.get( e.name ).push( e );
} );
…
return { output, ...traceObject, attributes: buildAttributes( eventsByName ) };
```

Where `buildAttributes` walks the registry, filters events by prefix, runs the aggregate per subcategory and at the top, and returns `{ cost: { total, components }, token_usage: { total, components }, … }`.

### What this buys us

- Adding `token_usage` (or anything else sumable) is one `registerSummableAttribute` call colocated with the producer SDK that owns it, plus producers using `prefix:subcategory` event names. **Zero changes to trace engine, workflow runner, or temporal_client.**
- The components is implicit in the event-id namespace, not encoded by the framework. New sub-source under an existing concept (someone later adds `cost:db`)? The producer just emits `cost:db` and it shows up under `components.db`. Nobody else has to know.
- The cost shape stops needing to carry token counts. The `components[i].tokens` and `info.tokens` smuggling goes away because tokens land cleanly under `attributes.token_usage`.
- `Attribute.COST` as a hardcoded framework constant disappears. It becomes documentation on the SDK that registers it.

WDYT? Totally open to "this is fine for now, we'll rework once tokens land" if the timeline doesn't allow. Just want it written down before the cost shape becomes API surface.

sdk/core/src/interface/aggregations.js
```diff
@@ -0,0 +1,23 @@
+import { Attribute } from '#trace_attribute';
+import Decimal from 'decimal.js';
+
+export const aggregateAttributes = attributes => ( {
+  cost: {
+    total: attributes
+      .filter( a => [ Attribute.HTTPRequestCost.TYPE, Attribute.LLMUsage.TYPE ].includes( a.type ) )
+      .reduce( ( sum, a ) => Decimal( a.total ).add( sum ).toNumber(), 0 )
```
catsby: I think we might be losing precision when we convert to `Decimal()` then then back to native floats with `.toNumber()` on every iteration of `reduce` here. I think this keeps it though:

```js
.reduce( ( sum, a ) => sum.add( a.total ), new Decimal( 0 ) ).toNumber()
```

Just call `.toNumber()` once at the end

sdk/llm/src/cost/index.js
```diff
@@ -14,29 +11,41 @@ const calcCost = ( tokens, ppm ) => Decimal( tokens ?? 0 ).div( M ).mul( ppm ).t
 export const calculateLLMCallCost = async ( { modelId, usage } ) => {
   try {
     const models = await fetchModelsPricing();
+
     if ( !models ) {
-      return { total: null, message: 'Failed to fetch models pricing' };
+      console.warn( 'Failed to fetch models pricing' );
```
catsby, line 16: nit: we should probably use the standard `logger` here, but I know other parts of sdk/llm just use `console` so we don't have to do that now

sdk/core/src/interface/aggregations.js
szanata: Good call.

sdk/llm/src/cost/index.js
szanata: I agree. But I would prefer to do in another PR and add logger to all libs, not just llm.

sdk/core/src/tracing/tools/build_trace_tree.spec.js
```diff
@@ -26,34 +27,66 @@ this can indicate it timed out or was interrupted.>>' );
...
     expect( result ).not.toBeNull();
-    expect( result.children[0].attributes ).toEqual( { latency_ms: 42, retries: 1 } );
+    expect( result.children[0].attributes ).toEqual( {
+      [Attribute.HTTPRequestCount.TYPE]: requestCount,
+      [Attribute.HTTPRequestCost.TYPE]: requestCost
+    } );
   } );
 
-  it( 'add_attr action overwrites prior value for the same attribute name', () => {
+  it( 'add_attr action overwrites prior value for the same attribute type', () => {
```
catsby, line 57: ❓ why do we want overwrites? If an activity has 2 llm calls or 2 http requests, all with associated charges, wont we overwrite the first?

sdk/core/src/tracing/tools/build_trace_tree.js
```diff
@@ -62,7 +62,7 @@ export default entries => {
     if ( action === EventAction.START ) {
       Object.assign( node, { input: details, startedAt: timestamp, kind, name } );
     } else if ( action === EventAction.ADD_ATTR ) {
-      node.attributes[details.name] = details.value;
+      node.attributes[details.type] = details;
```
catsby, line 65: I asked about this in the spec, but we're overwriting here when I think we should be accumulating? 

```
node.attributes[details.type] ??= [];
node.attributes[details.type].push( details );
```

Unless I'm mistaken here, which wouldn't be the first time :D

sdk/core/src/interface/workflow.js
```diff
@@ -69,7 +74,9 @@ export function workflow( { name, description, inputSchema, outputSchema, fn, op
 
     // Run the internal activity to retrieve the workflow trace destinations (only for root workflows, not nested)
     const traceDestinations = isRoot ? ( await steps[ACTIVITY_GET_TRACE_DESTINATIONS]( executionContext ) ) : null;
-    const traceObject = { trace: { destinations: traceDestinations } };
+
+    const attributes = [];
+    setHandler( defineSignal( 'add_attribute' ), e => attributes.push( e ) );
```
catsby: nit: should we make this a constant, maybe somewhere in [consts/index.js](https://github.com/growthxai/output/blob/fb7438aacee1406ac409ad9ce252b891bd5c9187/sdk/core/src/consts.js)?

sdk/llm/src/cost/index.js
```diff
@@ -14,29 +11,41 @@ const calcCost = ( tokens, ppm ) => Decimal( tokens ?? 0 ).div( M ).mul( ppm ).t
...
+    // When there aren't reasoning costs, the providers doesn't differentiate reasoning vs output, so the price is included in the output
+    if ( Number.isFinite( pricing.reasoning ) && Number.isFinite( reasoningTokens ) ) {
+      llmUsage.addUsage( { type: 'reasoning', ppm: pricing.reasoning, amount: reasoningTokens } );
+    }
+
+    return llmUsage;
   } catch ( error ) {
     console.error( 'Error calculating LLM call costs', error );
-    return { total: null, message: `Error calculating LLM call costs: ${error.constructor.name} - ${error.message}` };
+    return null;
```
catsby, line 49: previously we returned null and an error message, now we're just returning null and logging the error, this seems like a bigger change in behavior. I didn't check all the callers of this function though, does it change behavior upstream?

catsby (changes requested):
Unless I'm mistaken, we're overwriting costs when they should be accumulated. I'd like to understand that before approving. The rest is probably minor

sdk/core/src/tracing/tools/build_trace_tree.js
szanata: Each event generates a new trace tree node, so one HTTP call is one node, one LLM call is another node.

So it does not make sense for a single node to have more than one attribute of any kind.

sdk/llm/src/cost/index.js
szanata: This was not being used anywhere else. I will add a note to this in the migration guide.

sdk/core/src/tracing/tools/build_trace_tree.js
catsby: 👍 I thought this was scoped to the activity itself, my mistake

sdk/core/src/tracing/tools/build_trace_tree.spec.js
catsby: I was mistaken here, resolving

catsby (approved):

PR #207, opened 2026-05-14
feat(core): update hooks to include run ID and HTTP request events

bnchrch:
## Overview

Framework side of the COS-216 hook-only usage tracking effort. Two additive changes so consumers can subscribe to every cost + HTTP event with full identity context: `emitEvent` now forwards `runId` (and `workflowId` / `activityId`) from `executionContext` on every payload, and `@outputai/http` emits a new `http:request` event from `sdk/http` on every call — covering `success`, `http_error`, and `network_error`. No public API changes, no behavior changes for existing consumers — just richer payloads and one new event name.

## Before/After merge

- **Before:** os-workflows#155 is pinned to the dev release `@outputai/core@0.4.1-dev.c0b98d8.0` + `@outputai/http@0.4.1-dev.c0b98d8.0` (both confirmed live on npm under the `dev` dist-tag). That dev release is what atlas#1947 consumes via the os-workflows webhooks today.
- **After:** Once this merges and the standard minor release goes out (`@outputai/core@0.5.0` + `@outputai/http@0.5.0`, per the changesets in this PR), bump os-workflows#155 off the dev tag onto the stable minors and re-pin before that PR lands. Atlas needs nothing — it only depends on the os-workflows webhook contract, which doesn't change.

## Related PRs

- atlas: growthxai/atlas#1947
- os-workflows: growthxai/os-workflows#155


ops/api.Dockerfile
bnchrch, line 1: Lets revert this file. We dont need these changes

docs/COS-216-HOOK-ONLY.md
bnchrch, line 1: Remove this file

sdk/http/src/fetch/logger.ts
```diff
@@ -1,8 +1,22 @@
...
+type HttpRequestOutcome = 'success' | 'http_error' | 'network_error';
+
+/** Single source of truth for the `http:request` event shape. */
+const emitHttpRequestEvent = ( payload: {
+  requestId: string,
+  method: string,
+  url: string,
+  status: number | undefined,
+  durationMs: number,
+  outcome: HttpRequestOutcome
```
szanata, line 15: I think the proper terms for outcome would be `success`, `error` and `failure`.

error is status >= 300, and failure is network problems, etc.

sdk/http/src/fetch/index.ts
```diff
@@ -43,22 +44,28 @@ export const fetch = async ( input: RequestInfo | Request, init?: RequestInit )
...
 
+  const method = request.method;
+  const url = request.url;
+  const startedAt = performance.now();
+
   await logRequest( { requestId, request } );
 
   try {
     const response = await undici.fetch( request );
+    const durationMs = Math.round( performance.now() - startedAt );
```
szanata: performance already returns in milliseconds, so there is no need to round.

szanata (commented):
I left some comments. It is also necessary to update docs to reflect the new payload on exiting hooks and the new hook.

catsby (approved):

PR #211, opened 2026-05-19
refactor(TUI): Componentizing, uniformizing look and feel, improving space usage and responsiveness

szanata:
## Summary

Refactoring TUI to componentize it, and make its looks more uniform, besides improving space usage and responsiveness.

### Main changes:
- Refactored the dev TUI shell around reusable chrome primitives instead of one-off layout pieces.
- Removed the old divider/status-icon components and replaced them with reusable `TabBar`, `ModalFrame`, `ContentTitle`, `InlineSnippet`, `WorkflowStatusIcon`, `DockerServiceStatusIcon`, and `RunInfoSidebar` components.
- Simplified layout spacing by removing several ad hoc margins/rules and leaning on Ink `gap`, borders, and shared chrome components for consistent spacing.
- Centralized overlay routing in `DevApp`, including expanded JSON, run modal, and the new steps modal.
- Converted the old run detail view into a modal-style steps view with shared modal chrome, step navigation, tabbed step content, and bounded content rendering.
- Added responsive terminal layout support, including a minimum terminal size guard and dynamic available-height calculation for panel content.
- Made service logs, recent run details, and step details use the available vertical space and truncate content consistently on resize.
- Standardized layout height ownership by exposing `getHeight()` or `useHeight()` from the components that own rendered rows.
- Reworked `MasterDetailPanel` so panels can pass available height and receive the exact detail row budget for their content.
- Reworked footer rendering into two rows: global shortcuts + item count on the first row, local panel shortcuts + CLI version on the second row.
- Normalized footer hints so global shortcuts are owned by the footer and panel hints only describe active panel behavior.
- Cleaned up hotkey behavior and hints so footer/modal shortcuts stay aligned with actual input handlers.
- Simplified Help by removing duplicated hotkey documentation and keeping active shortcuts in the footer as the source of truth.
- Made the search bar persist while a query is active, even when changing tabs.
- Replaced local service banners with the shared toast system and kept pause state visible inside the log pane.
- Added a shared toast display and updated services actions/restarts/failure notifications to use it.
- Improved JSON rendering/editor behavior, including bounded overflow footers and cleaner active-line styling in the editor.
- Added shared panel helpers for selection, windowing, title formatting, JSON-value checks, cycling tabs, clamping, and small string/date utilities.
- Split runs state into separate run-list and run-step pane tabs, reducing overloaded UI state.
- Removed stale constants and dead code left behind by the dynamic-height refactor.
- Removed unneeded AWS env passthrough from the dev docker-compose service configuration.
- Added focused Vitest coverage for exported dev UI helpers, visible list builders, layout height helpers, JSON rendering, JSON editing, run detail extraction, and panel utilities.
- Fixing CLI to read `OUTPUT_AWS_*` instead of `AWS_*` env vars for the API.

#### Workflows panel
Screenshot 2026-05-19 at 23 34 48

#### Recent runs
Screenshot 2026-05-19 at 23 35 05

#### Recent runs - attributes
Screenshot 2026-05-19 at 23 35 31

#### Recent runs - view json modal
Screenshot 2026-05-19 at 23 35 40

#### Recent runs - view steps modal
Screenshot 2026-05-19 at 23 35 55

#### Services
Screenshot 2026-05-19 at 23 36 20

#### Help screens
Screenshot 2026-05-19 at 23 36 26
Screenshot 2026-05-19 at 23 36 43

#### Minimum size (100x40)
Screenshot 2026-05-19 at 23 37 06

#### Unsupported size guard
Screenshot 2026-05-19 at 23 37 33

## Test plan

[x] manual
[x] unit tests


sdk/cli/src/assets/docker/docker-compose-dev.yml
```diff
@@ -88,9 +88,6 @@ services:
       - NODE_ENV=development
       - OUTPUT_API_PORT=3001
       - OUTPUT_CATALOG_ID=${OUTPUT_CATALOG_ID:-main}
-      - OUTPUT_AWS_REGION=${AWS_REGION:-us-west-1}
```
szanata, line 91: The API expects `OUTPUT_AWS_*` env vars, but when running locally, the CLI was loading `AWS_*` env vars for it, even thou it is using `OUTPUT_AWS_*` for the worker.

This change make this consistent between services and also align it with the docs, that only mention `OUTPUT_AWS_*`.

sdk/cli/src/views/dev/dev_app.tsx
```diff
@@ -54,7 +109,7 @@ const useGlobalInput = ( opts: {
       return;
     }
 
-    if ( ui.search.open || ui.runModal.open || ui.expandedJson.open ) {
+    if ( ui.search.open || ui.runModal.open || ui.expandedJson.open || ui.runsView === 'detail' ) {
```
bnchrch: The overlay-is-open check is in three spots and they have to stay in sync:

- `useGlobalInput` early-returns when `ui.search.open || ui.runModal.open || ui.expandedJson.open || ui.runsView === 'detail'`
- `overlayFor` renders an overlay under the same conditions, expressed positionally
- `StepsModal`'s `useInput` re-checks `ui.tab === 'runs' && ui.runsView === 'detail' && !ui.expandedJson.open`

Add a fourth overlay and you've got three edits to remember. A single derived `hasOverlay` on `ui` (or in this file) that all three paths consume would mean the next overlay only touches the registry. Worth doing now while the surface is small.

sdk/cli/src/views/dev/modals/run_modal.tsx
```diff
@@ -227,7 +238,7 @@ export const RunModal: React.FC<{ workflowName: string; workflowPath?: string }>
```
bnchrch, line 238: Nit, take or leave: the `mode` union plus the shape of this `useInput` body is already a state machine. A small dispatch finishes the move and gets you exhaustiveness for free if `Mode` grows:

```tsx
const inputHandlers: Record<Mode, ( input: string, key: Key ) => void> = {
  select:       handleSelectInput,
  edit_name:    handleEditNameInput,
  error:        handleErrorInput,
  edit_content: () => {},
  submitting:   () => {}
};
useInput( ( i, k ) => inputHandlers[mode]( i, k ) );
```

sdk/cli/src/views/dev/components/content_title.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { Box, Text } from 'ink';
+
+export const getHeight = (): number => 2;
```
bnchrch: `export const getHeight = (): number => 2` is decoupled from the JSX that produces it. Bump `marginBottom` to 2, add a line of text, and the height claim is silently wrong while the layout math downstream still trusts it.

The pattern repeats across chrome components. Two ways out: derive the height from the same layout constants the JSX uses, or measure once and assert. Either way, named constants shared between the export and the render block keep them honest. Not blocking, flagging because the drift won't show up in a test, it'll show up as a one-row layout glitch on a future tweak.

sdk/cli/src/views/dev/dev_app.tsx
```diff
@@ -36,6 +44,53 @@ const TAB_NUMBER_KEYS: Record<string, Tab> = {
   4: 'help'
 };
 
+const getContentRows = ( opts: {
```
bnchrch, line 47: Every chrome row has to be remembered in this subtraction list, and forgetting one is "off by N rows" debugging.

```tsx
const CHROME = [ headerH, tabBarH, searchH, toastsH, footerH ];
const contentRows = Math.max( 1, opts.rows - CHROME.reduce( ( s, h ) => s + h, 0 ) );
```

Then "did I forget one?" is a `.length` check. Cheaper than the alternative when a future row lands.

sdk/cli/src/views/dev/dev_app.tsx
```diff
@@ -169,35 +271,60 @@ const Shell: React.FC<{
...
+    visibleWorkflowCount: visibleWorkflows.length,
+    runCount: runs.length,
+    visibleRunCount: visibleRuns.length,
+    serviceCount: services.length,
+    phase
+  } );
+  const overlay = overlayFor( { ui, detailRun, rows } );
+
+  useEffect( () => {
+    if ( ui.runsView === 'detail' && !detailRun ) {
```
bnchrch:
```tsx
useEffect( () => {
  if ( ui.runsView === 'detail' && !detailRun ) {
    setRunsView( 'list' );
  }
}, [ detailRun, setRunsView, ui.runsView ] );
```

This plus the `?? visibleRuns[0]` fallback at line ~256 are the same recovery dance. Could `runsView` just be derived?

```tsx
const effectiveRunsView = ui.runsView === 'detail' && detailRun ? 'detail' : 'list';
```

No effect, no render-then-correct race. Genuine question, is the persisted `detail` mode load-bearing when the run goes missing?

bnchrch (commented):
Big-picture reads really well. Chrome/components/modals split is clear, primitives are right-sized, `MasterDetailPanel` is the right abstraction. A handful of structural things worth raising before merge, mostly around overlay coupling and the height-ownership pattern. None of them block.

sdk/cli/src/views/dev/utils/panel_helpers.ts
```diff
@@ -37,3 +55,37 @@ export const computeWindowStart = (
...
+): {
+  selectedIndex: number;
+  setSelectedIndex: Dispatch<SetStateAction<number>>;
+  selectPrevious: () => void;
+  selectNext: () => void;
+} => {
+  const [ rawIndex, setSelectedIndex ] = useState( initialIndex );
+  const selectedIndex = clampIndex( rawIndex, count );
+
+  useEffect( () => {
```
bnchrch: Every consumer reads `selectedIndex` (the clamped value); the effect just writes it back into `rawIndex`. If nobody depends on `rawIndex` being authoritative, the effect can go, derived is already correct and the writeback costs a render. What's the case the effect covers?

sdk/cli/src/assets/docker/docker-compose-dev.yml
```diff
@@ -88,9 +88,6 @@ services:
...
-      - OUTPUT_AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-}
-      - OUTPUT_AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY:-}
       - TEMPORAL_ADDRESS=temporal:7233
```
bnchrch, line 91: PR description has two lines I'm trying to reconcile:
> - Removed unneeded AWS env passthrough from the dev docker-compose service configuration.
> - Fixing CLI to read `OUTPUT_AWS_*` instead of `AWS_*` env vars for the API.

After this, does a dev with `AWS_*` set in their shell still get creds to the API container, or do they have to switch to `OUTPUT_AWS_*`? If the latter, worth a changeset line so nobody silently loses creds on pull.

bnchrch (approved):
Approving. Comments above are non-blocking — happy to land as-is or wait for any of them, your call.

sdk/cli/src/views/dev/dev_app.tsx
szanata: Good callout. I’m going to leave this as-is for now.

I agree a named helper/registry would make sense once we add another overlay, but at the current size it would mostly add a layer without removing much code. Also, search.open is part of the same “block global input” condition but isn’t really an overlay, so I’d want to name/model this carefully rather than introduce an overlay abstraction that is slightly inaccurate.

If this grows, I’d probably extract a small blocksGlobalInput(ui) helper first, before moving to a fuller overlay registry.

sdk/cli/src/views/dev/modals/run_modal.tsx
szanata: Agreed that this is effectively a small state machine. I’m going to leave it inline for now because the mode handlers are still short and keeping the logic in one place is easier to scan at this size.
If this grows another mode or the existing mode branches get more behavior, I’d move it to a `Record<Mode, handler>` dispatch as suggested so we get exhaustiveness and cleaner mode boundaries.

sdk/cli/src/views/dev/components/content_title.tsx
szanata: You are right. I improved the code, so it is less disconnected.

sdk/cli/src/views/dev/dev_app.tsx
szanata: I’m going to leave this as-is for now. The current list is short, explicit, and each height already comes from the component-owned helper, so I don’t think the array/reduce version buys us much yet. If this grows beyond the current chrome pieces, I agree this would be a good cleanup.

sdk/cli/src/views/dev/utils/panel_helpers.ts
szanata: Good point. Consumers only use the derived selectedIndex, so writing the clamped value back into rawIndex is redundant and adds an extra render when the list size changes. I’ll remove the effect and keep clamping as a derived value.

sdk/cli/src/views/dev/dev_app.tsx
szanata: It simplified this and removed the correcting.

sdk/cli/src/assets/docker/docker-compose-dev.yml
szanata: Correct, after this change local Docker Compose users need to provide the OUTPUT_AWS_* variables if they want the API container to read remote traces from S3. That matches production and the documented env names; the previous AWS_* passthrough was accidental and inconsistent with the worker.

In practice this only affects local dev setups using remote trace fetching, which should be uncommon since the dev compose path primarily uses local traces. I don’t think we need compatibility for the mistaken AWS_* names, but I will mention this in the changeset.

PR #214, opened 2026-05-20
feat(core): stamp eventId on every framework emit (OUT-456)

bnchrch:
## Problem

Framework events carry no stable per-emit identity, so consumers (webhook retries, ClickHouse dedup, audit logs) synthesize one — usually badly. `requestId` isn't safe: `cost:http:request` and `http:request` for the same fetch share it.

## Solution

- **Bus layer** — [bus.js](sdk/core/src/bus.js) wraps `messageBus.emit` to stamp a UUID v4 `eventId` on every object payload. Covers `emitEvent` events *and* lifecycle hook payloads; future emit sites get it for free
- **Hooks** — `eventId` forwarded through [hooks/index.js](sdk/core/src/hooks/index.js); 5 hook payload TS types declare it
- Callers can pre-set `eventId` for deterministic retry scenarios
- Docs updated: `cost-events`, `packages/http`, `packages/core`, `migrations/v0.4.0-to-v0.5.0`

Closes [OUT-456](https://linear.app/growthx/issue/OUT-456)

## Test plan

- [x] lint + build clean; `sdk/core` 593, `sdk/llm` 216, `sdk/http` 71 — all green
- [ ] CI green

### DIFF
--- a/sdk/core/src/bus.js
+++ b/sdk/core/src/bus.js
@@ -1,3 +1,30 @@
 import { EventEmitter } from 'node:events';
+import { randomUUID } from 'node:crypto';
 
-export const messageBus = new EventEmitter();
+const emitter = new EventEmitter();
+const originalEmit = emitter.emit.bind( emitter );
+
+/**
+ * Every object payload emitted through `messageBus` is stamped with a UUID v4
+ * `eventId` — a stable per-emit idempotency key for downstream consumers
+ * (webhook retry handling, ClickHouse `ReplacingMergeTree` dedup, audit logs,
+ * etc.). Stamping happens at the bus layer, so any emit site — `emitEvent`,
+ * lifecycle sinks, activity interceptor, future emitters — gets the field for
+ * free.
+ *
+ * Callers may pre-set `eventId` on the payload to override the generated one
+ * (useful for deterministic retry scenarios). Non-object payloads (primitives,
+ * null, undefined, arrays) pass through unchanged.
+ */
+emitter.emit = ( event, ...args ) => {
+  if ( args.length === 0 ) {
+    return originalEmit( event );
+  }
+  const [ payload, ...rest ] = args;
+  if ( payload && typeof payload === 'object' && !Array.isArray( payload ) ) {
+    return originalEmit( event, { ...payload, eventId: payload.eventId ?? randomUUID() }, ...rest );
+  }
+  return originalEmit( event, payload, ...rest );
+};
+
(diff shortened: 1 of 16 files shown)


szanata (approved):
I think we should stamp this on all hooks. But I am ok on ruling this just these at first.

However, this needs to be mentioned in the docs as all event payloads are mentioned there.

sdk/core/src/bus.js
```diff
@@ -1,3 +1,36 @@
 import { EventEmitter } from 'node:events';
+import { randomUUID } from 'node:crypto';
 
-export const messageBus = new EventEmitter();
+const emitter = new EventEmitter();
+const originalEmit = emitter.emit.bind( emitter );
+
+const isObject = value => value !== null && typeof value === 'object' && !Array.isArray( value );
```
szanata: We already have an utils function for that:
`isPlainObject` from `#utils`.

sdk/core/src/bus.js
```diff
@@ -1,3 +1,36 @@
...
+ * null, undefined, arrays) pass through unchanged.
+ */
+emitter.emit = ( event, ...args ) => {
+  if ( args.length === 0 ) {
+    return originalEmit( event );
+  }
+
+  const [ payload, ...rest ] = args;
+  if ( !isObject( payload ) ) {
+    return originalEmit( event, payload, ...rest );
```
szanata: I would still do `return originalEmit( event, ...args )` in this case.

sdk/core/src/bus.js
```diff
@@ -1,3 +1,36 @@
...
+ * `eventId` — a stable per-emit idempotency key for downstream consumers
+ * (webhook retry handling, ClickHouse `ReplacingMergeTree` dedup, audit logs,
+ * etc.). Stamping happens at the bus layer, so any emit site — `emitEvent`,
+ * lifecycle sinks, activity interceptor, future emitters — gets the field for
+ * free.
+ *
+ * Callers may pre-set `eventId` on the payload to override the generated one
+ * (useful for deterministic retry scenarios). Non-object payloads (primitives,
+ * null, undefined, arrays) pass through unchanged.
+ */
```
szanata, line 12: I think just `"Every object payload emitted through `messageBus` is stamped with a UUID v4 eventId"`. All the rest her just state the obvious, and polute it a bit.

szanata (commented):
I left some comments suggesting small improvements. Overall this is good.

PR #219, opened 2026-05-22
refactor(core): Improving the reliability of the activity <-> workflo…

szanata:
## Summary

Improving the reliability of the activity <-> workflow communication via Temporal signals and making it less prone to break the whole worker on runtime failures:

### Reduce chance of losing signals
All signal promises are added to an array, all promises are awaited up to 30s before finishing the activity. This ensures that in-flight signals will be handled properly. The timeout of 30s is very generous as this is a race condition that normally takes a few instants to resolve.

### Preventing signals failures from breaking the worker
All `.signals()` call now have a `.catch()` to handle and log possible failures. These failures are common runtime occurrences, normally when a signal is in-flight but the workflow already finished executing.

### Better separation of concern
All signal related code now lives exclusively in the activity interceptor, proving the trace engine with just a callback to send the attributes when they arrive. 

## Test plan

- [x] Added new unit tests and refactored old ones, all passing;
- [x] All e2e test passed;
- [x] Manually tested against local Temporal;
- [x] Manually tested against remote Temporal;

### DIFF
--- a/sdk/core/src/worker/interceptors/activity.js
+++ b/sdk/core/src/worker/interceptors/activity.js
@@ -13,18 +13,6 @@ const log = createChildLogger( 'ActivityInterceptor' );
 
 const IN_FLIGHT_SIGNALS_TIMEOUT_MS = 30_000;
 
-const flushSignals = async signals => {
-  try {
-    await allSettledWithTimeout( signals, IN_FLIGHT_SIGNALS_TIMEOUT_MS );
-  } catch ( error ) {
-    if ( error.isTimeout ) {
-      log.warn( 'Some usage/cost attributes were missed because not all activity signals were sent to the workflow' );
-    } else {
-      throw error;
-    }
-  }
-};
-
 /*
   This interceptor wraps every activity execution with cross-cutting concerns:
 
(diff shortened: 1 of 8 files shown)


sdk/core/src/worker/interceptors/activity.js
```diff
@@ -36,46 +42,95 @@ export class ActivityExecutionInterceptor {
...
+  /**
+   * Returns a workflow entry by its name or throws error
+   * @param {string} workflowName
+   * @returns {object} Workflow entry
+   * @throws {Error}
+   */
+  getWorkflowEntry( workflowName ) {
+    const workflowEntry = this.workflowsMap.get( workflowName );
+    if ( !workflowEntry ) {
+      throw new Error( `Activity interceptor: workflow "${workflowName}" not found in workflowsMap.` );
```
catsby, line 54: previously if a workflow name wasn't found, we also returned a list of available workflows, but we're not doing that here. Was that intentional? See the old line 55 below for what I mean

catsby (approved):
One observation/question otherwise ✅

sdk/core/src/worker/interceptors/activity.js
szanata: Yes it was intentional:
1. That might make this error string very long and pollute it during debug.
2. This error is very unlikely to happen in reality, it is there because technically when reading from the map we might not find a value, but in the real world, all values will be there.

PR #222, opened 2026-05-22
feat(cli): support side-by-side dev stacks

bnchrch:
## Problem

The CLI-shipped docker compose template exposes parameterized host ports for `api` (`OUTPUT_API_HOST_PORT`) and `temporal-ui` (`OUTPUT_TEMPORAL_UI_HOST_PORT`), but the `temporal` service has no `ports:` block at all in `0.5.x`. The old `OUTPUT_TEMPORAL_PORT` knob was removed in #171 and never replaced.

## Solution
image

- **Compose** — restore a parameterized port on the `temporal` service: `'${OUTPUT_TEMPORAL_HOST_PORT:-7233}:7233'`. Default unchanged; opt-in for multi-stack.
- **CLI** — plumb `OUTPUT_TEMPORAL_HOST_PORT` through `config.ports` using the existing `parsePort` helper, and surface it in the `output dev` description, the project `.env.example.template`, and the `cli.mdx` env-var reference.
- **Docs** — extend the "Running multiple dev stacks" section in `cli.mdx` with the full five-var recipe (`DOCKER_SERVICE_NAME`, `OUTPUT_CATALOG_ID`, and the three `OUTPUT_*_HOST_PORT` knobs).
- **Collision hint** — new `port_collision.ts` util parses the three known docker bind-failure shapes, maps the colliding host port back to the env var that owns it (preferring resolved-port lookup so user overrides resolve correctly), and prepends a two-line hint to the error before Ink unmounts. Unknown ports fall through to a generic "stop the process holding it" suggestion.


derencius (approved):

sdk/cli/src/utils/port_availability.ts
```diff
@@ -0,0 +1,41 @@
+import net from 'node:net';
+
+/**
+ * Probe each port by attempting to bind a TCP server on `0.0.0.0`. Returns
+ * the first port whose listen call errors with `EADDRINUSE`, or `null` when
+ * every port is free.
+ *
```
catsby, line 5: this doc string seems to document the `findUnavailablePorts()` function below, and not the `isPortTake()` function it's next to

catsby (approved):
Nit on doc string otherwise ✅

sdk/cli/src/utils/port_availability.ts
bnchrch: @claude fix this

bnchrch:
@claude fix the ci lint failure

PR #224, opened 2026-05-25
chore(api): increase temporal max message size

bnchrch:
## Problem

Workflows whose result envelope exceeds 4 MiB fail to surface their output via the API:

```
GET /workflow/<wid>/runs/<rid>/result
→ 500 { "message": "8 RESOURCE_EXHAUSTED: Received message larger than max (4296236 vs 4194304)" }
```

4,194,304 bytes is gRPC's default `max_recv_msg_size`. The API server's `Connection.connect()` never passed `channelArgs`, so it inherited that cap. Atlas's cost-rollup pipeline (Agent HQ) silently swallows the 500, leaving completed runs with `total_cost_micro_usd=0`, `total_tokens=0`, and zero `Analytics::WorkflowEvent` rows.

## Solution

- **API** — `Connection.connect()` in `api/src/clients/temporal_client.js` now passes `channelArgs` with `grpc.max_receive_message_length` and `grpc.max_send_message_length`
- **Config** — new `TEMPORAL_GRPC_MAX_MESSAGE_SIZE_BYTES` env var in `api/src/configs.js`, default `33554432` (32 MiB); uses `z.preprocess(coalesceEmptyString, ...)` so empty strings fall back to the default
- **Docs** — env var documented in `docs/guides/api/configuration.mdx`

## Test plan

- [x] `npm run lint` passes
- [x] `npm run build:packages` passes
- [x] `npm test` — 1900 tests pass, including new coverage in `temporal_client.spec.js` and `configs.spec.js` (default, override, and empty-string fallback)
- [ ] Manual smoke: trigger a workflow whose result exceeds 4 MiB against a locally-rebuilt `outputai/api:dev` and confirm `GET /workflow/:id/runs/:rid/result` returns 200
- [ ] Atlas-side: re-run `OutputWorkflows::Client#workflow_result` on a previously-failing Competitor Researcher run and confirm the cost rollup populates

## Notes for reviewers

- **Worker side intentionally unchanged** — `NativeConnection.connect()` does not accept `channelArgs`; `NativeConnectionOptions` only exposes `{ address, tls, proxy, metadata, apiKey, disableErrorCodeMetricTags, plugins }` and the Rust core's `ClientOptions` has no gRPC tuning fields. The first revision of this PR had a worker change that was a silent no-op — reverted in `c093c39d`. If we ever need to raise gRPC limits on the worker side, that has to happen via Temporal Server config, not the SDK.

## Follow-up (out of scope)

- Slim variant of `/runs/:rid/result` (returning only `aggregations` + `attributes`) — would only trim HTTP egress, not the underlying gRPC envelope, so it doesn't address the root cause; deferred.
- Atlas's `WorkflowEndProcessor` currently swallows 5xx silently — surfacing it as a Sentry event is tracked separately.

### DIFF
--- a/api/src/configs.js
+++ b/api/src/configs.js
@@ -11,6 +11,9 @@ const envSchema = z.object( {
   TEMPORAL_WORKFLOW_EXECUTION_TIMEOUT: z.string().optional().default( '24h' ),
   TEMPORAL_WORKFLOW_EXECUTION_MAX_WAITING: z.coerce.number().optional().default( 300_000 ), // 5minutes
   TEMPORAL_API_KEY: z.string().optional(),
+  // gRPC max message size for the Temporal client connection (send + receive).
+  // gRPC's default receive cap is 4 MiB; some workflow result envelopes exceed it.
+  TEMPORAL_GRPC_MAX_MESSAGE_SIZE_BYTES: z.coerce.number().int().positive().optional().default( 32 * 1024 * 1024 ),
   OUTPUT_API_PORT: z.coerce.number().optional().default( 3000 ),
   OUTPUT_API_SERVICE_NAME: z.string().optional().default( 'output-api' ),
   OUTPUT_CATALOG_ID: z.string().regex( /^[a-z0-9_.@-]+$/i ),
@@ -34,7 +37,8 @@ export const temporal = {
   apiKey: safeEnvVar.TEMPORAL_API_KEY,
   namespace: safeEnvVar.TEMPORAL_NAMESPACE,
   workflowExecutionTimeout: safeEnvVar.TEMPORAL_WORKFLOW_EXECUTION_TIMEOUT,
-  workflowExecutionMaxWaiting: safeEnvVar.TEMPORAL_WORKFLOW_EXECUTION_MAX_WAITING
+  workflowExecutionMaxWaiting: safeEnvVar.TEMPORAL_WORKFLOW_EXECUTION_MAX_WAITING,
+  grpcMaxMessageSizeBytes: safeEnvVar.TEMPORAL_GRPC_MAX_MESSAGE_SIZE_BYTES
 };
 
 export const api = {
--- a/api/src/configs.js
+++ b/api/src/configs.js
@@ -4,6 +4,8 @@ class InvalidEnvVarsErrors extends Error { }
 
 export const isProduction = process.env.NODE_ENV === 'production';
 
+const coalesceEmptyString = v => v === '' ? undefined : v;
+
 const envSchema = z.object( {
   NODE_ENV: z.string().optional().default( 'development' ),
   TEMPORAL_ADDRESS: z.string().optional().default( 'localhost:7233' ),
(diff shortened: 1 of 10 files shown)


sdk/core/src/worker/index.js
```diff
@@ -54,12 +55,23 @@ const callerDir = process.argv[2];
...
   }
-  const connection = await NativeConnection.connect( { address, tls: Boolean( apiKey ), apiKey, proxy } );
+  // channelArgs raises gRPC's 4 MiB default cap so the worker's Client (catalog + interceptor) can
+  // exchange larger payloads with Temporal (see COS-370).
+  const connection = await NativeConnection.connect( {
+    address,
+    tls: Boolean( apiKey ),
+    apiKey,
+    proxy,
+    channelArgs: {
```
catsby: I don't believe `NativeConnection` supports `channelArgs` in it's connection settings:

- `connect`: https://typescript.temporal.io/api/classes/worker.NativeConnection#connect
- `NativeConnectionOptions`: https://typescript.temporal.io/api/interfaces/worker.NativeConnectionOptions

sdk/core/src/worker/index.js
```diff
@@ -54,12 +55,23 @@ const callerDir = process.argv[2];
...
 
-  log.info( 'Connecting Temporal...' );
+  log.info( 'Connecting Temporal...', { grpcMaxMessageSizeBytes } );
   const proxy = grpcProxy ? { type: 'http-connect', targetHost: grpcProxy } : undefined;
   if ( proxy ) {
     log.info( 'Using gRPC proxy', { targetHost: grpcProxy } );
   }
-  const connection = await NativeConnection.connect( { address, tls: Boolean( apiKey ), apiKey, proxy } );
+  // channelArgs raises gRPC's 4 MiB default cap so the worker's Client (catalog + interceptor) can
+  // exchange larger payloads with Temporal (see COS-370).
```
catsby:
```suggestion
  // exchange larger payloads with Temporal.
```

nit: we shouldn't link or reference internal issue(s) or system(s)

api/src/configs.js
```diff
@@ -11,6 +11,9 @@ const envSchema = z.object( {
   TEMPORAL_WORKFLOW_EXECUTION_TIMEOUT: z.string().optional().default( '24h' ),
   TEMPORAL_WORKFLOW_EXECUTION_MAX_WAITING: z.coerce.number().optional().default( 300_000 ), // 5minutes
   TEMPORAL_API_KEY: z.string().optional(),
+  // gRPC max message size for the Temporal client connection (send + receive).
+  // gRPC's default receive cap is 4 MiB; some workflow result envelopes exceed it.
+  TEMPORAL_GRPC_MAX_MESSAGE_SIZE_BYTES: z.coerce.number().int().positive().optional().default( 32 * 1024 * 1024 ),
```
catsby: The worker config uses a different zod validation/definition: 

```
z.preprocess( coalesceEmptyString, z.coerce.number().int().positive().default( 32 * 1024 * 1024 ) )
```

the difference is that the worker will take an empty-string env var (`''`) and fall back to the default, where as this one will error. The config spec here doesn't test `''`

catsby (commented):
The zod validations are different across the api and worker, I think we should use the `coalesceEmptyString` version for both

sdk/core/src/worker/index.js
bnchrch: Good catch — verified against the installed typings. `NativeConnectionOptions` only exposes `{ address, tls, proxy, metadata, apiKey, disableErrorCodeMetricTags, plugins }`, and `toNativeClientOptions()` doesn't forward channelArgs to the Rust core's `ClientOptions` either. Worker change was a silent no-op (only "worked" because the vitest mock accepts any shape). Fully reverted in c093c39d — worker-side gRPC tuning has to come from Temporal Server config, not the SDK.

sdk/core/src/worker/index.js
bnchrch: Reworded in c093c39d — dropped the COS-370 reference from the surviving API comment too.

api/src/configs.js
bnchrch: Aligned with the worker idiom in c093c39d — inlined `coalesceEmptyString` in `api/src/configs.js` and switched to `z.preprocess(coalesceEmptyString, z.coerce.number().int().positive().default(...))`. Added an empty-string test in `configs.spec.js` confirming the fallback.

catsby (approved):

PR #225, opened 2026-05-27
fix(http): propagate requestId across Response.clone() (COS-489)

bnchrch:
## Overview
Make the `requestId` tag attached by `@outputai/http`'s custom fetch survive `Response.clone()`. Unblocks every service client that wants to emit cost from inside a `ky` `afterResponse` hook — required by all 7 client PRs in the COS-489 wave (Jina, Semrush, DataForSEO, Exa, Tavily, SerpAPI, Firecrawl).

## Problem / Solution
`ky` clones the response before passing it to `afterResponse` hooks (see `ky/distribution/core/Ky.js`: `const clonedResponse = ky.#decorateResponse(response.clone());`). The clone is a fresh `Response` object that doesn't carry the `requestIdSymbol` property `addRequestCost` reads to correlate the request — so any hook calling `addRequestCost(response, value)` warns "did not originate from @outputai/http" and silently drops the cost.

Tried a header fallback first (`response.headers.set('x-output-request-id', requestId)`), but undici Responses are immutable on the received side — `headers.set()` throws.

Patches `clone()` on the original response so each clone re-runs `addRequestIdToResponse`, propagating the tag through any depth of clones. Recursive — clones of clones inherit too.

## Notes for reviewers
- `configurable: true` on the symbol property (was `false`) — needed because the recursion calls `defineProperty` on already-tagged clones in pathological cases. No external API depends on it being locked.
- Verified E2E with the Jina hook on `os-workflows#222` against `context_brand_research`: workflow result includes `http:request:cost { total: 0.00006662 }` and `aggregations.cost.total` correctly sums Jina + LLM.

### DIFF
--- a/sdk/http/src/fetch/utils.spec.ts
+++ b/sdk/http/src/fetch/utils.spec.ts
@@ -315,5 +315,17 @@ describe( 'fetch/utils', () => {
         value: 'req-456'
       } );
     } );
+
+    it( 'propagates the request id to clones (ky afterResponse passes the clone)', () => {
+      const response = new Response( 'ok' );
+      addRequestIdToResponse( response, 'req-789' );
+
+      const cloned = response.clone();
+      expect( ( cloned as unknown as Record<symbol, unknown> )[requestIdSymbol] ).toBe( 'req-789' );
+
+      // recursion: clones of clones inherit too
+      const grandchild = cloned.clone();
+      expect( ( grandchild as unknown as Record<symbol, unknown> )[requestIdSymbol] ).toBe( 'req-789' );
+    } );
   } );
 } );
--- a/sdk/http/src/fetch/utils.spec.ts
+++ b/sdk/http/src/fetch/utils.spec.ts
@@ -293,12 +293,11 @@ describe( 'fetch/utils', () => {
   } );
 
   describe( 'addRequestIdToResponse', () => {
-    it( 'stores request id under requestIdSymbol and returns the same response', () => {
+    it( 'stores request id under requestIdSymbol on the response', () => {
       const response = new Response( 'ok' );
 
-      const enriched = addRequestIdToResponse( response, 'req-123' );
+      addRequestIdToResponse( response, 'req-123' );
 
-      expect( enriched ).toBe( response );
       expect( ( response as unknown as Record<symbol, unknown> )[requestIdSymbol] ).toBe( 'req-123' );
     } );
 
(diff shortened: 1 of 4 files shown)


sdk/http/src/fetch/utils.spec.ts
```diff
@@ -315,5 +315,17 @@ describe( 'fetch/utils', () => {
...
+      const response = new Response( 'ok' );
+      addRequestIdToResponse( response, 'req-789' );
+
+      const cloned = response.clone();
+      expect( ( cloned as unknown as Record<symbol, unknown> )[requestIdSymbol] ).toBe( 'req-789' );
+
+      // recursion: clones of clones inherit too
+      const grandchild = cloned.clone();
+      expect( ( grandchild as unknown as Record<symbol, unknown> )[requestIdSymbol] ).toBe( 'req-789' );
+    } );
```
catsby, line 328: do we need a test that demonstrates `addRequestCost()` usage as well? It's mentioned in the description but we don't test it here. I see it has some tests in `sdk/http/src/cost.spec.ts` but I'm not sure it's testing this fix end-to-end if that makes sense?

sdk/http/src/fetch/utils.ts
```diff
@@ -100,12 +100,23 @@ export const parseBody = async ( r : Request | Response ) : Promise<string | obj
...
- * @param requestId
+ * Tag a response with its request id so downstream code (e.g. addRequestCost)
+ * can correlate. Stores the id under a private symbol AND patches `clone()` so
+ * the tag propagates to clones — ky clones the response before invoking
+ * `afterResponse` hooks, and undici headers are immutable on received
+ * responses, so a symbol re-attached inside `clone()` is the only path that
+ * survives.
  */
-export const addRequestIdToResponse = ( response: Response, requestId: string ) =>
+export const addRequestIdToResponse = ( response: Response, requestId: string ) : Response => {
```
catsby: should users use this `Response` that's returned? It looks like this function mutates in-place, but the return to me makes it seem like I should use what I get back?

catsby (approved):

bnchrch:
Addressed both review comments in 40713508:

- **`utils.ts` signature** → tightened to `: void`. The function is a mutator; the return-the-input pattern was misleading. Verified the only external call site (`sdk/http/src/fetch/index.ts:58`) already discarded the return, so no consumer changes.
- **`addRequestCost` end-to-end test** → added in `cost.spec.ts` ("records cost on a cloned response (regression: ky afterResponse hooks)"). Wires `addRequestIdToResponse` → `response.clone()` → `addRequestCost(clone, value)` and asserts `Tracing.addEventAttribute` fires + no `console.warn`. The pre-existing `utils.spec.ts` "propagates the request id to clones" test stays — it pins the mechanism; the new test pins the contract.

Changeset updated to mention the signature tightening alongside the bug fix.

All 47 tests in `sdk/http` pass locally.

PR #229, opened 2026-05-28
refactor(core,cli,api): Improving workflow aggregations [OUT-465]

szanata:
## Summary

- Removed property `.attributes` from workflow result wrapper object: Workflows will no longer accumulate or expose attributes;
- Added `__output_workflow_wrapper_version=1` field on workflow wrapper object to better version it;
- Removed Signals-based communication between Activities and Workflows to share individual attributes:
  - Each activity now aggregates all attributes of the events that happened within it. This is returned in a new wrapper around the activity:
  ```js
  {
    __output_activity_wrapper_version: 1, // internal flag to indicate this wrapper's version
    output: ..., // the raw output from the activity
    aggregations: {  // aggregation object with total llm/http usage and cost from all requests of this activity
      cost: {
        total: 1 // total cost from all http and llm requests
      },
      tokens: { // breakdown of all llm tokens used
        total: 10,
        input: 3,
        cached_input: 1,
        output: 4,
        reasoning: 2
      },
      httpRequests: { // total number of http calls made
        total: 3
      }
    }
  }
  ```
  - Workflows now read these aggregations and merge them to create the final `.aggregations` object returned in its result, which is unchanged;
  - When Activities fail, a fallback Signal is sent with the aggregations so workflows can still compute them, avoiding data loss.
- Removed `.attributes` from API's workflow result (`/run` and `/result`);
- Removed "Attributes" tab from the "Recent Runs > [workflow view]" on the TUI;
- 
## Test plan

- [x] Manual
- [x] Unit tests
- [x] e2e


sdk/core/src/interface/workflow.js
```diff
@@ -78,21 +89,71 @@ export function workflow( { name, description, inputSchema, outputSchema, fn, op
...
+        };
+      } )() )
+    };
+
+    setHandler( defineSignal( Signal.SEND_AGGREGATIONS ), a => {
+      workflowResult.aggregations = mergeAggregations( workflowResult.aggregations, a );
+    } );
+
+    /**
+     * @IMPORTANT Keep support for deprecated add_attribute Signal to allow for Temporal replays.
```
bnchrch, line 122: 📖 🔮 Theres 4 copies of `@TODO This can be removed 30days after this release` in this file. 30 days from when though? Once this merges nothing tracks it and these shims tend to live forever. Could we (a) pin an actual date or version in the comment, and (b) drop a 🔮 follow-up issue so we actually delete the dual-path logic later?

sdk/core/src/interface/workflow.js
```diff
@@ -78,21 +89,71 @@ export function workflow( { name, description, inputSchema, outputSchema, fn, op
       activityOptions: memo.activityOptions ?? activityOptions // Also preserve the original activity options
     } );
 
-    // Run the internal activity to retrieve the workflow trace destinations (only for root workflows, not nested)
-    const traceDestinations = isRoot ? ( await steps[ACTIVITY_GET_TRACE_DESTINATIONS]( executionContext ) ) : null;
+    // Creates the result wrapper with information about the workflow
+    const workflowResult = {
+      [WORKFLOW_WRAPPER_VERSION_FIELD]: 1,
+      aggregations: aggregateAttributes( [] ),
+      ...( isRoot && await ( async () => {
```
bnchrch: 💅 The conditional-spread-of-an-async-IIFE took me a couple reads, and `...( isRoot && ... )` leans on `...false` spreading nothing. WDYT about a small named function?
```suggestion
      ...( isRoot ? await resolveRootTrace( steps, executionContext ) : {} )
```
where `resolveRootTrace` does the getTraceDestinations call + unwrap. Reads top to bottom.

sdk/core/src/interface/workflow.js
```diff
@@ -78,21 +89,71 @@ export function workflow( { name, description, inputSchema, outputSchema, fn, op
       activityOptions: memo.activityOptions ?? activityOptions // Also preserve the original activity options
     } );
 
-    // Run the internal activity to retrieve the workflow trace destinations (only for root workflows, not nested)
-    const traceDestinations = isRoot ? ( await steps[ACTIVITY_GET_TRACE_DESTINATIONS]( executionContext ) ) : null;
+    // Creates the result wrapper with information about the workflow
+    const workflowResult = {
+      [WORKFLOW_WRAPPER_VERSION_FIELD]: 1,
+      aggregations: aggregateAttributes( [] ),
```
bnchrch: 💅 `aggregateAttributes( [] )` as the way to spell "empty totals" is a bit indirect, and the specs all redefine `emptyAggregations` by hand. Could we export a named `emptyAggregations()` from `internal_utils/aggregations` and reuse it here and in the tests?

sdk/core/src/interface/workflow.js
```diff
@@ -116,33 +177,43 @@ export function workflow( { name, description, inputSchema, outputSchema, fn, op
                 ...( extra?.options?.activityOptions && { activityOptions: deepMerge( activityOptions, extra.options.activityOptions ) } )
               }
             } );
-            attributes.push( ...( result.attributes ?? [] ) );
+            /**
+             * @IMPORTANT Keep support for deprecated ".attributes" from workflow results to allow for Temporal replays.
+             * @TODO This can be removed 30days after this release
+             */
+            if ( result?.attributes ) {
+              workflowResult.aggregations = mergeAggregations( workflowResult.aggregations, aggregateAttributes( result.attributes ) );
```
bnchrch: 💅 `workflowResult.aggregations = mergeAggregations( workflowResult.aggregations, X )` shows up ~6 times. A tiny closure dries it up and reads nicer:
```js
const absorb = aggregations => {
  workflowResult.aggregations = mergeAggregations( workflowResult.aggregations, aggregations );
};
```
then `absorb( aggregations )`, `absorb( aggregateAttributes( [ attribute ] ) )`, etc.

sdk/core/src/interface/workflow.js
```diff
@@ -78,21 +89,71 @@ export function workflow( { name, description, inputSchema, outputSchema, fn, op
...
+         */
+        return {
+          trace: {
+            destinations: isActivityResultWrapped( result ) ? result.output : result
+          }
+        };
+      } )() )
+    };
+
+    setHandler( defineSignal( Signal.SEND_AGGREGATIONS ), a => {
```
bnchrch: 💅 tiny: `a` -> `aggregations` so the handler reads on its own.

sdk/core/src/interface/workflow.js
```diff
@@ -78,21 +89,71 @@ export function workflow( { name, description, inputSchema, outputSchema, fn, op
...
 
-    const attributes = [];
-    setHandler( defineSignal( Signal.ADD_ATTRIBUTE ), e => attributes.push( e ) );
+    /**
+     * unwraps step result to extract "aggregations" and return plain result.
+     * @param {Function} step
+     * @param  {...any} args
+     * @returns {any} The step "output"
+     */
+    const callStepFn = async ( step, ...args ) => {
```
bnchrch: 💅 📖 `callStepFn` does more than call, it unwraps + accumulates aggregations as a side effect. Maybe `invokeAndAbsorb` or `runStep`? Also the jsdoc says `@returns the step output` but thats only true on the wrapped branch, worth a word on the legacy passthrough.

sdk/core/src/utils/utils.js
```diff
@@ -193,6 +193,41 @@ export const deepMerge = ( a, b ) => {
...
+ * - Object "a" fields that don't exist on object "b" will be preserved;
+ *
+ * If "b" isn't an object, a new object equal to "a" is returned
+ *
+ * @param {object} a - The base object
+ * @param {object} b - The target object
+ * @param {function} resolver - A function that return the value to be kept. First argument is value a, second is value b
+ * @returns {object} A new object
+ */
+export const deepMergeWithResolver = ( a, b, resolver ) => {
```
bnchrch, line 191: 💅 ❓ `deepMergeWithResolver` is ~95% `deepMerge`. Could `deepMerge` just become `deepMergeWithResolver( a, b, ( _a, b ) => b )` to kill the duplication? Also the `[k]: ( () => {...} )()` inline IIFE is a little hard to scan, a ternary or small `resolveValue` helper would read cleaner. To be clear this is a nit (💅), feel free to disregard.

bnchrch (approved):
Approved, no major issues

sdk/core/src/interface/workflow.js
szanata: I changed so aggregations start null instead of an empty object

sdk/core/src/utils/utils.js
szanata: I made `deepMerge` use `deepMergeWithResolver` behind the scenes

PR #234, opened 2026-05-29
feat(core): Adding new logic to stop/start catalog workflows [OUT-469]

szanata:
## Summary

Adding new logic to stop/start catalog workflows on worker start that handle concurrent starts safely:
1. When starting a worker, calculate a hash from the source code;
2. Check if there is a running catalog workflow;
3. If the running catalog workflow hash matches the worker hash, keep using that catalog;
4. If the running catalog workflow hash doesn't match, tries to stop it;
5. After, even if stop failed, tries to start new catalog;
6. If the catalog start fails due the catalog already existing, check its hash again, if it is the same, stops the process;
7. Else, throws the error.

This allows that many instances of the same worker start at the same time, as they will not try to stop/start new catalogs.

### Logs from reused catalog:
image

### Logs from newly started catalog:
image

## Test plan

Unit tests, manual tests and e2e tests, all passed.

### DIFF
--- a/sdk/core/src/worker/index.js
+++ b/sdk/core/src/worker/index.js
@@ -13,6 +13,7 @@ import { bootstrapFetchProxy } from './proxy.js';
 import { messageBus } from '#bus';
 import './log_hooks.js';
 import { BusEventType } from '#consts';
+import { hashSourceCode } from './loader_tools.js';
 
 const log = createChildLogger( 'Worker' );
 
@@ -54,6 +55,9 @@ const callerDir = process.argv[2];
   log.info( 'Creating workflows catalog...' );
   const catalog = createCatalog( { workflows, activities } );
 
+  log.info( 'Computing catalog source code hash...' );
+  const catalogHash = await hashSourceCode( callerDir );
+
   log.info( 'Connecting Temporal...' );
   const proxy = grpcProxy ? { type: 'http-connect', targetHost: grpcProxy } : undefined;
   if ( proxy ) {
@@ -81,7 +85,7 @@ const callerDir = process.argv[2];
   registerShutdown( { worker, log } );
 
   log.info( 'Running worker...' );
-  await Promise.all( [ worker.run(), startCatalog( { connection, namespace, catalog } ) ] );
+  await Promise.all( [ worker.run(), startCatalog( { connection, namespace, catalog, catalogHash } ) ] );
 
   log.info( 'Closing connection...' );
   await connection.close();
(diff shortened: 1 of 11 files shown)


sdk/core/src/worker/interceptors.js
```diff
@@ -6,5 +6,9 @@ const __dirname = dirname( fileURLToPath( import.meta.url ) );
 
 export const initInterceptors = ( { activities, workflows, connection } ) => ( {
   workflowModules: [ join( __dirname, './interceptors/workflow.js' ) ],
-  activityInbound: [ () => new ActivityExecutionInterceptor( { activities, workflows, connection } ) ]
+  activity: [
+    () => ( {
```
szanata, line 10: This change is unrelated, it is just that `activityInbound` was deprecated. This is the new syntax.

sdk/core/src/worker/index.js
```diff
@@ -54,6 +55,9 @@ const callerDir = process.argv[2];
   log.info( 'Creating workflows catalog...' );
   const catalog = createCatalog( { workflows, activities } );
 
+  log.info( 'Computing catalog source code hash...' );
+  const catalogHash = await hashSourceCode( callerDir );
```
catsby, line 59: nit: it looks like the underlying `hashElement()` can throw. If it does, we don't seem to handle it and I imagine the resulting error would be direct from `hashElement` without much context. Maybe wrap here? 

```suggestion
  const catalogHash = await hashSourceCode( callerDir ).catch( error => {
    throw new Error( `Failed to hash source code at '${callerDir}': ${error.message}`, { cause: error } );
  } );
```

sdk/core/src/worker/loader_tools.js
```diff
@@ -302,3 +303,17 @@ export async function *importComponents( files ) {
...
+ * Creates a hash of all source code in a given dir
+ *
+ * @param {string} rootDir
+ * @returns {string} Hash value
+ */
+export const hashSourceCode = async rootDir => {
+  const { hash } = await hashElement( rootDir, {
+    folders: { exclude: [ '.*', 'node_modules', 'test_coverage', 'vendor', 'test' ] },
+    files: { include: [ '*.js', '*.cjs', '*.mjs', '*.ts', '*.yaml', '*.yml', '*.json', '*.prompt' ] }
+  } );
```
catsby: edge case: by default the hash includes the root path. For our container based deploys this is fine, the calling dir is always `"/app"`. For other users who are not using Docker-ish deployments, the roots could be different even if the source files are the same, like to VMs or something. 

The consequence is simply unnecessary stop/start of the catalog workflow, so it's not major, but maybe worth including `ignoreRootName` as an option. 

Not a blocker though

catsby (approved):
LGTM

sdk/core/src/worker/index.js
szanata: I added an error wrapper as you suggested, but I put it directly inside the `hashSourceCode()` function

sdk/core/src/worker/loader_tools.js
szanata: Good call. I have added `ignoreRootName`

catsby (approved):
LGTM! Thanks!

PR #236, opened 2026-06-02
refactor(core): new hooks payload and temporal info propagation [OUT-467]

szanata:
### Workflow hooks
Updated `onWorkflowStart()`, `onWorkflowEnd()`, `onWorkflowError()` hooks payload:

```js
{ eventId, workflowDetails, error? }
```

Where `workflowDetails` is an abstraction over [Temporal's `workflowInfo()`](https://typescript.temporal.io/api/interfaces/workflow.WorkflowInfo):
```ts
{
  attempt: number,
  continuedFromExecutionRunId?: string | undefined,
  firstExecutionRunId: string,
  parent?: {
    runId: string,
    workflowId: string,
    namespace: string
  } | undefined,
  root?: {
    runId: string,
    workflowId: string
  } | undefined,
  runId: string,
  runStartTime: number, // epoch
  startTime: number, // epoch
  workflowId: string,
  workflowType: string
}
```

### Error hook
Updated `onError()` hooks payload. The fields change according to the `source`:

#### Source is "workflow"
```js
{ eventId, source, workflowDetails, error }
```
Where `workflowDetails` is the same as in workflow hooks.

#### Source is "activity"
```js
{ eventId, source, workflowDetails, activityInfo, outputActivityKind, error }
```

Where `activityInfo` is Temporal's `activityInfo()` [function return](https://typescript.temporal.io/api/interfaces/activity.Info).

And `outputActivityKind` is the framework flavor of the Temporal Activity: `step`, `evaluator` or `internal_step`.

#### Source is "runtime"
```js
{ eventId, source, error }
```

### "on()" hook
Updated `on()` hooks with better typing and a new envelope.

#### Envelope
All events have these fields:
- `eventId`
- `workflowDetails`: Same from other hooks
- `activityInfo`: From Temporal
- `outputActivityKind`

#### Typing
Besides the envelope fields, each event also has its own fields, their types are specified by the event emitter:
- `on<HttpRequestEvent>( 'http:request', handler )` from `@outputai/http`;
- `on<HttpRequestCostEvent>( 'cost:http:request', handler )` from `@outputai/http`;
- `on<LLMUsageEvent>( 'cost:llm:request', handler )` from `@outputai/llm`

### Execution Context
Updated `getExecutionContext()` from `core/sdk_activity_integration` to return:
```js
{ activityInfo, workflowFilename }
```

### Trace File
Updated trace workflow node ids to use Temporal `runId`, instead of `workflowId`. This helps to create trees when child workflows "continued as new".

## Test plan

- [x] Manual
- [x] Unit tests
- [x] e2e tests


sdk/core/src/worker/log_hooks.js
```diff
@@ -15,43 +15,34 @@ const workflowLog = createChildLogger( 'Workflow' );
...
-const shouldLogActivityEvent = ( { kind } ) => kind !== ComponentType.INTERNAL_STEP;
+const serializedActivityFields = activityInfo => ( {
+  activityId: activityInfo.activityId,
+  activityType: activityInfo.activityType,
+  workflowId: activityInfo.workflowExecution.workflowId,
+  workflowType: activityInfo.workflowType,
+  runId: activityInfo.workflowExecution.runId
+} );
+
+const shouldLogActivity = activityInfo => activityInfo.activityType !== ACTIVITY_GET_TRACE_DESTINATIONS;
```
bnchrch, line 26: ❗ I think this quietly changes what gets logged. The old check was `kind !== ComponentType.INTERNAL_STEP`, which skips *every* internal step. This one only skips `ACTIVITY_GET_TRACE_DESTINATIONS`, so `sendHttpRequest` (the other `INTERNAL_STEP`) will start showing up in activity logs. `trace_engine.js` also still filters on `INTERNAL_STEP`, so the two modules now disagree on what counts as internal.

The payload already carries `outputActivityKind`, so we can keep the old behavior and stay consistent:
```suggestion
const shouldLogActivity = ( { outputActivityKind } ) => outputActivityKind !== ComponentType.INTERNAL_STEP;
```
(this needs `ComponentType` back in the `#consts` import). Was the narrower filter intentional? If so I'm missing the why.

sdk/core/src/worker/interceptors/activity.js
```diff
@@ -42,25 +42,24 @@ export class ActivityExecutionInterceptor {
...
   }
 
   async execute( input, next ) {
-    const startDate = Date.now();
-
-    const { workflowExecution: { workflowId }, activityId: id, activityType: name, workflowType: workflowName } = Context.current().info;
-    const { executionContext } = headersToObject( input.headers );
-    const { type: kind } = this.activities?.[name]?.[METADATA_ACCESS_SYMBOL];
-    const { path: workflowFilename } = this.getWorkflowEntry( workflowName );
+    const activityInfo = activityInfoFn();
```
bnchrch, line 58: ❗ Looks like `duration`/`durationMs` is gone from here and from `onWorkflowEnd`/`onWorkflowError`. `workflowDetails` gives us `startTime`/`runStartTime` but no end time, so consumers can't recompute it on their side.

sdk/core/src/internal_utils/workflow_context.js
```diff
@@ -0,0 +1,29 @@
+import { workflowInfo, continueAsNew, inWorkflowContext } from '@temporalio/workflow';
+
+export class WorkflowContext {
```
bnchrch, line 3: ❓ 💅 Why a class with a single static `build()` here? It never holds state or gets instantiated, so the `class`/`static` is doing nothing a plain function wouldn't. `trace_info.js` does the same thing, but the sibling right next door, `createWorkflowDetails`, is already a plain function. Could we make all three match?
```suggestion
export const buildWorkflowContext = () => {
```
WDYT? Happy to be wrong if there's a reason the builders want to be classes.

sdk/core/src/internal_utils/trace_info.js
```diff
@@ -0,0 +1,21 @@
...
+
+export class TraceInfo {
+
+  /**
+   * Builds the trace information propagated through workflow memo and activity headers.
+   * @param {object} options - Arguments to build trace information
+   * @param {boolean} options.disableTrace - Whether trace event emission should be disabled
+   * @returns {object} trace information
+   */
+  static build( { disableTrace } ) {
```
bnchrch, line 11: 🤔 Thinking out loud, not a request. We now have three projections of `workflowInfo()`: `WorkflowContext.info` (`{workflowId, runId}`), this `TraceInfo`, and `createWorkflowDetails`. They ride different paths (workflow fn vs memo vs headers) so this is probably fine, but the overlap between `WorkflowContext.info` and `workflowDetails` made me pause. Is there a world where those two converge later?

sdk/core/src/internal_utils/workflow_context.js
```diff
@@ -0,0 +1,29 @@
...
+   * @returns {object} context
+   */
+  static build() {
+    if ( !inWorkflowContext() ) {
+      return {
+        control: {
+          continueAsNew: async () => {},
+          isContinueAsNewSuggested: () => false
+        },
+        info: { workflowId: 'test-workflow', runId: 'test-run' }
```
bnchrch, line 16: 💅 Small one: the `'test-workflow'` / `'test-run'` / no-op `continueAsNew` test values now live inside the prod builder. It reads a little like test fixtures leaking into core. Not blocking, just flagging since it used to sit in the test path in `workflow.js`.

eslint.config.js
```diff
@@ -203,6 +203,16 @@ export default [
     }
   },
 
+  // Keep sdk/core TS specs linted without requiring a tsconfig for test-only files
+  {
+    files: [ 'sdk/core/**/*.spec.ts' ],
```
bnchrch, line 208: 💅 The `Pick<WorkflowInfo, ...>` drift-check in `temporal_context.spec.ts` is a nice idea, but it's the only TS spec in an otherwise-JS core package and it costs us this eslint carve-out + precedent. Could a JSDoc `@satisfies` or a tiny `.d.ts` type test get the same drift safety without the config change? Genuinely asking, not blocking.

sdk/core/src/tracing/processors/s3/index.js
```diff
@@ -7,7 +7,7 @@ import { JsonStreamStringify } from 'json-stream-stringify';
 
 const log = createChildLogger( 'S3 Processor' );
 
-const createRedisKey = ( { workflowId, workflowName } ) => `traces/${workflowName}/${workflowId}`;
+const createRedisKey = runId => `traces/${runId}`;
```
bnchrch, line 10: 💅 `createRedisKey( runId )` takes a bare string while every other trace fn in here takes the `traceInfo` object. Tiny consistency nit, take it or leave it:
```suggestion
const createRedisKey = ( { runId } ) => `traces/${runId}`;
```

bnchrch (commented):
Really solid refactor. Swapping the hand-rolled `executionContext`/`parentId` threading for Temporal's own info objects makes the hook payloads a lot cleaner, and `onError` going from "remap every field" to "just forward" is great.

Two things I'd love to sort before merge (both blocking): a behavior change in activity logging, and the dropped `duration` fields. The rest are nits, feel free to disregard. Notes inline.

bnchrch (approved):
Pre-approving to unblock. The two blocking notes (the activity log filter change and the dropped duration) are still worth a look before merge, but nothing here needs another round from me. Nice work.

sdk/core/src/worker/log_hooks.js
szanata: Yes, this was intentional. webhook steps make sense to be logged, as they carry some logic and might break.

sdk/core/src/worker/interceptors/activity.js
szanata: The `duration` was a convenience using Temporal's `startTime - Date.now()`.

But that was not accurate, didn't considered retries and was affected by delays.

To get accurate duration one has to consume Temporal's Temporal Workflow Execution History API.

If the user still needs an approximation. It can do it by himself using `Date.now()` when the event is fired.

sdk/core/src/internal_utils/workflow_context.js
szanata: I did this way because the calls itself uses temporal `workflowInfo()` function to build, and that technically is a mutation, so it felt more natural to me for this to be a factory-style class than a plain function.

sdk/core/src/internal_utils/trace_info.js
szanata: WorkflowContext info will be refactored in due time to expose `workflowInfo()` from temporal directly. I had faced this already, but decided not to do anything because I don't want to introduce too many new changes at one.

TraceInfo is different, it is immutable and reflects the root workflow, so that must be isolated. But I have a refactor planned to it as well.

sdk/core/src/internal_utils/workflow_context.js
szanata: Yes, but this is not for our tests, this is to allow users to write united tests and call the `context` argument without using `workflowInfo()` which doesnt work outside temporal. It allow for users to spy().

eslint.config.js
szanata: Those would flag drifts, but not break build without introducing more tools.

The TS test runs and fails properly. I agree that it looks out of the place. We can evaluate other solutions, but for now I think it is better to have something. At least it wasnt necessary to introduce `tsconfig`

sdk/core/src/tracing/processors/s3/index.js
szanata: I tend to leave that way when it is only a single arg. I also don't have any strong feelings.

sdk/core/src/worker/interceptors/activity.js
szanata: I added the `eventDate` to all hooks together with `eventId`

PR #239, opened 2026-06-03
feat(core): Adding worker telemetry logs [OUT-265]

szanata:
### Telemtry
Added worker telemetry logs: print Temporal worker status and node memory every X ms, configured by `OUTPUT_WORKER_TELEMETRY_INTERNAL_MS` env var. Default `0` - off.

#### Dev example
```
[info] Telemetry: Worker { status: { runState: "RUNNING", numHeartbeatingActivities: 0, workflowPollerState: "POLLING", activityPollerState: "POLLING", hasOutstandingWorkflowPoll: true, hasOutstandingActivityPoll: true, numCachedWorkflows: 1, numInFlightWorkflowActivations: 0, numInFlightActivities: 0, numInFlightNonLocalActivities: 0, numInFlightLocalActivities: 0 }, memory: { availableMemory: 7500000000, constrainedMemory: 20000000000000000000, memoryUsage: { rss: 582348800, heapTotal: 400000000, heapUsed: 200000000, external: 800000000, arrayBuffers: 300000000 } } }
```

### Prod example
```json
{
  "environment": "production",
  "level": "info",
  "memory": {
    "availableMemory": 7500000000,
    "constrainedMemory": 20000000000000000000,
    "memoryUsage": {
      "arrayBuffers": 1445268,
      "external": 800000000,
      "heapTotal": 400000000,
      "heapUsed": 200000000,
      "rss": 300000000
      }
    },
  "message": "Worker",
  "namespace": "Telemetry",
  "service": "output-worker",
  "status": {
    "activityPollerState": "POLLING",
    "hasOutstandingActivityPoll": true,
    "hasOutstandingWorkflowPoll": true,
    "numCachedWorkflows": 1,
    "numHeartbeatingActivities": 0,
    "numInFlightActivities": 0,
    "numInFlightLocalActivities": 0,
    "numInFlightNonLocalActivities": 0,
    "numInFlightWorkflowActivations": 0,
    "runState": "RUNNING",
    "workflowPollerState": "POLLING"
  },
  "timestamp": "2026-06-02T21:54:29.261+00:00"
}
```

### Logger refactor
Refactored the logger feature to have tests on it and make the loading of the options dynamic based on the NODE_ENV and to fix an issue where the development JSON serializer was not working for nested properties.

## Test plan

- [x] Manually tests with prod config
- [x] Manually tests with dev config
- [x] Unit tests
- [x] e2e tests


sdk/core/src/worker/configs.js
```diff
@@ -7,6 +7,7 @@ const coalesceEmptyString = v => v === '' ? undefined : v;
 
 const envVarSchema = z.object( {
   OUTPUT_CATALOG_ID: z.string().regex( /^[a-z0-9_.@-]+$/i ),
+  OUTPUT_WORKER_TELEMETRY_INTERNAL_MS: z.preprocess( coalesceEmptyString, z.coerce.number().int().nonnegative().default( 0 ) ),
```
catsby: Looks like a typo here: `OUTPUT_WORKER_TELEMETRY_INTERNAL_MS` should probably be `OUTPUT_WORKER_TELEMETRY_INTERVAL_MS`?

sdk/core/src/logger/development.js
```diff
@@ -0,0 +1,61 @@
...
+ * @param {string} v - A text value
+ * @returns {string} The color
+ * */
+const getColor = v =>
+  assignedColors.get( v ) ?? assignedColors.set( v, COLORS[assignedColors.size % COLORS.length] ).get( v );
+
+export const options = {
+  level: 'debug',
+  transports: [ new transports.Console() ],
+  format: format.combine(
```
catsby, line 52: we lost the development `defaultMeta` here

https://github.com/growthxai/output/blob/34badf9feb80a9f8ab83c5945de7342ab8d32d0a/sdk/core/src/logger.js#L60

sdk/core/src/logger/production.js
```diff
@@ -0,0 +1,15 @@
+import { transports, format } from 'winston';
+
+export const options = {
+  level: 'info',
+  transports: [ new transports.Console() ],
+  defaultMeta: {
+    service: 'output-worker',
+    environment: process.env.NODE_ENV || 'development'
```
catsby: if this file is load dynamically based on `NODE_ENV`, do we still need or want this check? and if we do, do we still want to default to `development` or should we default to `production`

sdk/core/src/worker/index.spec.js
```diff
@@ -172,6 +176,18 @@ describe( 'worker/index', () => {
     await vi.waitFor( () => expect( exitMock ).toHaveBeenCalled() );
   } );
 
+  it( 'calls setupTelemetry with worker', async () => {
```
catsby: does this test anything that isn't being tested on line 135?

sdk/core/src/logger/index.js
```diff
@@ -0,0 +1,14 @@
+import winston from 'winston';
+
+const { options } = await import( process.env.NODE_ENV === 'production' ? './production.js' : './development.js' );
+
+// creates the de root winston logger
```
catsby:
```suggestion
// creates the the root winston logger
```

sdk/core/src/logger/development.js
```diff
@@ -0,0 +1,61 @@
...
+};
+
+/** This instance color schema */
+const COLORS = shuffleArray( Object.values( Color ) );
+
+/** Stores all assigned colors per namespace. */
+const assignedColors = new Map();
+
+/**
+ * Get the previous assigned color for this value or if not present, assign a new one an store it
```
catsby:
```suggestion
 * Get the previous assigned color for this value or if not present, assign a new one and store it
```

sdk/core/src/logger/development.js
```diff
@@ -0,0 +1,61 @@
...
+export const options = {
+  level: 'debug',
+  transports: [ new transports.Console() ],
+  format: format.combine(
+    format.colorize(),
+    format.metadata(),
+    format.printf( ( { level, message, metadata } ) => {
+      const { namespace, ...fields } = metadata;
+      const jsonText = Object.keys( fields ).length > 0 ? formatJson( fields ) : null;
+      return `[${level}] \x1b[38;5;${getColor( namespace )}m${namespace}: ${message}\x1b[0m${jsonText ? ' ' + jsonText : ''}`;
```
catsby, line 58: does `namespace` have a default value? We provide it to child loggers but could the root logger be used and would it have a namespace?

catsby (approved):
✅ pending some typo cleanups

sdk/core/src/logger/development.js
catsby: that was supposed to be a question but I guess I cut it short... do we need to restore the default meta here?

szanata: No. It was not being printed. We were ignoring those fields.

sdk/core/src/logger/production.js
szanata: No, I forgot to remove this.

sdk/core/src/logger/development.js
szanata: The root logger is not used and it is not being exported.

sdk/core/src/worker/index.spec.js
szanata: You are right. That was added to be explicit, but it doesnt make sense.

sdk/core/src/worker/configs.js
szanata: Oh, I can't believe I left that in... I am fixing.

PR #243, opened 2026-06-05
refactor(llm): Making providers optional peer deps [OUT-473]

szanata:
### Changing providers dependencies to peer
Updated all `@ai-sdk/*` provider packages from `/llm` to be peer dependencies.

The library still ships support for the same providers as before, but the peer dependency can satisfied by a range: 
 - `@ai-sdk/amazon-bedrock`: `>=4 <5`
- `@ai-sdk/anthropic`: `>=3 <4`
- `@ai-sdk/azure`: `>=3 <4`
- `@ai-sdk/google-vertex`: `>=4 <5`
- `@ai-sdk/openai`: `>=3 <4`
- `@ai-sdk/perplexity`: `>=3 <4`

Also updates `ai` to be a peer, since when using newer versions of the providers, users might need to resolve to a different `ai` versions, its range is `>=6 <7`.
 
This setup allows users to bring their own versions of these providers, aside from what we ship (as long as they satisfy our supported range).

### Providers are not lazy init

Providers are imported in the code bundle, but they are only initialized when required by a `.prompt` file during runtime.

### Removed web search tools
Removed Exa, Tavily and Perplexity search dependencies as they were used just to re-export some web search tools. Users can bring their own versions now.

### Renamed getRegisteredProviders

The function `.getRegisteredProviders()` was misleading, as it listed built-in providers beside registered, and returned only their NAMES no the whole function. It was replaced by `.getProviderNames()`.

## Test plan

- [x] manual
- [x] unit
- [x] e2e


sdk/llm/src/index.d.ts
```diff
@@ -384,7 +379,7 @@ export function registerProvider(
  *
  * @returns Array of provider name strings
  */
-export function getRegisteredProviders(): string[];
+export function getProviderNames(): string[];
```
catsby, line 427: This looks like a breaking change that's not mentioned in the changeset, which itself is labeled as just a patch update. Is this not a breaking change?

sdk/llm/src/ai_provider.js
```diff
@@ -0,0 +1,100 @@
...
+    console.debug( `LLM: Provider "${name}" loaded` );
+    return provider;
+  } catch ( error ) {
+    if ( error.code === 'MODULE_NOT_FOUND' && error.message.startsWith( `Cannot find module '${pkg}'` ) ) {
+      throw new FatalError( `Provider "${name}" requires "${pkg}". Install it to use this provider.`, { cause: error } );
+    }
+    if ( [ 'ERR_REQUIRE_ESM', 'ERR_REQUIRE_ASYNC_MODULE', 'ERR_PACKAGE_PATH_NOT_EXPORTED' ].includes( error.code ) ) {
+      throw new FatalError( `Provider "${name}" package "${pkg}" cannot be loaded synchronously. Use a compatible version.`, { cause: error } );
+    }
+    throw error;
```
catsby: Should we include the provider name and package if we have it? Otherwise I wonder if we'll just throw a raw AI-SDK error without any idea where it came from in our code

test_workflows/src/workflows/web_search/workflow.ts
catsby, line 1: Why remove this entire workflow that demonstrated/tested using these external web search tools, instead of updating them to do so in the new way?

catsby (changes requested):
This seems like an optimization step gone too far. I'm concerned this reduces our onboarding experience and gives a less "batteries included" feel. Let's discuss in Slack

test_workflows/src/workflows/web_search/workflow.ts
szanata: I think these are out of the scope of the test_workflows. That belongs to doc examples.

szanata:
@clint. I updated the PR so the dependencies are not optional.

bnchrch (commented):
Really like where this lands. The `ai_provider.js` / `ai_model.js` split is clean, and moving provider/model/tool validation up into `validatePrompt` lets `loadTextModel` become a one liner. Nice.

One thing I want to talk through before approving, plus a few nits inline.

The eager-import path has a consequence the docs don't match. Because the six `@ai-sdk/*` imports at the top of `ai_provider.js` are static, a missing package throws `ERR_MODULE_NOT_FOUND` the moment someone imports `@outputai/llm`, before any prompt is read and before `getProvider` runs. So two things the docs promise can't actually happen today: the "throws an error telling you which package to add" message (the only friendly errors in `getProvider` are unknown-name and init-failure), and "install only what you use" (install only anthropic and the import blows up).

Two ways to square it, your call. Both fine by me:
1. Keep option 2 as we decided and just fix the docs to match the code. Cheapest.
2. Go to the option-3 shape: move imports into `getProvider` as `await import()`, catch `ERR_MODULE_NOT_FOUND`, rethrow as "install @ai-sdk/x". Delivers the friendly error + real uninstallability, but makes the load path async (the ripple you stepped back from).

The only hard blocker is that the docs and the code disagree right now (see the inline on `llm.mdx`). WDYT?

🧪 Separately: the `web_search` workflow got deleted rather than migrated. It was our only e2e/example using provider tools and the search exports we're dropping. The migration guide shows people how to re-import `perplexitySearch` from `@perplexity-ai/ai-sdk`, but we no longer ship a worked example, and it drops an e2e this PR says it ran. Could we migrate it to import from `@perplexity-ai/ai-sdk` instead of removing it?

docs/guides/packages/llm.mdx
```diff
@@ -584,6 +584,25 @@ Company size: {{ size }} employees
...
+| `bedrock` | `@ai-sdk/amazon-bedrock` |
+| `perplexity` | `@ai-sdk/perplexity` |
+
+For example, an OpenAI and Anthropic project should include:
+
+```bash
+npm install @outputai/llm @ai-sdk/openai @ai-sdk/anthropic
+```
+
+If a prompt uses a built-in provider whose package is not installed, `@outputai/llm` throws an error telling you which package to add.
```
bnchrch: ❗ This isn't quite true with the current setup. The provider packages are imported statically at the top of `ai_provider.js`, so a missing package throws a generic `ERR_MODULE_NOT_FOUND` when `@outputai/llm` is first imported, before `getProvider` ever runs. There's no "which package to add" error on that path today. Either we make the imports dynamic (`await import()` inside `getProvider`) so this message can actually fire, or we reword here. Tied to this: line 587 calls them "optional", but with static imports they're effectively required (the changeset says required too). Can we get the two to agree?

sdk/llm/src/ai_model.js
```diff
@@ -1,181 +1,59 @@
...
-  const provider = providers[providerName];
-
-  if ( !provider ) {
-    const availableProviders = Object.keys( providers ).join( ', ' );
-    throw new Error( `Invalid provider "${providerName}". Valid providers: ${availableProviders}` );
-  }
-
-  return provider( modelName );
-}
+export const loadTextModel = prompt => getProvider( prompt.config.provider )( prompt.config.model );
```
bnchrch, line 10: 💎 The three duplicated "Invalid provider, valid providers…" blocks collapsing into one `getProvider()` lookup is great. Reads really well now.

sdk/llm/src/ai_model.js
```diff
@@ -1,181 +1,59 @@
...
+  const provider = getProvider( providerName );
 
   if ( !provider.tools || typeof provider.tools !== 'object' ) {
-    throw new Error( `Provider "${providerName}" does not support provider-specific tools.` );
+    throw new ValidationError( `Provider "${providerName}" does not support provider-specific tools.` );
   }
 
-  const tools = {};
+  const list = Object.entries( provider.tools ).filter( ( [ _, v ] ) => typeof v === 'function' ).map( e => e[0] );
+  const invalid = Object.keys( toolsConfig ).filter( name => !list.includes( name ) );
```
bnchrch: 💅 Small idiom mix: the filter destructures `([ _, v ])` but the map indexes `e[0]`, and the `// load all tools` comment below just restates the code. Reads cleaner with a named list and one `Object.fromEntries`:
```suggestion
  const supportedTools = Object.entries( provider.tools )
    .filter( ( [ , factory ] ) => typeof factory === 'function' )
    .map( ( [ name ] ) => name );
  const invalid = Object.keys( toolsConfig ).filter( name => !supportedTools.includes( name ) );
```
and the return becomes:
```js
return Object.fromEntries(
  Object.entries( toolsConfig ).map( ( [ name, config ] ) => [ name, provider.tools[name]( config ) ] )
);
```
Totally optional.

sdk/llm/src/ai_provider.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+  if ( providerInitializers[name] ) {
+    try {
+      return initializedProviders[name] = providerInitializers[name]( { fetch: customFetch } );
+    } catch ( error ) {
+      throw new FatalError( `Failed to initialize provider "${name}": ${error.message}`, { cause: error } );
+    }
+  }
+
+  throw new FatalError( `Unsupported provider "${name}"` );
```
bnchrch, line 85: ❓ 💅 Small consistency thing: `getProvider` throws `FatalError` for an unsupported provider, but `loadImageModel`/`loadTools` throw `ValidationError` for the sibling "provider can't do X" config errors. They're all non-retryable config mistakes. Any reason they differ, or should they be one type? Nit.

szanata: Both `ValidationError` and `FatalError` are "fatal" errors (they don't retry the workflow, instead finish it). I thought that validation was more idiomatic on that scenario.

docs/guides/packages/llm.mdx
szanata: Fixed the docs, guide, etc

sdk/llm/src/ai_model.js
szanata: I improved that code part.

sdk/llm/src/ai_model.js
```diff
@@ -1,181 +1,60 @@
...
+  const provider = getProvider( providerName );
 
-  if ( !provider.tools || typeof provider.tools !== 'object' ) {
-    throw new Error( `Provider "${providerName}" does not support provider-specific tools.` );
+  if ( !provider.tools ) {
+    throw new ValidationError( `Provider "${providerName}" does not support provider-specific tools.` );
   }
 
-  const tools = {};
+  const supportedTools = Object.keys( provider.tools );
```
bnchrch, line 48: ❗ The rewrite dropped the `typeof v === 'function'` filter here (and relaxed the guard above from `typeof provider.tools !== 'object'` to just `!provider.tools`), and the two tests that covered those cases got removed with it. So now `Object.keys(provider.tools)` advertises any non-callable keys as available tools, and a prompt referencing one passes the `invalidTools` check then blows up on `provider.tools[name](args)` with a raw `TypeError` instead of our friendly "Invalid tool(s)" error. Edge case, but it's exactly the cross-version fragility this PR is trying to avoid. One-liner restores it:
```suggestion
  const supportedTools = Object.keys( provider.tools ).filter( name => typeof provider.tools[name] === 'function' );
```
Mind also re-adding the two dropped tests ('tools value is not an object' and the non-function property one) so this stays covered?

bnchrch (commented):
Thanks for the quick turnaround. The docs blocker is fully fixed (the false "tells you which package to add" line is gone and the ranges read clearly), and I really like that `web_search` got migrated to `@perplexity-ai/ai-sdk` instead of deleted. `loadTools` reads great now too.

One thing the rewrite introduced that I'd like to fix before merge (inline on `ai_model.js`), plus two notes on the newer changes:

- 📚 With `ai` now a peer dep (nice, consistent), can we make sure the migration guide / install docs tell people to install `ai` alongside the `@ai-sdk/*` packages? On pnpm-style strict peers a missing `ai` won't auto-install.
- 🔮 `CompatibleToolSet` is a sensible fix for the cross-zod-peer problem and the JSDoc explaining *why* is great. Just flagging that it's a hand-rolled mirror of AI SDK's `Tool`, so it can drift as the SDK changes. Worth a glance whenever we bump the `ai` range. Not blocking.

sdk/llm/src/index.d.ts
bnchrch: ❗ +1 to clint here. Renaming `getRegisteredProviders` → `getProviderNames` and removing the search re-exports are breaking API changes, but this is labeled `minor`. For 0.x that can be policy, but the bump (or at least the changeset note) should make the breakage explicit since there's a migration guide riding on it. What's our convention for breaking changes pre-1.0?

bnchrch (approved):

catsby (approved):

PR #246, opened 2026-06-08
refactor(llm): Better NoGeneratedObjectError handling

szanata:
## Summary

Follow up to https://github.com/growthxai/output/pull/245. This keeps the same idea but creates a copy of AI SDK's. `NoGeneratedObjectError` with the better message, so to keep all other properties and support to `.isInstance()`.

## Test plan

- [x] unit
- [x] manual

### DIFF
--- a/sdk/llm/src/utils/error_handler.js
+++ b/sdk/llm/src/utils/error_handler.js
@@ -48,10 +48,9 @@ const toFatalError = ( error, extraMessage = '' ) => new FatalError(
 /**
  * Map an AI SDK error to a framework specific error:
  *
- * Unrecoverable errors become FatalErrors.
- *
- * NoObjectGeneratedError from invalid schema are reinitialized with a better message.
- *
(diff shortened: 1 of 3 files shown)


bnchrch (approved):

sdk/llm/src/utils/error_handler.js
```diff
@@ -42,20 +45,35 @@ const toFatalError = ( error, extraMessage = '' ) => new FatalError(
...
 
+/**
+ * Map an AI SDK error to a framework specific error:
+ *
+ * Unrecoverable errors become FatalErrors.
+ *
+ * NoObjectGeneratedError from invalid schema are reinitialized with a better message.
+ *
+ * @param {object} error - Original Error
+ * @returns {object} A new Error
```
catsby, line 55: nit: unless I'm mistaken, there are 3 paths that return the same error, not a new one. Around lines 60, line 78, and the fall-through one on line 121. I'll defer to others if that distinction matters though

catsby (approved):

PR #247, opened 2026-06-09
feat(credentials): asymmetric (sealed) credentials mode

nicastelo:
## Summary

Adds an opt-in **sealed** credentials mode alongside the existing symmetric scheme. Sealed credentials are encrypted to a **committed public key** and decrypted with a **private key only the runtime needs** — so contributors can add credentials without any secret on their machine, and there's no shared encryption key to misconfigure. Legacy symmetric credentials keep working unchanged; adoption is per-environment and optional.

## Why

The default scheme is symmetric — one key both encrypts and decrypts. So everyone who edits credentials needs the shared secret, and committing a file encrypted with the **wrong** `OUTPUT_CREDENTIALS_KEY_PRODUCTION` fails *silently*: production can't decrypt it, every workflow that uses those credentials breaks, and you only find out at runtime, in prod, on first use. The natural failsafes don't help: a startup check only runs once the bad file is already committed and deploying (too late to prevent it), and a CI decrypt check would mean handing the decryption key to CI — so the failsafe becomes the exposure.

Sealed mode splits the roles: the public key (encrypt) is committed; the private key (decrypt) lives only in the runtime and your secret manager. There's no secret encryption key to get wrong in the first place, and because the encryption identity is public, `verify` can confirm a file was sealed for the canonical key with **no secret at all** — so it runs in CI, *before merge*, and exposes nothing. A wrong-key commit fails that check instead of reaching production.

## How it works

Each value is sealed individually with a libsodium `crypto_box_seal`-style construction: an ephemeral X25519 keypair does ECDH with the recipient public key, HKDF-SHA256 derives an AES-256-GCM key, and the value is encrypted under it; the ephemeral public key is prepended so the private-key holder can reconstruct the shared secret. (Built on `@noble/curves` + `@noble/ciphers` + `@noble/hashes`.)

Files stay plain YAML — names are visible (they aren't secret), only values are encrypted:

```yaml
# config/credentials/production.yml.enc (sealed)
__format__: sealed-v1
__recipient__: 29964e49...        # the committed public key
anthropic:
  api_key: "sealed:BcqmP+jqI2N2..."
openai:
  api_key: "sealed:fCkZn2jgrRSW..."
```

## CLI

| Command | What it does |
|---|---|
| `output credentials init --sealed` | Generate a keypair; write the gitignored private key, the committed public key, and a sealed file. |
| `output credentials set <path> <value>` | Seal a value with the committed public key — **no private key needed**. |
| `output credentials verify` | Confirm the file was sealed for the committed public key. **Needs no secret** → safe in CI. With a private key present, also confirms every value opens. |
| `output credentials migrate --to-sealed` | Convert a legacy symmetric file to sealed form; emit the new private key to configure in the runtime. |
| `edit` / `show` / `get` | Work transparently on sealed files (private key required — they decrypt). |

The runtime private key uses the **same** env var as before (`OUTPUT_CREDENTIALS_KEY[_ENV]`) — it now holds the X25519 private key.

## ⚠️ Security model & the write-access trade-off

Because the encryption key is **committed**, sealing is a public operation: **anyone who can write to the repo can add or change a credential value** (they cannot *read* existing values — that still requires the private key). This is the deliberate trade for keyless contribution; with the symmetric scheme only key-holders could write.

Move write-control to the repo, where it belongs:

- **CODEOWNERS** on `config/credentials/**` and the `.pub` files
- **Branch protection**: required review (so CODEOWNERS approval is mandatory) + signed commits
- The key-free **`verify`** in CI to catch a wrong-key commit before it deploys

The private key never enters the repo, so even a malicious write can only *add* a value sealed to the canonical key — never exfiltrate an existing one. See the operations guide for the full setup.

Additional hardening:

- **Fail-fast**: a wrong/missing private key throws `SealedRecipientMismatchError` at load, not a generic decrypt error surfaced lazily on first use. The recipient header is mandatory and must match the key.
- `seal()` validates the recipient public key; `open()` rejects malformed tokens.
- `migrate` writes atomically (temp + fsync + rename) with rollback, so a failed migration is non-destructive; key files are written `0600`.
- `edit` preserves the ciphertext of unchanged values, so a single edit only diffs what actually changed.

## Backward compatibility

Fully opt-in. Legacy symmetric files are auto-detected and handled exactly as before. Migration is per-environment; nothing changes until you run `init --sealed` or `migrate --to-sealed`.

## Open questions for reviewers

- **Rotation / re-keying** is currently done via `edit` (decrypt with the old key, reseal to the new recipient). There's no dedicated `rotate` command in v1 — acceptable?
- The **`verify` guarantee is scoped**: the key-free CI check catches a wrong *recipient*; catching a corrupt *value* requires a key-holding verify. Documented as such.

## Testing

- New unit suites for the crypto core (`sealedbox`) and the sealed provider path.
- Integration tests for `init --sealed`, `set` without a key, `migrate`, wrong-key rejection, `0600` perms, edit diff-preservation, and recipient-rotation safety.
- Full suite green: 2079 tests.

New dependencies: `@noble/curves`, `@noble/hashes` (both `2.2.0`, same vendor as the existing `@noble/ciphers`).

## Background

This isn't theoretical — it's the production incident that prompted the feature. A credentials file was committed encrypted with the wrong `OUTPUT_CREDENTIALS_KEY_PRODUCTION`; production didn't have a matching key, couldn't decrypt, and every workflow that relied on those credentials broke. Nobody caught it until runtime in prod, and recovery meant redeploying a corrected file and re-encrypting the affected secrets. Under sealed mode that commit never reaches production — there's no secret key to mismatch, and the key-free `verify` flags a wrong recipient in CI, before merge.

### DIFF
--- a/.changeset/sealed-credentials.md
+++ b/.changeset/sealed-credentials.md
@@ -1,0 +1,44 @@
+---
+"@outputai/credentials": minor
+"@outputai/cli": minor
+---
+
+## Asymmetric (sealed) credentials
+
+Adds an opt-in **sealed** credentials mode alongside the existing symmetric scheme.
+Sealed credentials are encrypted to a **committed public key** and decrypted with a
+**private key that only the runtime needs** — so contributors can add credentials
+without any secret on their machine, and there is no secret encryption key to
+misconfigure.
+
+Each value is sealed individually (libsodium `crypto_box_seal`-style: ephemeral X25519
+ECDH → HKDF-SHA256 → AES-256-GCM), so the credential file is plain YAML with visible
+keys and `sealed:` values plus a `__format__`/`__recipient__` header.
+
+**`@outputai/credentials`**
+- New crypto/format API: `generateKeypair`, `publicKeyFromPrivate`, `seal`, `open`,
+  `sealTree`, `openTree`, `resealTree`, `detectFormat`, `parseSealedDocument`,
+  `serializeSealedDocument`, `openSealedDocument`, `isSealedValue`, `isValidKeyHex`, and
+  `resolvePublicKeyPath` / `resolveWorkflowPublicKeyPath`.
+- The provider auto-detects sealed files and opens them with the configured private key.
+  The recipient header is mandatory and must match the key, so a wrong or missing key
+  fails fast with `SealedRecipientMismatchError` (never opens blindly). `seal` validates
+  the recipient public key and `open` rejects malformed tokens. Legacy symmetric files
+  keep working unchanged.
+
+**`@outputai/cli`**
+- `output credentials init --sealed` — generate a keypair, write the gitignored private
+  key and the committed public key, and create a sealed file.
+- `output credentials set` — seals the value with the public key; no private key needed.
+- `output credentials verify` — verifies a file was sealed for the committed public key.
+  This check needs **no secret**, so it is safe to run in CI.
+- `output credentials migrate --to-sealed` — convert a legacy symmetric file to sealed
+  form and emit the new private key to configure in the runtime. Writes are atomic with
+  rollback so a failed migration is non-destructive, and key files are written `0600`.
+- `edit` / `show` / `get` transparently support sealed files (private key required).
+  `edit` preserves the ciphertext of unchanged values so a single edit only diffs what
+  actually changed.
+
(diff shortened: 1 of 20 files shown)


bnchrch:
Closing in favor of https://github.com/growthxai/output/pull/250

PR #250, opened 2026-06-09
feat(cli): warn before re-encrypting credentials with a mismatched key

bnchrch:
## Problem

The `credentials set` and `credentials edit` commands re-encrypt the credentials file using the currently-resolved key (`OUTPUT_CREDENTIALS_KEY` env var or key file). If the wrong key is in use — e.g. a stale env var, or the wrong environment's key — the file can be silently re-encrypted under that key.

- Until now the only signal was a cryptic AES-GCM decryption failure with no clear remediation.
- There was no supported way to intentionally proceed (re-key) when that's actually what you want.

## Solution

- **Service** — add `checkKeyMatchesCredentials()` (returns `no_file` / `match` / `mismatch` by attempting a real `decrypt()` with the resolved key) and a shared `reEncryptKeyMismatchMessage()` in `credentials_service.ts`.
- **`set` / `edit`** — after the existing-file check, gate on the result: a `mismatch` without `--force` aborts with a clear warning that the wrong key may be in use and values would be discarded. New `--force` / `-f` flag proceeds anyway, skipping decryption and re-encrypting from an empty file under the current key.

Scope is intentionally limited to `set` and `edit`. `init` is unchanged — it already guards via its existence check and always generates a fresh key by design.

> Note: credentials use symmetric AES-256-GCM, so the issue's "private key" maps to the single `credentials.key` / `OUTPUT_CREDENTIALS_KEY`.

## Test plan

- [x] `npm run lint` passes
- [x] `sdk/cli` `tsc --noEmit` passes (0 errors)
- [x] Unit specs (`set`/`edit`/service) — 40 pass, incl. new mismatch-with/without-`--force` cases
- [x] `credentials_service.integration` — 24 pass, incl. new `no_file` / `match` / `mismatch` cases for the helper
- [ ] Reviewer: manual smoke — `init`, point `OUTPUT_CREDENTIALS_KEY` at a different key, then `credentials set` (expect abort) and `credentials set --force` (expect warn + write)

## Follow-up (out of scope)

`copy_assets.spec.ts` fails locally in this worktree because the build's `copy-assets` step hits a `native-copyfiles`/Node glob incompatibility (`options.exclude must be of type function`), leaving `dist/templates/` empty. Pre-existing and unrelated to this change.

Ref: [OUT-478](https://linear.app/growthx/issue/OUT-478)

### DIFF
--- a/.changeset/warn-reencrypt-credentials-mismatched-key.md
+++ b/.changeset/warn-reencrypt-credentials-mismatched-key.md
@@ -1,0 +1,5 @@
+---
+"@outputai/cli": patch
+---
+
+`credentials set` and `credentials edit` now check whether the current key can decrypt the existing credentials file before re-encrypting. On a key mismatch they abort with a clear warning that the wrong key may be in use and the file would be re-encrypted under a different key. Pass `--force` / `-f` to proceed anyway (re-encrypts from empty, discarding the undecryptable values).
--- a/sdk/cli/src/commands/credentials/edit.spec.ts
+++ b/sdk/cli/src/commands/credentials/edit.spec.ts
@@ -27,6 +27,8 @@ describe( 'credentials edit command', () => {
   beforeEach( () => {
     vi.clearAllMocks();
     vi.mocked( credentialsService.credentialsExist ).mockReturnValue( true );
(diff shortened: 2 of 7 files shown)


szanata (approved):
I think these are a great addition to the feature. 👍

PR #251, opened 2026-06-09
feat(api): expose workflow failure details and log Temporal errors

bnchrch:
## Problem

- When a workflow fails, `/workflow/run` and `/workflow/:id/result` returned only a single `error` string from the deepest cause — the friendly message, error type, and retryable flag were lost, so Atlas/Studio couldn't present meaningful failures. (It also tended to surface "activity failure" while hiding the real message in the stack.)
- When a Temporal client call failed in production (e.g. `ServiceError: Failed to query Workflow`), API logs showed only that one line — no gRPC `code`/`details`/`metadata`, cause chain, query name, or task queue — making it impossible to tell catalog-unavailable from worker saturation, a query timeout, or bad routing.

## Solution

- **API responses** — failed `WorkflowResult`s now include a structured `errorDetails { message, name, retryable, activityId, cause }` alongside the existing `error` string (additive, backward-compatible). `message`/`name` come from the **innermost** `ApplicationFailure` (the user's actual error), skipping Temporal's `WorkflowFailedError`/`ActivityFailure` wrappers — so you get "Foo"/`Error`, not "Activity task failed". `retryable` reflects Temporal's `nonRetryable` flag (`null` when unreported); `activityId` is the failing activity key (`"{workflow}#{step}"`, e.g. `test_error#thrower`; `null` for workflow-level/child-workflow failures); `cause` is a sanitized chain (`name`/`message`, **no stack**).
- **Logging** — a single `serializeErrorChain()` walks the `.cause` chain and folds in gRPC `code`/`codeName`/`details` + **redacted** metadata (key names + safe allowlist; auth headers never logged). `error_handler` logs unhandled 500s with that chain + `stack` + annotated `taskQueue`/`query`/`workflowId`. **Client-facing HTTP responses stay sanitized** (no stack/cause/metadata).
- **Docs** — `errorDetails` added to the `WorkflowResultResponse` OpenAPI schema; `api/openapi.json`, the docs copy, and the Orval-generated CLI client regenerated.

Closes OUT-233 — Ref: [OUT-233](https://linear.app/growthx/issue/OUT-233/expose-workflow-failure-details-in-output-api-responses)

## Proof

**Surface A** — `POST /workflow/run` of a deliberately failing workflow (`test_error`):
```json
{
  "status": "failed",
  "error": "Foo",
  "errorDetails": {
    "message": "Foo",
    "name": "Error",
    "retryable": true,
    "activityId": "test_error#thrower",
    "cause": { "name": "WorkflowFailedError",
      "cause": { "name": "ActivityFailure",
        "cause": { "name": "ActivityFailure",
          "cause": { "name": "Error", "message": "Foo" } } } }
  }
}
```

**Surface B** — catalog query against an unreachable Temporal; rich server log, client still gets a sanitized 500:
```
[error] ServiceError: Failed to query Workflow {"taskQueue":"main","query":"get","stack":"… at getCatalog … at runWorkflow …","cause":{"name":"ServiceError","message":"Failed to query Workflow","cause":{"name":"Error","message":"14 UNAVAILABLE: … ECONNREFUSED …:7233","code":14,"codeName":"UNAVAILABLE","details":"…"}}}
[http] POST /workflow/run 500 61b
```

## Test plan

- [x] `npm run lint` passes
- [x] API unit tests — pass across `utils.spec.js`, `temporal_client.spec.js`, `error_handler.spec.js` (incl. the double-wrapped-chain regression + `activityId` extraction)
- [x] Manual smoke (`./run.sh dev`, Node 24): both surfaces verified live (failure shape confirmed pre-rename; activity-key format `test_error#thrower` confirmed from the live trace)
- [ ] `npm test` full suite — green except 2 pre-existing `sdk/cli` copy-assets tests (see Follow-up)

## Notes for reviewers

- Addresses @szanata's review (see inline replies): unified the two serializers into `serializeErrorChain`, removed the `alreadyLogged` flag (logging centralized in `error_handler`), inlined the duck-type, `type`→`name`, `&&`-spreads.
- **Renamed `failure` → `errorDetails`** (type `WorkflowErrorDetails`) and added `activityId` (the failing step) — see commit 5146ed2. Kept `error` (string) + `errorDetails` (object) rather than collapsing to one field: `error` is a string in the shipped API and read as one by Atlas/Studio/CLI, so collapsing is a breaking change; OUT-233 is additive. The rename sharpens the summary-vs-structured split. Open to a single structured field in a future major bump.
- **Duck-typing the Temporal failure** (`typeof e.type === 'string'`) instead of `instanceof ApplicationFailure` keeps `utils.js` free of a `@temporalio` import and works on the protobuf-reconstructed chain.

## Follow-up (out of scope)

- gRPC availability errors (`UNAVAILABLE`/`DEADLINE_EXCEEDED`/`RESOURCE_EXHAUSTED`) → clean `503 CatalogNotAvailableError` + sanitized `rootCause`. Deferred. Related: OUT-452.
- `native-copyfiles@2.0.1` breaks `build:packages` / `./run.sh dev` on current Node (`fs.glob` `exclude` must be a function) — reproduced on Node 22 **and** 24. Fixed locally in `copyassets.sh` (find/cp) but kept out of this PR — worth its own `fix(cli)` PR.

### DIFF
--- a/api/src/clients/temporal_client.js
+++ b/api/src/clients/temporal_client.js
@@ -171,7 +171,7 @@ export const extractWorkflowInput = history => {
  * @property {string|number|boolean|object|array|undefined|null} output - The workflow output
  * @property {object|null} trace - Trace information
  * @property {string|null} error - Error message if failed
- * @property {object|null} failure - Structured failure details if failed (message, type, retryable, cause), null otherwise
+ * @property {object|null} errorDetails - Structured failure details if failed (message, name, retryable, activityId, cause), null otherwise
  */
 
 /**
@@ -205,10 +205,10 @@ const buildWorkflowResult = ( { workflowId, status, runId, input, result, error 
       trace: extractErrorDetail( error, 'trace' ),
       aggregations: extractErrorDetail( error, 'aggregations' ),
       error: extractErrorMessage( error ),
-      failure: extractFailure( error )
+      errorDetails: extractFailure( error )
     } : {
       error: null,
-      failure: null
+      errorDetails: null
     } )
   } );
 
--- a/api/src/clients/temporal_client.js
+++ b/api/src/clients/temporal_client.js
@@ -1,6 +1,6 @@
 import { Client, Connection, defaultPayloadConverter } from '@temporalio/client';
 import { temporal as temporalConfig } from '#configs';
-import { buildWorkflowId, extractErrorDetail, extractErrorMessage, extractFailure, serializeTemporalError, takeFromAsyncIterable } from '#utils';
+import { buildWorkflowId, extractErrorDetail, extractErrorMessage, extractFailure, takeFromAsyncIterable } from '#utils';
 import { logger } from '#logger';
 import {
   WorkflowNotFoundError,
@@ -41,12 +41,10 @@ const getCatalog = async ( { client, taskQueue } ) => {
     if ( error instanceof WorkflowNotFoundError ) {
       throw new CatalogNotAvailableError( 3 );
     }
-    // Annotate context the error_handler's serializer can't recover on its own, then log the full
-    // nested Temporal/gRPC detail here — the only place that knows the task queue + query name.
+    // Annotate the catalog/query context (the only place that knows it) so the error_handler can
+    // surface it when it logs the failure centrally.
     error.taskQueue = taskQueue;
     error.query = 'get';
-    logger.error( 'Catalog query failed', { taskQueue, query: 'get', ...serializeTemporalError( error ) } );
-    error.alreadyLogged = true;
     throw error;
   }
 };
@@ -299,9 +297,7 @@ export default {
           }
           // Other errors (timeout, not found, etc.) are still thrown
           error.workflowId = workflowId;
-          if ( runId ) {
-            error.runId = runId;
-          }
+          error.runId = runId;
           throw error;
         }
       },
--- a/api/src/clients/temporal_client.js
+++ b/api/src/clients/temporal_client.js
@@ -1,6 +1,6 @@
 import { Client, Connection, defaultPayloadConverter } from '@temporalio/client';
 import { temporal as temporalConfig } from '#configs';
-import { buildWorkflowId, extractErrorDetail, extractErrorMessage, takeFromAsyncIterable } from '#utils';
+import { buildWorkflowId, extractErrorDetail, extractErrorMessage, extractFailure, serializeTemporalError, takeFromAsyncIterable } from '#utils';
 import { logger } from '#logger';
 import {
   WorkflowNotFoundError,
@@ -41,6 +41,12 @@ const getCatalog = async ( { client, taskQueue } ) => {
     if ( error instanceof WorkflowNotFoundError ) {
       throw new CatalogNotAvailableError( 3 );
     }
+    // Annotate context the error_handler's serializer can't recover on its own, then log the full
+    // nested Temporal/gRPC detail here — the only place that knows the task queue + query name.
+    error.taskQueue = taskQueue;
+    error.query = 'get';
+    logger.error( 'Catalog query failed', { taskQueue, query: 'get', ...serializeTemporalError( error ) } );
+    error.alreadyLogged = true;
     throw error;
   }
 };
@@ -167,6 +173,7 @@ export const extractWorkflowInput = history => {
  * @property {string|number|boolean|object|array|undefined|null} output - The workflow output
  * @property {object|null} trace - Trace information
  * @property {string|null} error - Error message if failed
+ * @property {object|null} failure - Structured failure details if failed (message, type, retryable, cause), null otherwise
  */
 
 /**
@@ -199,9 +206,11 @@ const buildWorkflowResult = ( { workflowId, status, runId, input, result, error 
     ...( error ? {
       trace: extractErrorDetail( error, 'trace' ),
       aggregations: extractErrorDetail( error, 'aggregations' ),
-      error: extractErrorMessage( error )
+      error: extractErrorMessage( error ),
+      failure: extractFailure( error )
     } : {
-      error: null
+      error: null,
+      failure: null
     } )
   } );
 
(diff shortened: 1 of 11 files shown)


api/src/clients/temporal_client.js
```diff
@@ -290,6 +299,9 @@ export default {
           }
           // Other errors (timeout, not found, etc.) are still thrown
           error.workflowId = workflowId;
+          if ( runId ) {
```
szanata: In which scenario runId wouldnt exist?

api/src/clients/temporal_client.js
```diff
@@ -41,6 +41,12 @@ const getCatalog = async ( { client, taskQueue } ) => {
     if ( error instanceof WorkflowNotFoundError ) {
       throw new CatalogNotAvailableError( 3 );
     }
+    // Annotate context the error_handler's serializer can't recover on its own, then log the full
+    // nested Temporal/gRPC detail here — the only place that knows the task queue + query name.
+    error.taskQueue = taskQueue;
+    error.query = 'get';
+    logger.error( 'Catalog query failed', { taskQueue, query: 'get', ...serializeTemporalError( error ) } );
+    error.alreadyLogged = true;
```
szanata: This seems brittle... a flag on the error to tell if the error was logged doesnt seem a good solution

api/src/utils.js
```diff
@@ -67,6 +67,150 @@ export const extractErrorDetail = ( e, key ) =>
...
+/**
+ * Duck-types a Temporal ApplicationFailure node. The worker interceptor wraps user errors as
+ * ApplicationFailure( message, type = constructorName, nonRetryable, details, cause ); only that
+ * node carries .type / .nonRetryable. We duck-type to keep this module free of @temporalio imports.
+ *
+ * @param {unknown} e
+ * @returns {boolean}
+ */
+const isApplicationFailure = e =>
+  Boolean( e ) && ( typeof e.type === 'string' || typeof e.nonRetryable === 'boolean' );
```
szanata: `Boolean( e )` although correct, is not usual at all in js.

api/src/utils.js
```diff
@@ -67,6 +67,150 @@ export const extractErrorDetail = ( e, key ) =>
...
+  if ( !error ) {
+    return null;
+  }
+  const appFailure = findApplicationFailure( error );
+  return {
+    message: appFailure?.message ?? extractErrorMessage( error ),
+    type: appFailure?.type ?? error.constructor?.name ?? error.name ?? null,
+    retryable: appFailure && typeof appFailure.nonRetryable === 'boolean' ? !appFailure.nonRetryable : null,
+    cause: serializeErrorChain( error )
+  };
```
szanata, line 181: why this serialization is not part of the serializeErrorChain in the first place?

szanata, line 181: `type` should be `name`

api/src/clients/temporal_client.js
```diff
@@ -199,9 +206,11 @@ const buildWorkflowResult = ( { workflowId, status, runId, input, result, error
     ...( error ? {
       trace: extractErrorDetail( error, 'trace' ),
       aggregations: extractErrorDetail( error, 'aggregations' ),
-      error: extractErrorMessage( error )
+      error: extractErrorMessage( error ),
+      failure: extractFailure( error )
```
szanata: I disagree with this. error and failure are the same thing.

I would vote to keep just `error` which is now and object.
People can check the message under `error.message`

api/src/utils.js
```diff
@@ -67,6 +67,150 @@ export const extractErrorDetail = ( e, key ) =>
...
+ *
+ * @param {Error} e
+ * @param {number} [depth=20]
+ * @returns {object|null}
+ */
+const findApplicationFailure = ( e, depth = 20 ) => {
+  if ( !e || depth <= 0 ) {
+    return null;
+  }
+  return isApplicationFailure( e ) ? e : findApplicationFailure( e.cause, depth - 1 );
```
szanata: I dont see any reason for isApplicationFailure to be its won method, this only adds noise, this could be simple enough if it was: _Recursively checks the error .cause chain and looks for nonRetriable errors`. This function is not being exported by itself anyway.

szanata (changes requested):
I left some comments about changes. Right now I don't think this should be added to the code without some refinements.

api/src/utils.js
```diff
@@ -67,6 +67,150 @@ export const extractErrorDetail = ( e, key ) =>
...
+      code: error.code,
+      codeName: GRPC_STATUS_NAMES[error.code] ?? 'UNKNOWN',
+      details: error.details,
+      metadata: redactMetadata( error.metadata )
+    } : {} ),
+    ...( error.workflowId ? { workflowId: error.workflowId } : {} ),
+    ...( error.runId ? { runId: error.runId } : {} ),
+    ...( error.taskQueue ? { taskQueue: error.taskQueue } : {} ),
+    ...( error.query ? { query: error.query } : {} ),
+    ...( error.cause ? { cause: serializeTemporalError( error.cause, depth + 1 ) } : {} )
```
bnchrch: Done — switched these to `...( cond && { ... } )`. (84e8c9f)

api/src/clients/temporal_client.js
bnchrch: I would prefer to keep both, and I think it is worth a quick chat. `error` is a **string** in the shipped API (0.6.0) and is read as one by Atlas/Flow Studio and our CLI, so turning it into an object is a breaking change. OUT-233 was scoped as additive (keep `error`, add structured detail), so I kept `error` (string) and added `failure` (object) alongside it. Happy to collapse to a single structured field in a future major bump if we want to take the break — I just did not want to break consumers in a patch. WDYT?

szanata: I think we could change this a minor change (breaking change) and make a single field. But we can chat about first. Thanks for addressing

bnchrch:
Follow-up (5146ed2): renamed the structured object `failure` → **`errorDetails`** (JSDoc type `WorkflowErrorDetails`), and added **`errorDetails.activityId`** = the failing activity key (`"{workflow}#{step}"`, e.g. `test_error#thrower`; `null` for workflow-level/child-workflow failures). OpenAPI + the generated CLI client are regenerated.

This also sharpens the `error` (string summary) vs `errorDetails` (structured) split — re: the thread about the two fields being redundant.

catsby (approved):

szanata (approved):

PR #253, opened 2026-06-10
refactor(llm): registry-driven prompt block parsing

bnchrch:
## Problem

Adding a new prompt block role or block option meant editing the grammar in three places, because roles and attributes were hardcoded:

- Block roles lived in the tokenizer's regex alternation (`(system|user|assistant|tool)`)
- Each attribute had its own bespoke regex in `parseBlockAttributes` (the source of the `options="cache"` false-positive footgun)
- The `.strict()` message schema enumerated allowed fields, and resolution branched per option

A general HTML/XML parser isn't a safe swap here: `escape.js` only escapes Liquid *variable* output, so static prompt bodies legitimately contain arbitrary `<...>` markup a markup parser would misparse — and it would own entity decoding, colliding with the existing escape/`decode()` contract.

## Solution

Separate **tokenizing** from **interpreting**, and make both data-driven:

- **`prompt/blocks.js`** — a focused tokenizer. Roles come from a `BLOCK_ROLES` set (the pattern is generated from it); one generic attribute parser turns the opening tag into `{ key: value | true }`. Content between role tags stays opaque, preserving the escape/decode contract.
- **`prompt/block_options.js`** — one registry where each option declares its `schema` *and* its `resolve`. `attributesSchema` (validation) and `resolveMessageProviderOptions` (AI-SDK boundary) both derive from this table.
- `parser.js` drops to frontmatter + `tokenizeBlocks`; validation validates a generic `attributes` object; `PromptMessage` carries `attributes` instead of `cache`/`options`.

Net effect: a new block role is one line in `BLOCK_ROLES`; a new option is one entry in the registry — no regex, no schema surgery, no resolver branching. `cache` and `options` from #252 are now the first two registry entries.

## Test plan

- [x] `npm run lint` passes
- [x] `npm test` — 2055 pass; only the 2 pre-existing `sdk/cli` copy-assets failures remain (unrelated; see #252)
- [x] New unit specs: `blocks.spec.js` (tokenizer, incl. opaque angle-bracket content + the `options="cache"` guard) and `block_options.spec.js` (registry schema + resolver); `ai_sdk_options.spec.js` keeps one integration test
- [x] Public `.d.ts` typechecks against the new `attributes` shape

## Notes for reviewers

- **Kept a (registry-generated) regex tokenizer rather than adding a parser dependency** — the body grammar is tiny and the content-safety constraint above makes a general markup parser a poor fit. The `tokenizeBlocks` interface (`{ role, attributes, content }`) lets us swap in a hand scanner later with no downstream change.
- **`PromptMessage` now exposes `attributes` instead of `cache`/`options`** — a shape change to a type only introduced in #252 (unreleased), so no migration. No new changeset for the same reason: this refactors unreleased code and the #252 changeset covers the release.

Builds on #252 (stacked — base retargets to `main` once #252 merges).

### DIFF
--- a/.changeset/prompt-caching-support.md
+++ b/.changeset/prompt-caching-support.md
@@ -1,0 +1,9 @@
+---
+"@outputai/llm": minor
+---
+
+Add per-message prompt caching to `.prompt` files.
+
+- `cache` shorthand on a block (`<system cache>`, `<system cache="1h">`) sets an Anthropic cache breakpoint covering the prompt prefix up to and including that block.
+- General `messageOptions` named sets in front matter, attached to blocks via `options="<name>"`, attach arbitrary per-message `providerOptions`.
+- `cache` resolves to `anthropic.cacheControl` for `provider: anthropic` and Claude models on `vertex`; for other providers it is ignored with a warning (use `messageOptions` with that provider's namespace).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -226,6 +226,49 @@ AI SDK uses `Record<string, Record<string, JSONValue>>` for `providerOptions` to
 
 The nesting is intentional architecture, not redundancy.
 
+### Per-Message Caching (Anthropic Prompt Cache)
+
+Anthropic prompt caching is a **per-message** directive, not a call-level one. Mark the block that ends your static prefix; that prefix is then cached and reused across calls.
+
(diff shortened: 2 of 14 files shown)


bnchrch:
Folded into #252 — consolidating into a single PR. The registry-based parsing now ships as part of the feature commit there.

PR #255, opened 2026-06-11
feat(cli): compute workflow cost from trace events with costs.yml override

bnchrch:
## Context

The `output workflow cost` command **ignored the cost values recorded in the trace** and re-derived every cost from `costs.yml` token/service rates. On real traces this produced wrong results:

- **`gpt-5.5` → $0** — no `costs.yml` entry, flagged "unknown model" and priced at $0 despite the trace recording $1.52 of actual spend.
- **`claude-opus-4-8` → $2.56 instead of $0.85** — no exact entry, so the prefix matcher fell back to `claude-opus-4` (15/75), 3× the rate actually charged (5/25).
- **HTTP costs → $0** — `api.exa.ai` / `api.firecrawl.dev` never appeared (exa needs a response body the trace lacks; firecrawl wasn't configured), despite $0.79 of recorded HTTP spend.

Every trace node already carries its as-charged cost: `llm` nodes in `attributes["llm:usage"]` (per-token-type `ppm`/`amount`/`total` + a top-level `total`) and billable `http` nodes in `attributes["http:request:cost"].total`.

## What this does

Makes the trace **events the source of truth** (the *Original* / as-charged cost) and applies `costs.yml` as an optional **override** layer (the *Adjusted* cost), showing **both** per model and per host. The bottom-line/JSON total is the Adjusted figure, with the as-charged total shown alongside.

- `findLLMCalls` reads `attributes['llm:usage']` (model, token amounts, as-charged total), falling back to legacy `output.usage`.
- `findHTTPCalls` captures host + as-charged cost from `attributes['http:request:cost']`.
- **LLM adjusted** = re-price the event tokens via `costs.yml` when the model is configured (reuses `calculateLLMCallCost`), else equals original.
- **HTTP adjusted** = `costs.yml` service recompute when usable (reuses `calculateServiceCost`), else falls back to the as-charged cost.
- HTTP grouped by host; dedup unchanged (`seenIds`, first-occurrence by node `id`).
- Formatter gains `Original` + `Adjusted` columns and a two-row total (Adjusted headline + As-charged); the unknown-model **error** is now an informational note.

### Output (real trace)
image

## Backward compatibility

Traces without `attributes` events fall back to the old `output.usage` + `costs.yml` path (there Original == Adjusted). `--format json` consumers gain `originalCost`/`adjustedCost`; `totalCost` remains (now the adjusted headline).

## Verification

- End-to-end on a real 126MB trace reproduces the manual figures exactly: LLM **$15.32**, HTTP **$0.79**, as-charged total **$16.11**; `gpt-5.5` and `api.exa.ai`/`api.firecrawl.dev` are no longer $0.
- First-occurrence dedup confirmed to yield exactly 623 LLM / 260 HTTP on the real trace.
- `cost_calculator.spec.ts` rewritten for the event-driven model (cases: rate-matches, prefix-mismatch override, no-override model, dedup, HTTP recompute fallback, request-based override, count-only exclusion).
- 49/49 cost tests pass; ESLint clean; `tsc --noEmit` clean for all cost files.

> Note: a full `npm run build:packages` is not a clean signal in this environment — the CLI `tsc` build trips on pre-existing missing optional peer deps (`@inquirer/prompts`, `@anthropic-ai/claude-agent-sdk`) unrelated to this change.


szanata:
Approved, this PR looks good. However two things:
1. I dont think we need to keep covering the old use case, attributes were inserted many versions ago
2. The trace file parsing still tries to parse fields that dont exist anymore, it was not introduced in this PR but it is something to keep an eye on.

szanata (approved):

bnchrch:
Thanks for the review @szanata — both points addressed in 308274e5f:

1. **Legacy support removed.** `findLLMCalls` now only matches nodes carrying an `llm:usage` attribute, and `aggregateHTTPCosts` only bills calls with an `http:request:cost` attribute. The eventless fallbacks (`output.usage`/`loadedPrompt` extraction, the costs.yml-driven HTTP branch with its `response_cost` billability rules, and `calculateLLMCallCost` itself) are deleted. Traces from SDKs predating cost attributes now report no costs, and eventless calls (e.g. clients that don't call `addRequestCost`) are excluded by policy — on the reference trace that only drops five $0 tavily rows; totals are unchanged.

2. **On the stale-field parsing:** the LLM-side offenders (`input.loadedPrompt.config.model`, `output.model`, `output.usage`) went away with point 1. The body-dependent costs.yml service rules (exa `costDollars`, jina token counts, per-item counts) are intentionally kept: body capture into traces is gated by `OUTPUT_TRACE_HTTP_VERBOSE`, which `docker-compose-dev` defaults to `true` — so they fire on local dev traces and fall back to the as-charged event cost on production traces. Documented on `calculateServiceCost`.

PR #259, opened 2026-06-12
chore: CLI performance updates

catsby:
## Summary

User @derencius shared some insight on some performance issues with the CLI (and a minor API one). I was able to verify the issues and this PR should clean things up and improve performance in some operations. 

Of note: it seems like the CLI was already setup to include `oclif.manifest.json`, which alone can greatly improve performance/response time with CLI commands, but I suppose we never introduced the actual file, so we fix that here.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -32,6 +32,7 @@ minimumReleaseAgeExclude:
 
 overrides:
   '@anthropic-ai/sdk@>=0.79.0 <0.91.1': '>=0.91.1'
+  '@types/node@<24': '>=24.10.15'
   '@hono/node-server@<1.19.13': '>=1.19.13'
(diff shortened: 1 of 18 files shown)


pnpm-workspace.yaml
```diff
@@ -32,6 +32,7 @@ minimumReleaseAgeExclude:
 
 overrides:
   '@anthropic-ai/sdk@>=0.79.0 <0.91.1': '>=0.91.1'
+  '@types/node@<24': '>=24.10.15'
```
szanata, line 35: Why this?

catsby: That came along with `oclif` to the `devDependencies`, having to do with it's transitive deps and node versions.

szanata (approved):

PR #260, opened 2026-06-12
refactor(core): Improving child workflow invocation and dropping aggregations [OUT-485]

szanata:
- Refactoring child workflow invocation:
  - Workflow calls no longer need to be rewritten by the webpack loader to wrap with `this.startWorkflow()`;
  - Any workflow invocation within the triggered inside an workflow will become a child

- Dropping workflow activity aggregations: Aggregations had flaws with continue as new and detached child workflows, the whole feature is being dropped in favor of improving hooks. Activity aggregations are kept, as that can be retrieved using the history API;

- Refactoring validations, simplifying code and adding runtime validation for workflow invocation options;

- Renaming type `WorkflowInvocationConfiguration` to `WorkflowInvocationOptions` and update the shape inside renaming `.options` to `.activityOptions`;

- Reorganized test workflows and added more tests covering nested (children) workflows common cases;


sdk/core/src/interface/workflow.js
```diff
@@ -1,202 +1,94 @@
...
+    const isRoot = !root;
 
-    /**
-     * @IMPORTANT Keep support for deprecated add_attribute Signal to allow for Temporal replays.
-     * @TODO This can be removed 30days after this release
-     */
-    setHandler( defineSignal( 'add_attribute' ), attribute => {
-      mergeAggregationsInWorkflowResult( aggregateAttributes( [ attribute ] ) );
-    } );
+    memo.stack = [ ...memo.stack ?? [], workflowId ];
```
bnchrch, line 72: 💎 Really like moving the "am I nested?" decision to a runtime `memo.stack` check instead of the build-time import rewriting. It kills the old "call silently runs inline instead of as a child" footgun and let us delete a pile of npm-resolution code. Nice one.

---
_Generated by [Claude Code](https://claude.ai/code)_

sdk/core/src/interface/workflow.js
```diff
@@ -1,202 +1,94 @@
...
-        invokeStep: async ( stepName, input, options ) =>
-          callStepAndUnwrapResult( steps[`${name}#${stepName}`], input, options ),
-        invokeSharedStep: async ( stepName, input, options ) =>
-          callStepAndUnwrapResult( steps[`${SHARED_STEP_PREFIX}#${stepName}`], input, options ),
-        invokeEvaluator: async ( evaluatorName, input, options ) =>
-          callStepAndUnwrapResult( steps[`${name}#${evaluatorName}`], input, options ),
-        invokeSharedEvaluator: async ( evaluatorName, input, options ) =>
-          callStepAndUnwrapResult( steps[`${SHARED_STEP_PREFIX}#${evaluatorName}`], input, options ),
+      // Creates an activity caller based on a prefix
+      const createCaller = prefix => async ( t, ...args ) => ( await steps[`${prefix}#${t}`]( ...args ) ).output;
```
bnchrch: ❗ 🔮 I think this breaks replay for in-flight workflows. `createCaller` here (and the trace fetch on L61) now read `.output` with no fallback, and the `SEND_AGGREGATIONS` / `add_attribute` handlers are gone too. All of those had "keep for replays" guards before (`isActivityResultWrapped`, the `OUT-468` note).

The worry: a workflow that started before the activity-wrapper migration and is still alive (a long runner, a continue-as-new, or one that just got evicted) replays its old non-wrapped result, `.output` comes back `undefined`, and we either return garbage or trip a non-determinism error.

❓ Are we confident no pre-`OUT-468` runs survive this deploy? If we can't promise that, could we gate it with `patched()` so old histories take the old branch and new ones take this one?

Please read this in a friendly tone. Also 🔮 if this can wedge a running workflow, the changeset probably wants to be more than `minor`.

---
_Generated by [Claude Code](https://claude.ai/code)_

sdk/core/src/interface/workflow.js
```diff
@@ -1,202 +1,94 @@
...
-      throw e;
+      return {
+        [WORKFLOW_WRAPPER_VERSION_FIELD]: 1,
+        output,
+        ...( traceDest && { trace: { destinations: traceDest } } )
+      };
+    } catch ( error ) {
+      if ( isRoot && traceDest ) {
+        // Append the trace destinations so it is carried to interceptor
+        error[METADATA_ACCESS_SYMBOL] = { trace: { destinations: traceDest } };
```
bnchrch, line 109: ❗ This clobbers any metadata already on the error. The old version spread it in first, so anything attached upstream is now lost.
```suggestion
        error[METADATA_ACCESS_SYMBOL] = { ...( error[METADATA_ACCESS_SYMBOL] ?? {} ), trace: { destinations: traceDest } };
```

---
_Generated by [Claude Code](https://claude.ai/code)_

sdk/core/src/worker/interceptors/activity.js
```diff
@@ -28,7 +25,7 @@ const log = createChildLogger( 'ActivityInterceptor' );
   - Headers injected by the workflow interceptor
 */
 export class ActivityExecutionInterceptor {
-  constructor( { activities, workflows, connection } ) {
+  constructor( { activities, workflows } ) {
```
bnchrch, line 28: 💅 Heads up, now that this constructor only takes `{ activities, workflows }`, the caller in `interceptors/index.js:11` still passes `connection`, so it's quietly dropped on the floor. Worth cleaning up the call (and the `initInterceptors` signature) in the same pass. To be clear this is a nit, feel free to disregard.

---
_Generated by [Claude Code](https://claude.ai/code)_

sdk/core/src/interface/workflow.spec.js
```diff
@@ -1,684 +1,626 @@
...
+        info: { workflowId: 'test-workflow', runId: 'test-run' }
+      } ) );
+      expect( validateOutputMock ).toHaveBeenCalledWith( output );
     } );
+  } );
 
-    it( 'merges extra.context into context when provided', async () => {
-      inWorkflowContextMock.mockReturnValue( false );
+  describe( 'child workflow trigger path', () => {
+    it( 'starts a child workflow when memo.stack already contains the current workflowId', async () => {
```
bnchrch, line 252: 🧪 Nice coverage on the direct case here. The one path I don't see a unit test for is the headline behavior: a workflow called through a helper defined outside the handler still becoming a child. That only has integration coverage over in `test_workflows/nested/*`. A unit test that drives the `memo.stack` path from an indirect call would catch a regression without standing up Temporal. WDYT?

---
_Generated by [Claude Code](https://claude.ai/code)_

bnchrch (commented):
Start in `sdk/core/src/interface/workflow.js`. That's the heart of this and it reads really well. Most of the diff is the `test_workflows` reorg and the `pnpm-lock.yaml` churn, so I'd skip those on a first pass. One blocking item I'd love to sort out before we merge (the replay-compat removals), plus a couple of smaller things below.

---
_Generated by [Claude Code](https://claude.ai/code)_

sdk/core/src/interface/workflow.spec.js
szanata: [1d95112](https://github.com/growthxai/output/pull/260/commits/1d951123e6dfe137757de691429649dc43d32640)

sdk/core/src/worker/interceptors/activity.js
szanata: [b46a102](https://github.com/growthxai/output/pull/260/commits/b46a1024bd66743a15ee64bce2f23c6c139abd58)

sdk/core/src/interface/workflow.js
szanata: I think it is ok. Only that part of the code assigns it. I think it makes less open to interpretations, like "what else is there"

sdk/core/src/interface/workflow.js
```diff
@@ -1,202 +1,94 @@
...
-    // Creates the result wrapper with information about the workflow
-    const workflowResult = {
-      [WORKFLOW_WRAPPER_VERSION_FIELD]: 1,
-      aggregations: null,
-      ...( isRoot && {
-        trace: {
-          destinations: await getTraceDestinations()
+    // if the stack already includes this workflowId, means the workflow() function was called
+    // from within a running workflow, meaning it is suppose to start a child workflow
+    const isChild = memo?.stack?.includes( workflowId );
```
catsby: does continue-as-new inherit this `memo` and will it have the same `workflowId` (just with a different `runId`)? Does that complicate things here or is that case covered?

szanata: Continue as new will have the same memo up until the parent.

So all children that continued as new will be detected as... children, so each one will have their own copy of the memo and will add their own `workflowId` (which is the same), so the branching stills works correctly, as the memo is immutable across executions and it is propagated parent -> child.

this scenario was tested and I added workflow examples as well.

sdk/core/src/interface/workflow.js
szanata: I am torn on this. But since you mentioned, I added backwards compatibility. [befe2bc](https://github.com/growthxai/output/pull/260/commits/befe2bc9e5600fd673dc20e5752eb3db31a50b75)

catsby (approved):

PR #262, opened 2026-06-15
refactor(api): Reorganizing temporal client

szanata:
## Summary

Reorganizing API Temporal Client code base, so it is easier to read and better to test.

## Test plan

- [x] Unit test
- [x] Manual tests


api/src/clients/temporal/workflow/get_result.js
```diff
@@ -0,0 +1,99 @@
...
+      throw new WorkflowNotFoundError( `Run "${resolvedRunId}" not found for workflow "${workflowId}"` );
+    }
+    throw error;
+  } );
+
+  if ( !firstPage.history ) {
+    logger.warn( 'Temporal getWorkflowExecutionHistory returned no history field', { workflowId, runId: resolvedRunId } );
+  }
+
+  const status = description.status.name?.toLocaleLowerCase();
```
bnchrch: 💅 This uses `toLocaleLowerCase()` while the other three status paths (`get_history`, `get_status`, `list_runs`) use plain `toLowerCase()`. Locale-dependent casing can bite (the Turkish dotless-i thing) and this is a standardized API value. Since `types.js` already centralizes status concerns (`WorkflowStatus`, `isWorkflowClosed`), what about one helper there and reusing it in all four?

```js
// types.js
export const formatStatus = name => name?.toLowerCase() ?? 'unspecified';
```

Kills the inconsistency and the duplication in one go.

api/src/clients/temporal/catalog.js
```diff
@@ -0,0 +1,42 @@
...
+ * @param {string} taskQueue - The task queue (for error messages)
+ * @returns {string} The canonical workflow name
+ * @throws {WorkflowNotFoundError}
+ */
+export const resolveWorkflowName = ( catalog, workflowName, taskQueue ) => {
+  const resolved = catalog.workflows.find( w => w.name === workflowName || w.aliases?.includes( workflowName ) );
+  if ( !resolved ) {
+    throw new WorkflowNotFoundError( `Workflow "${workflowName}" is not available at worker "${taskQueue}"` );
+  }
+  return resolved.name;
```
bnchrch, line 45: 🔮❓ Looks like the `'Workflow alias resolved'` info log got dropped here (the old `resolveWorkflowName` logged `{ alias, resolvedName, taskQueue }` when an alias differed from the canonical name). Intentional? It was handy for spotting which aliases are in use in prod. Happy to keep it gone if it was noisy, just flagging.

api/src/clients/temporal/workflow/get_status.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    const getHandle = vi.fn().mockReturnValue( { describe } );
+    const client = { workflow: { getHandle } };
+
+    const result = await getStatus( { client }, 'workflow-id', 'run-id' );
+
+    expect( getHandle ).toHaveBeenCalledWith( 'workflow-id', 'run-id' );
+    expect( result ).toEqual( {
+      workflowId: 'workflow-id',
+      runId: 'run-id',
+      status: 'completed',
```
bnchrch: 🧪 Only `get_result` asserts the `continued_as_new` mapping. Worth a one-liner here and in `list_runs`/`get_history` specs too, since the rename is the whole point of the PR and these paths each format status independently right now.

bnchrch (approved):
pre-approved, small ntis.

api/src/clients/temporal/catalog.js
szanata: Yep. Added it back.

api/src/clients/temporal/workflow/get_result.js
szanata: Makese sense

api/src/clients/temporal/workflow/get_status.spec.js
szanata: It was updated with prior changes.

PR #263, opened 2026-06-15
feat(core,cli): add opt-in workflow bundle check (OUT-474)

bnchrch:
## Problem

- A Temporal worker **crash-loops** when `workflow.ts` — or anything it transitively imports — pulls in a `node:` built-in (e.g. `node:fs`); it can't exist in the workflow V8 bundle.
- The bundle only runs **at worker startup**, so `tsc --noEmit` never catches it. A recent incident (a transitive `node:fs`) crash-looped the worker until a fix was deployed.

## Solution

- **core** — `output-worker --check` (or `OUTPUT_WORKER_CHECK=1`) bundles workflows via the same `bundleWorkflowCode` path `Worker.create` uses, then exits 0/1 with the offending module named. `bin/worker.sh` routes it to a config-free `check.js` — no Temporal connection, no worker env.
- **core** — new `@outputai/core/worker` export (`webpackConfigHook`, `loadWorkflows`, `createWorkflowsEntryPoint`, `bundleWorkflows`) for building custom checks.
- **core** — worker and check share one `workflowInterceptorModules` source; `bundle.spec.js` asserts parity so they can't drift.
- **cli** — scaffold + the `fix_package` migration gain an opt-in `output:worker:check` script; the generated README documents it with a copy-paste CI snippet. **Not** wired into any build.

## Interface

```bash
npm run output:worker:check     # = output-worker --check  (run after building dist/)
```
```js
import { bundleWorkflows } from '@outputai/core/worker';
await bundleWorkflows( process.cwd() );  // rejects if a workflow can't be bundled
```

## Test plan

- [x] `npm run lint` passes
- [x] `npm test` — 2092 pass incl. new parity test `bundle.spec.js`; the only 2 failures are pre-existing `copy_assets` (unbuilt `dist/templates`, see Follow-up)
- [x] Happy path: `output:worker:check` on `test_workflows` (33 workflows) → pass, ~2.6s, no Temporal/worker env
- [x] Failure path: a transitive `import 'node:fs'` → `tsc` passes, check fails (`UnhandledSchemeError: node:fs`), exit 1
- [x] `@outputai/core/worker` import is pure + env-free (does not start a worker)

## Notes for reviewers

- **`--check` routes to a dedicated `check.js`, not a branch in `index.js`** — `index.js` imports `configs.js`, which eagerly validates `OUTPUT_CATALOG_ID` (required, no default). A branch there would force full worker env on every check; the separate entry keeps it runnable in a bare CI.
- **`workflowInterceptorModules` moved to a config-free `interceptors/modules.js`** — importing it via `interceptors/index.js` pulls `activity.js` → `configs.js` (same env validation). `index.js` re-exports it, so parity holds.
- **`bundleWorkflows` calls `loadActivities`** though it only bundles — that step writes `temp/__activity_options.js`, which the workflow interceptor module imports during bundling (mirrors the worker's startup order).
- **Reused `bundleWorkflowCode` over a static AST scan** — it's the exact bundler `Worker.create` runs, so the check can't diverge from real worker behavior, and it adds no new dependency.

## Follow-up (out of scope)

`sdk/cli`'s `copyassets.sh` fails on Node 24 (`native-copyfiles` passes an array to `globSync`'s `exclude`, which now must be a function), so `build:packages` and `copy_assets.spec.ts` fail locally. Pre-existing, unrelated to this PR — worth a separate ticket.

Closes OUT-474


sdk/core/src/worker/bundle.js
```diff
@@ -0,0 +1,26 @@
...
+ * transitive import graph.
+ *
+ * @param {string} rootDir directory to discover workflows in
+ * @returns {Promise<import('@temporalio/worker').WorkflowBundleWithSourceMap>}
+ */
+export async function bundleWorkflows( rootDir ) {
+  const workflows = await loadWorkflows( rootDir );
+  // Writes worker/temp/__activity_options.js, which the workflow interceptor module
+  // imports — the worker generates it via loadActivities() before Worker.create bundles.
+  await loadActivities( rootDir, workflows );
```
szanata, line 23: Why activities are needed here? Since this is only to check worfklows... this doesnt seem necessary.

sdk/core/src/worker/bundle.js
```diff
@@ -0,0 +1,26 @@
...
+  const workflowsPath = createWorkflowsEntryPoint( workflows );
+  return bundleWorkflowCode( { workflowsPath, workflowInterceptorModules, webpackConfigHook } );
```
szanata, line 25: The same goes for interceptors. They actual interceptors are not necessary....

sdk/core/src/worker/check.js
```diff
@@ -0,0 +1,20 @@
...
+
+( async () => {
+  log.info( 'Checking workflow bundle...', { callerDir } );
+  await bundleWorkflows( callerDir );
+  log.info( '✓ Workflow bundle check passed' );
+  process.exit( 0 );
+} )().catch( error => {
+  log.error( '✗ Workflow bundle check failed', { message: error?.message, stack: error?.stack } );
+  process.exit( 1 );
+} );
```
szanata: Is is devised to run via cmd to test dry run build to detect errors. I dont think logger is necessary, the output doesnt need to be formatted. The output needs to be 0 (ok) or 1 (fail).

If the error output is necessary, that could be a simple console.error( error ).

The logger will unnecessary format these messages.

I think adding other outputs make this hard to integrate of cli, terminal, etc.

sdk/core/src/worker/bundle.js
bnchrch: Intentional — the check is meant to be a faithful dry-run of what `Worker.create` bundles at startup, so it fails on exactly what the worker would. `Worker.create` bundles the workflows **and** the workflow-side interceptor modules (`initInterceptors().workflowModules`), so the check passes the same `workflowInterceptorModules`. `bundle.spec.js` pins these to the worker's so the two can't silently drift.

If we'd rather scope the check to just user-workflow code, that's simpler but drops the worker-boot guarantee — happy to switch if you'd prefer.

sdk/core/src/worker/bundle.js
bnchrch: Tied to the interceptor modules: the workflow-side interceptor (`interceptors/workflow.js`) statically imports `../temp/__activity_options.js`, which `loadActivities` generates. Without it, bundling the interceptor fails with a module-not-found (a false failure) instead of a real result — so the check mirrors the worker's prep order (`loadWorkflows` → `loadActivities` → `createWorkflowsEntryPoint` → bundle).

sdk/core/src/worker/check.js
bnchrch: Good call — done in 42842b3b1. `check.js` no longer uses the logger: exit 0 on success, `console.error(error)` + exit 1 on failure, so the exit code is the signal.

One note: the remaining `Scanner: …` lines come from the shared `loadWorkflows`/`loadActivities` (reused for worker parity), not from `check.js`. Their winston level is hardcoded, so fully silencing stdout would mean changing shared worker logging — happy to do that as a follow-up if you want exit-code-only output.

sdk/core/src/worker/public.js
```diff
@@ -0,0 +1,3 @@
+export { webpackConfigHook } from './bundler_options.js';
```
szanata: Why are we exposing all these if we only need `bundleWorkflows()`? Actually, why we need this file at all, could we just expose `check.js` directly? I think the least we expose the better, as this creates dependencies elsewhere and makes the code harder to change. I am agains exposing `loader` and `bundle_options` at all.

sdk/core/src/worker/check.js
szanata: You could run the `bundleWorkflows` with `spawnChild` and mute stdout, keeping only stderr. So if the code breaks you have the error to propagate. So you can avoid changing the log system.

szanata: Not a blocker thou

sdk/core/src/worker/interceptors/index.js
```diff
@@ -1,11 +1,10 @@
-import { dirname, join } from 'path';
-import { fileURLToPath } from 'node:url';
 import { ActivityExecutionInterceptor } from './activity.js';
+import { workflowInterceptorModules } from './modules.js';
 
-const __dirname = dirname( fileURLToPath( import.meta.url ) );
+export { workflowInterceptorModules };
```
szanata: These can be combined into `export { workflowInterceptorModules } from './modules.js';` as export also has local visibility

sdk/core/src/worker/bundle.js
szanata: Makes sense. I can leave like this, or we can create the file manually, since we are trying to validate only workflow code. I leave that to you.

sdk/core/src/worker/public.js
bnchrch: Trimmed in ea464d151 — `@outputai/core/worker` now exports only `bundleWorkflows`. It composes `webpackConfigHook` + `loadWorkflows`/`createWorkflowsEntryPoint` internally, so the surface is just the one entry point you'd actually call; `loader` and `bundler_options` are back to internal.

sdk/core/src/worker/check.js
bnchrch: Done in ea464d151 — muted stdout in `check.js` (kept stderr + exit code), so the discovery logs no longer hit stdout. Went in-process (override `process.stdout.write`) instead of a child process — same effect, one fewer moving part. Verified: happy path → clean stdout, exit 0; bad import → error on stderr, exit 1.

sdk/core/src/worker/interceptors/index.js
bnchrch: `export { workflowInterceptorModules } from './modules.js'` is a pure re-export — it wouldn't create a local binding, and `initInterceptors` uses `workflowInterceptorModules` in its body, so that form breaks. Since `modules.js` is the source and nothing else imported it from here, I dropped the re-export entirely instead (kept the plain `import` for local use). Net less surface — ea464d151.

sdk/core/src/worker/bundle.js
bnchrch: Thanks — keeping the full `loadActivities` call so the check stays a faithful mirror of the worker's startup; the manual-temp-file route would just duplicate that logic. 👍

.changeset/out-474-workflow-bundle-check.md
```diff
@@ -0,0 +1,9 @@
+---
+"@outputai/core": patch
+---
+
+Add an opt-in `output-worker --check` workflow bundle check (and a `@outputai/core/worker` export) that reproduces the worker's webpack bundling without a Temporal server, catching bad workflow imports — e.g. a transitive `node:` built-in — before they crash-loop the worker at startup. `tsc` cannot detect these; only the Temporal bundle can.
+
+- `output-worker --check` (or `OUTPUT_WORKER_CHECK=1`) bundles workflows via the same `bundleWorkflowCode` path `Worker.create` uses, exits non-zero with the offending module named, and needs no Temporal connection or worker env.
+- New `@outputai/core/worker` subpath exports `webpackConfigHook`, `loadWorkflows`, `createWorkflowsEntryPoint`, and `bundleWorkflows` for building custom checks.
```
szanata: This pharse is out of date.

sdk/core/bin/worker.sh
```diff
@@ -23,4 +23,10 @@ sdk_dir="$(dirname "$script_dir")"
 
 cd ${sdk_dir}
 
+# Route `output-worker --check` (or OUTPUT_WORKER_CHECK=1) to the bundle-check entry,
+# which validates the workflow bundle without a Temporal connection or worker config.
+if [[ " $* " == *" --check "* || "${OUTPUT_WORKER_CHECK}" == "1" || "${OUTPUT_WORKER_MODE}" == "check" ]]; then
```
szanata: Are we planning to use those env vars (OUTPUT_WORKER_CHECK/OUTPUT_WORKER_MODE)?

If they are not used in the command we should not expose them.

sdk/core/src/worker/public.js
szanata, line 1: I saw the you trimmed this file. I don't understand why we want to expose this function?

dont we need just a check? Or are we planning on surfacing another interface?

szanata: I saw this comment:

```
core — new @outputai/core/worker export (webpackConfigHook, loadWorkflows, createWorkflowsEntryPoint, bundleWorkflows) for building custom checks.
```

I dont necessarily agree. I think that exposing the `check` is enough for most use cases, we don't need to expose a bundle tool, at least not yet. I don't think it might not be useful in the future, but if there is not a use case now. I would avoid.

bnchrch: Agreed — dropped the `@outputai/core/worker` export entirely in 41c6b7649 (deleted `public.js` + `public.d.ts` + the `./worker` subpath). `output-worker --check` is the only surface now; `bundleWorkflows` stays internal (used by `check.js` + the parity test). We can add a programmatic export later if a real use case shows up.

sdk/core/bin/worker.sh
bnchrch: Good catch — dropped them in 41c6b7649. `worker.sh` now routes only on `--check`; `OUTPUT_WORKER_CHECK`/`OUTPUT_WORKER_MODE` are gone since nothing used them.

.changeset/out-474-workflow-bundle-check.md
bnchrch: Fixed in 41c6b7649 — the changeset now describes only the `output-worker --check` command (no export bullet, no `OUTPUT_WORKER_CHECK` mention).

szanata (approved):

PR #264, opened 2026-06-16
feat(api,core): Adding connection healthcheck [OUT-492]

szanata:
## Summary

Adding connection healthcheck on the API and Worker. If the connection with Temporal fails 3+ times in a row, it will shutdown the service gracefully.

### API
- Refactored lifecycle slightly so to integration the connection monitor properly.
- Added the `ConnectionMonitor` class that health check Temporal Connection every 1min. If it is not healthy or timeouts (5s) it counts a failure, after 3 failures it invokes the API shutdown (code: 1).
- Logs:
api

### Worker
- Worker is much more complex, and required full rewrite of the worker lifecycle.
- Added a `ConnectionMonitor` class that is cancellable (`.stop()`).
- Updated `startCatalog` into a class `CatalogJob` which is also cancellable (`.interrupt()`).
- Added wrap around `workflow.run()` so it does not throw error, instead store it.
- Added graceful shutdown when any of those 3 services above detects a failure.
- Added fallback shutdown via `.finally()` to ensure that even on unexpected errors, services are shutdown.
- Scenarios:
#### Interruption (SIGINT, SIGTERM):
interrruption

#### Boot error (Temporal down):
startup

#### Connection loss
connection_lost

#### CastalogJob error
catalog_job

#### Worker failure (temporal failure)
worker_failure

## Test plan

- [x] Added unit tests
- [x] Manually tested


api/src/index.js
```diff
@@ -42,7 +42,12 @@ const resolveCatalogField = ( body, route ) => {
 
 const app = express();
 
-const client = await temporalClient.init().catch( e => {
+const client = await temporalClient.init( {
+  onConnectionLost: e => {
+    logger.error( 'Temporal connection lost', { error: e.message, errorType: e.constructor.name, stack: e.stack } );
+    process.exit( 1 );
```
bnchrch: ❓ This skips the existing `shutdown()` at the bottom of the file, which drains in-flight HTTP requests via `server.close()` before exiting. Was that intentional? I get that Temporal is already dead at this point so there's nothing to close on that side, but requests that don't touch Temporal (like `/health`) would still benefit from a clean drain.

If the intent is "die fast, let the orchestrator restart," totally fine. Just want to make sure it's a conscious choice.

sdk/core/src/worker/index.js
```diff
@@ -88,7 +89,11 @@ const callerDir = process.argv[2];
   setupTelemetry( { worker } );
 
   log.info( 'Running worker...' );
-  await Promise.all( [ worker.run(), startCatalog( { connection, namespace, catalog, catalogHash } ) ] );
+  await Promise.all( [
+    worker.run(),
+    startCatalog( { connection, namespace, catalog, catalogHash } ),
+    setupConnectionMonitor( { connection } )
```
bnchrch: ❓ When this rejects (after 3 failures), `Promise.all` rejects, but `worker.run()` is still alive. Does the rejection bubble up to something that calls `worker.shutdown()`? If not, in-flight activities might not get a chance to finish their current task before the process dies.

bnchrch (approved):
LGTM!

sdk/core/src/worker/index.js
szanata: Yep. You are right. I will update it

api/src/index.js
szanata: Integrated!

sdk/core/src/worker/index.js
szanata: Integrated as well.

sdk/core/src/worker/connection_monitor.js
```diff
@@ -0,0 +1,113 @@
...
+    throw new Error( 'Connection health check timed out' );
+  } );
+
+  #healthcheck = async () => this.#connection.healthService.check( {} );
+
+  #sleep = async () => delay( this.#CHECK_INTERVAL_MS, 0, { ref: false } );
+
+  #watch = async () => {
+    try {
+      const health = await Promise.race( [ this.#healthcheck(), this.#getTimeout(), this.#cancellation.promise ] );
```
bnchrch: 💅 The API side uses `withDeadline`. Should we match it here in core?

Looks like that's the Temporal recommendation: without a deadline the health check can hang forever. Right now the `Promise.race` stops us waiting at 5s, but the actual call keeps running.

```js
#healthcheck = async () => {
  const deadline = Date.now() + this.#CHECK_TIMEOUT_MS;
  return this.#connection.withDeadline( deadline, () => this.#connection.healthService.check( {} ) );
};
```

Then `#getTimeout` goes away.

sdk/core/src/worker/connection_monitor.js
```diff
@@ -0,0 +1,113 @@
...
+      } else {
+        log.warn( 'Connection unhealthy', { error: error.message, failures: this.#failures } );
+      }
+    }
+
+    await Promise.race( [ this.#sleep(), this.#cancellation.promise ] );
+    if ( this.#cancellation.completed ) {
+      return true;
+    }
+    return this.#watch();
```
bnchrch: 💅 Nit, feel free to ignore. `#watch` calls itself each tick, so the pending promises stack up and never clear until shutdown. Slowly grows on a long-running process. (API's `#check` does the same.)

A loop avoids it:

```js
#watch = async () => {
  while ( !this.#cancellation.completed ) {
    // ...the existing check body, minus the trailing `return this.#watch()`...
    await Promise.race( [ this.#sleep(), this.#cancellation.promise ] );
  }
  return true;
};
```

bnchrch (commented):
Took a deeper pass since the LGTM. One question + a couple nits, nothing blocking.

❓ What happens when the process exits on connection loss?
- connection drops ~3 min
- we exit
- then what brings it back?

Temporal already reconnects on its own through blips, so I want to be sure exiting beats just waiting it out. As long as something restarts the process, makes sense. I just don't want a short hiccup to leave a worker dead until someone notices.

Also: API degrades first (`/ready` → 503), the worker just exits. Intentional?

sdk/core/src/worker/connection_monitor.js
szanata: We cant. I tried it. But the API uses `Connection` (temporal client) and the worker uses `NativeConnection` (temporal worker) and they `.withDeadline()` works differently. Basically the one from `NativeConnection` doesnt respect the deadline :X

sdk/core/src/worker/connection_monitor.js
szanata: I will take a look.

szanata:
@bnchrch 

>❓ What happens when the process exits on connection loss?

The orchestrator. If this is running in Render, the whole web worker dies and gets restarted. Same goes for docker-compose, Kubernetes etc, we exit with code 1, telling the job manager that this pod/container failed.

> Also: API degrades first (/ready → 503), the worker just exits. Intentional?

Yes, API detects degraded health, starts rejecting calls. When the connection is finally lost, it shutdown the api (code 1), so its operator knows it failed.

> Temporal already reconnects on its own through blips, so I want to be sure exiting beats just waiting it out.

Yes, it indeed tries to restore the connection, and it retries forever, meaning that if the connection is lost deterministically, the worker will still be "online".

TL;DR
The whole point of this task is to avoid having API/worker losing connection but not breaking, so they would just sit there, useless, until somebody sees it. With these changes they fail hard if they no longer can perform.

sdk/core/src/worker/connection_monitor.js
szanata: Yep. I dont think there were any problems doing the other way, but this is a case where `while` seems better. Thanks for suggesting it, I have updated the code.

PR #265, opened 2026-06-16
chore: fix catalog hash calculation

catsby:
## Summary

Worker catalog hash calculation was reading `*.json` files etc. from things like `logs/` which if contained a lot of trace files could result in slow startup time. This PR adds more exclusions to the list of places to calculate the hash from


szanata (approved):
That is a solid yes from me 🚀

PR #266, opened 2026-06-18
fix(llm): route system blocks to the AI SDK system option (OUT-494)

bnchrch:
## Problem

The AI SDK emits this warning on every `generateText`/`streamText` call, flooding worker logs and inflating our Datadog bill:

> System messages in the prompt or messages fields can be a security risk because they may enable prompt injection attacks. Use the system option instead when possible. Set `allowSystemInMessages` to true to suppress this warning, or false to throw an error.

- Root cause: `loadAiSdkTextOptions` placed every `<system>` block into the `messages` array as a `role:'system'` entry — exactly what the AI SDK flags. The top-level `system` option is the provider-recommended slot.
- `generateText`/`streamText` were affected; the `Agent` class already routed system content through the `instructions`/`system` slot.

## Solution

- **`loadAiSdkTextOptions`** — splits resolved messages: `<system>` blocks → top-level `system` option (as `SystemModelMessage[]`, so per-message `cacheControl`/`providerOptions` survive); user/assistant/tool stay in `messages`.
- **`Agent`** — consumes the split `system` directly as `instructions`, removing the system-filtering it previously did by hand.
- **Defense-in-depth** — `generateText`/`streamText` and the Agent's generate/stream set `allowSystemInMessages: true` (overridable) for caller-supplied message histories.

## Proof

Real AI SDK (`ai@6.0.168`) + `MockLanguageModelV3`, fed the exact shape `loadAiSdkTextOptions` now produces:

```
PASS: call resolved, text = "ok"
system delivered to provider: [{"role":"system","content":"You are concise.","providerOptions":{"anthropic":{"cacheControl":{"type":"ephemeral"}}}}]
cacheControl preserved on system: {"anthropic":{"cacheControl":{"type":"ephemeral"}}}
```

System content travels via the `system` slot with Anthropic `cacheControl` intact — caching is preserved.

## Test plan

- [x] `npm run lint` passes (full repo)
- [x] `npm test` — `@outputai/llm` suite green (279 tests), incl. new coverage in `ai_sdk_options.spec.js` (system hoisting, providerOptions preserved, multi-system, no-system) and updated `agent.spec.js`/`ai_sdk.spec.js`
- [ ] `npm run build:packages` — pre-existing unrelated failure at `sdk/cli` `copy-assets` (`native-copyfiles` `options.exclude` type error); `sdk/llm` has no build step
- [ ] Staging smoke (needs the newer `ai` that emits the warning): confirm the warning is gone from worker logs and `cachedInputTokens > 0` on a workflow using a `<system options="cached">` prompt

## Notes for reviewers

- **Hoisting does not break prompt caching.** The AI SDK `system` option is typed `string | SystemModelMessage | Array<SystemModelMessage>` and `SystemModelMessage` carries `providerOptions`, so `cacheControl` on a `<system options="cached">` block rides along (see Proof). That's why we hoist rather than only suppressing.
- **`allowSystemInMessages` is a no-op on our pinned `ai@6.0.168`** — the warning and the option are both from a newer 6.x. The hoist is the version-independent fix; the flag is belt-and-suspenders for when we bump.

## Follow-up (out of scope)

The warning is observed in a deployed environment running a newer `ai` than the lockfile pin (`6.0.168` contains neither the warning nor `allowSystemInMessages`), and `package.json` devDeps have drifted ahead of the installed store. Worth reconciling the deployed `ai` version with the lockfile separately.

Closes OUT-494

### DIFF
--- a/sdk/llm/src/agent.js
+++ b/sdk/llm/src/agent.js
@@ -43,17 +43,14 @@ export class Agent extends AIToolLoopAgent {
 
     const { loadedPrompt, tools } = prepareTextPrompt( { prompt, variables, promptDir: resolvedPromptDir, skills, tools: toolsArg } );
 
-    const { messages: allMessages, ...constructorOptions } = loadAiSdkTextOptions( loadedPrompt );
+    const { system, messages, ...constructorOptions } = loadAiSdkTextOptions( loadedPrompt );
 
-    // Extract system messages as `instructions` for the ToolLoopAgent constructor
-    // and keep user messages for generate() calls — avoids provider errors
-    // with multiple system messages during multi-step tool loops.
-    // Pass message objects (not a string) so per-message providerOptions are preserved.
-    const systemMessages = allMessages.filter( isRole( ROLE.SYSTEM ) );
-
+    // loadAiSdkTextOptions already routes system blocks to the `system` slot
+    // (preserving per-message providerOptions); pass them as the agent's
+    // `instructions` and keep the user messages for generate()/stream() calls.
     super( {
       ...constructorOptions,
-      ...( systemMessages.length > 0 ? { instructions: systemMessages } : {} ),
+      ...( system ? { instructions: system } : {} ),
       ...( tools ? { tools } : {} ),
       stopWhen: stopWhen ?? stepCountIs( maxSteps ),
       ...rest
@@ -61,7 +58,7 @@ export class Agent extends AIToolLoopAgent {
 
     this.#prompt = prompt;
     this.#modelId = loadedPrompt.config.model;
-    this.#initialMessages = allMessages.filter( isRole( ROLE.USER ) );
+    this.#initialMessages = messages.filter( isRole( ROLE.USER ) );
     this.#store = conversationStore ?? null;
   }
 
@@ -80,7 +77,7 @@ export class Agent extends AIToolLoopAgent {
     const traceId = startTrace( { name: 'Agent.generate', prompt: this.#prompt } );
     try {
       const messages = await this.#fetchMessages( userMessages );
-      const response = await super.generate( { messages, ...callOptions } );
+      const response = await super.generate( { messages, allowSystemInMessages: true, ...callOptions } );
       const wrapped = await wrapTextResponse( { traceId, response, modelId: this.#modelId } );
       await this.#storeMessages( userMessages, wrapped );
       return wrapped;
@@ -96,6 +93,7 @@ export class Agent extends AIToolLoopAgent {
       const messages = await this.#fetchMessages( userMessages );
       return super.stream( {
         messages,
+        allowSystemInMessages: true,
         ...callOptions,
         ...wrapStreamOnFinishResponse( { traceId, modelId: this.#modelId, onFinish } ),
         onError( event ) {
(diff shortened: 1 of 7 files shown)


szanata (commented):
~I see that `loadAiSdkTextOptions()` is now splitting chat messages and returning discrete `system` and `messages`, which is good, but for the actual LLM usage, nothing changes, `system` is still set as `{ instructions }` and the warning is being suppressed by `allowSystemInMessages: true`.~

~This paragraph makes me think there is a better option:~
> System messages in the prompt or messages fields can be a security risk because they may enable prompt injection attacks. Use the system option instead when possible. Set allowSystemInMessages to true to suppress this warning, or false to throw an error.

~More precisely:~
> [...] Use the system option instead when possible.

~I think it is referring to this: https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-text#system~

~Shouldn't  we start using that instead of just suppressing the warning, even more so now that we already extracted the system messages?~

My bad, the `loadAiSdkTextOptions` result is being deconstructed to AI arguments. I didnt see that at first.

sdk/llm/src/agent.js
```diff
@@ -43,25 +43,22 @@ export class Agent extends AIToolLoopAgent {
...
+      ...( system ? { instructions: system } : {} ),
       ...( tools ? { tools } : {} ),
       stopWhen: stopWhen ?? stepCountIs( maxSteps ),
       ...rest
     } );
 
     this.#prompt = prompt;
     this.#modelId = loadedPrompt.config.model;
-    this.#initialMessages = allMessages.filter( isRole( ROLE.USER ) );
+    this.#initialMessages = messages.filter( isRole( ROLE.USER ) );
```
szanata, line 62: Since system messages were already extract, do we still nee this filter?

sdk/llm/src/ai_sdk_options.js
```diff
@@ -1,20 +1,31 @@
...
     throw new FatalError( `Prompt "${prompt.name}" has no chat-style messages. Add role-tagged blocks like <system> or <user>.` );
   }
+  const isSystem = isRole( ROLE.SYSTEM );
+  const resolvedMessages = resolveMessageProviderOptions( prompt );
+  const system = resolvedMessages.filter( isSystem );
+
   const options = {
     model: loadTextModel( prompt ),
-    messages: resolveMessageProviderOptions( prompt ),
+    ...( system.length > 0 ? { system } : {} ),
```
szanata: A small nit: I would still return `system: []` even when there are none, just for consistency.

szanata (approved):

sdk/llm/src/ai_sdk_options.js
bnchrch: Done in 4a58c359 — `loadAiSdkTextOptions` now always returns `system` (as `[]` when there are no system blocks). Confirmed `system: []` / `instructions: []` are no-ops in the AI SDK (the provider prompt is just the user turn), so this is purely a consistency change.

sdk/llm/src/agent.js
bnchrch: Still needed — after the split, `messages` is system-free but can still contain authored `<assistant>`/`<tool>` blocks (several prompts use `<assistant>`, e.g. `generate_text/prompt@v1`). The filter keeps only `<user>` turns as the per-call seed; `generateText`/`streamText` still send the other blocks via `messages`. Clarified the comment and added a test locking the contract in 4a58c359.

PR #267, opened 2026-06-18
refactor(core): Moving worker loader to a folder

szanata:
## Summary

- Broke `worker/loader.js` into `worker/loader/workflows.js`, `worker/loader/activities.js` and `worker/loader/hooks.js`.
- Broke `worker/loader_tools.js` into `worker/loader/tools.js` and `worker/loader/matchers.js`.
- Removed unnused eslint config.
- Moved `createWorkflowsEntryPoint` logic to inside `loadWorkflows()`.
- Updated `loadWorkflows()` and `loadActivities()` to return objects.
- Added missing test coverage.
- Removed `activityMatchersBuilder` in favor of `buildActivityMatcher`.

The goal was to simply, add better tests and make the code more readable.

## Test plan

- [x] Manual
- [x] Added unit tests


sdk/core/src/worker/loader/matchers.js
```diff
@@ -0,0 +1,45 @@
...
+ * Creates a matcher function that based on "path", matches:
+ * - path/steps.js
+ * - path/evaluators.js
+ * - path/steps/*
+ * - path/evaluators/*
+ * @param {string} path
+ * @returns {function(string): boolean}
+ */
+export const buildActivityMatcher = path => {
+  const exp = new RegExp( `^${RegExp.escape( `${path}${sep}` )}(?:steps|evaluators)(?:\\.js$|${RegExp.escape( sep )})` );
```
bnchrch: ❗ `RegExp.escape` is Node 24+ only. On Node 22 (still the current LTS) it's `undefined`, and since this runs on every worker boot, anyone on Node 22 crashes at startup with `RegExp.escape is not a function`. The old matcher didn't need it, and we don't set an `engines` field to warn people. CI only passes because it's on Node 24.

Either pin the floor:
```json
"engines": { "node": ">=24" }
```
or escape the path ourselves and skip the built-in. What Node versions are we promising to support?

sdk/core/src/worker/loader/activities.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  // writes down the activity option overrides
+  const optionsContent = `export default ${JSON.stringify( activityOptionsMap, undefined, 2 )};`;
+  const optionsFile = writeFileInTempDir( optionsContent, ACTIVITY_OPTIONS_FILENAME );
+
+  // system activities
+  activities[ACTIVITY_SEND_HTTP_REQUEST] = sendHttpRequest;
+  activities[ACTIVITY_GET_TRACE_DESTINATIONS] = getTraceDestinations;
+  return { activities, optionsFile };
```
bnchrch, line 74: 💅 `optionsFile` gets returned but nobody reads it. `index.js` only pulls `activities`, and `bundle.js` throws the result away (just the test checks it). `loadWorkflows` returning `entrypoint` makes sense since that one's used. Drop it here, or is it for something coming up?

sdk/core/src/worker/loader/workflows.js
```diff
@@ -0,0 +1,82 @@
...
+const log = createChildLogger( 'Workflow Loader' );
+
+/**
+ * Creates a temporary index file importing all workflows for Temporal.
+ * @param {object[]} workflows
+ * @returns {string} Filename
+ */
+const createWorkflowsEntrypoint = workflows => {
+  // default system catalog workflow
+  const catalog = { name: WORKFLOW_CATALOG, path: import.meta.resolve( '../catalog_workflow/workflow.js' ) };
```
bnchrch, line 17: ❓ The catalog line builds a `file://` URL here, but every other workflow in the generated file is a plain absolute path (from `matchFiles`), so the entrypoint ends up mixing the two. Did the bundler definitely swallow the `file://` one? And any reason not to keep them the same, e.g. `fileURLToPath()` here? The test only checks the `catalog` name, not the path, so it wouldn't catch a slip.

sdk/core/src/worker/loader/hooks.js
```diff
@@ -0,0 +1,28 @@
+import { join } from 'node:path';
+import { existsSync } from 'node:fs';
+import { createChildLogger } from '#logger';
+
+const log = createChildLogger( 'Scanner' );
```
bnchrch: 💅 Tiny one: this logger is still `'Scanner'` while the others are `'Activities Loader'` / `'Workflow Loader'`. Now that it's just hooks, `'Hooks Loader'` reads better. Feel free to ignore.

bnchrch (approved):
Nice cleanup, the split reads much better and the tests came along for the ride. Approving. One thing I'd like us to sort out before it bites someone on an older Node: the new matcher leans on `RegExp.escape`. Left that plus a few small notes inline.

sdk/core/src/worker/loader/activities.js
szanata: I am planning to start using it to remove the interceptors leap of faith reading a file that might not be there. This will come later.

sdk/core/src/worker/loader/hooks.js
szanata: You are right. I didnt pay attention when migrating this one.

sdk/core/src/worker/loader/matchers.js
szanata: In my mind we support node 24+ only. since we already use `.toArray()` in a couple places. So I would said is node 24. I think we can introduce the constrain in the `package.json`.

szanata: I took a step back. `.toArray()` is Node 22. So I added a polyfill for `RegExp.espace` instead. Thanks for the suggestion.

sdk/core/src/worker/loader/workflows.js
szanata: The worker accepts `file://` without any problems. I move that import.meta.resolve because it is very convenient (no need to import `fileURLToPath` and `join` or `resolve`) and it already resolves an absolute path.

PR #269, opened 2026-06-19
fix(cli): recursive dataset/eval discovery for nested workflow folders (OUT-496)

bnchrch:
## Problem

- `output workflow test` / `dataset generate` / `dataset list` find nothing for workflows in nested folders. `resolveDatasetsDir` assumed a flat path (`src/workflows/<name>/tests/datasets`), so a workflow registered as `context_writing_style_anchor` but living at `src/workflows/context/writing/style_anchor/` resolved to nothing.
- The worker already discovers `workflow.ts` recursively, so nested workflows register fine — only CLI dataset/eval discovery was broken. Converting `_`→`/` is unsafe, and the symlink workaround double-registers the workflow under `npx output dev`.
- Secondary: when `tests/evals/workflow.ts` doesn't compile to `dist` (a tsconfig that excludes `src/**/tests`), `<wf>_eval` never registers and the CLI fails with a generic `WorkflowNotFoundError`.

## Solution

- **Shared resolver** — extracted the catalog-based "find a registered workflow's dir by name" logic (already proven for scenarios) into `sdk/cli/src/utils/workflow_dir.ts`. `resolveWorkflowDir()` tries the flat path first (offline, unchanged for flat layouts), then falls back to the worker catalog for nested folders. `scenario_resolver.ts` now consumes it and re-exports its two public symbols.
- **Datasets** — `resolveDatasetsDir` / `resolveDefaultDatasetsDir` are now catalog-aware; the write path targets the nested `tests/datasets` even on first generation. `dataset list` gets nested support for free.
- **Diagnostic** — `output workflow test` checks the catalog for `<wf>_eval` before running anything; if its source exists but isn't registered, it points at the tsconfig/`dist` fix instead of a bare `WorkflowNotFoundError`.
- **Docs** — nested-folder discovery note, and the `tests/evals` → `dist` tsconfig guidance (exclude `*.spec.ts`/`*.test.ts`, not the whole `tests` tree).

## Test plan

- [x] `eslint` — clean on all changed files
- [x] `tsc --noEmit` (sdk/cli) — clean
- [x] `vitest run` — 689 cli unit tests pass, including new `workflow_dir.spec.ts` and nested read/write cases in `datasets.test.ts`
- [ ] e2e against `./run.sh dev`: `output workflow test <nested>` and `dataset generate <nested> --trace` resolve without a symlink

> Note: 2 `copy_assets.spec.ts` failures in the local run are pre-existing and unrelated — they assert on `sdk/cli/dist` build artifacts not produced in a source-only test run.

## Notes for reviewers

- **Dataset resolution now defaults its base path to `getWorkflowsBasePath()` instead of `process.cwd()`** — this matches scenario resolution and makes datasets resolve correctly under `output dev` (which sets `OUTPUT_WORKFLOWS_DIR`). In-project CLI use is unchanged: `getWorkflowsBasePath()` returns `process.cwd()` when the env var is unset.
- **The eval-registration pre-check is a second catalog fetch** (the nested dataset read may also fetch it). Kept separate for clarity — the catalog GET is cheap and only `test` pays it. The check is best-effort: if the catalog is unreachable it's skipped so the normal run path surfaces the error.

Closes OUT-496

### DIFF
--- a/.changeset/out-496-recursive-dataset-discovery.md
+++ b/.changeset/out-496-recursive-dataset-discovery.md
@@ -1,0 +1,5 @@
+---
+"@outputai/cli": patch
+---
+
+Resolve datasets and evals for nested workflow folders by the workflow's registered name (via the worker catalog), keeping the flat-path lookup as an offline fast path. `output workflow test`, `dataset generate`, and `dataset list` now work for workflows in nested directories (e.g. `src/workflows/a/b/c` registered as `a_b_c`) without a symlink. `output workflow test` also fails fast with an actionable message when a `<wf>_eval` workflow's source exists but didn't compile to `dist`.
--- a/docs/guides/evaluators/datasets-and-cli.mdx
+++ b/docs/guides/evaluators/datasets-and-cli.mdx
@@ -135,6 +135,10 @@ output workflow test blog_generator --dataset happy_path,edge_case
 
 Use `--cached` during development when iterating on evaluators — it's fast because it skips the workflow entirely. Use `--save` when you want to capture fresh output and eval results.
 
+<Note>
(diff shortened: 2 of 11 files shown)


szanata (approved):
Looks good to me.

PR #270, opened 2026-06-19
refactor(core): Reorganizing core exposed tools

szanata:
## Summary

Reorganizing core exports for other modules:
- Renaming `./sdk_activity_integration` endpoint to `./sdk/runtime`;
  - Designed to have runtime dependent tools;
  - Moved in `emitEvent` as `Event.emit`;
  - Moved in `getExecutionContext` as `Context.getActivityContent`;
  - Moved in `Tracing`;
- Added new endpoint `./sdk/helpers` :
  - Design to have sandbox safe tools:
  - Moved in `resolveInvocationDir()` as `Path.resolveInvocationDir()`;
  - Moved in `getMetadata()` as `ComponentMetadata.getName()` (plus `ComponentMetadata.has()`);
- Removed methods from `./sdk_utils` endpoints:
  - `throws`: not used;
  - `setMetadata`: moved to internal utils;
  - `getMetadata`: replace by `ComponentMetadata.*` at `./sdk/helpers`;
  - `resolveInvocationDir`: replaced by `Path.resolveInvocationDir()` at `./sdk/helpers`.

## Test plan
- [x] Added unit tests
- [x] Manually tested


sdk/core/src/sdk_interface/workflow/component.d.ts
```diff
@@ -0,0 +1,23 @@
...
+  type: string,
+  options: object
+};
+
+/**
+ * Extract attributes from a Outputai component
+ * @param fn
+ * @returns
+ */
+export declare function readAttributes( Function: fn ): Attribute | undefined;
```
catsby: Isn't this a parameter named `Function` of type `fn`? Line 23 below has `fn: Function`.

sdk/evals/src/index_sandbox.ts
```diff
@@ -29,10 +29,10 @@
  * is fixed to generate output from the modified AST even when rewriteFnBodies() returns
  * false — i.e. when collectTargetImports() stripped imports but no fn bodies needed
  * rewriting. This also requires injecting metadata stubs for the stripped evaluator
- * references, since evalWorkflow() calls getMetadata(def.evaluator) at module init time
- * and the identifiers would otherwise be undefined. Once the rewriter properly strips the
- * evaluator → judge → @outputai/llm chain, this file and the "output-workflow-bundle" export
- * condition in package.json can be deleted.
+ * references, since evalWorkflow() calls Context.readAttributes(def.evaluator) at module
```
catsby: Should this be `Component`?

```suggestion
 * references, since evalWorkflow() calls Component.readAttributes(def.evaluator) at module
```

sdk/core/src/sdk_interface/workflow/component.d.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Check if a function is an Outputai component
+ * @param fn
+ */
+export declare function isComponent( fn: Function ): Boolean;
```
catsby: Should this be `boolean`? 

```suggestion
export declare function isComponent( fn: Function ): boolean;
```

sdk/core/src/sdk_interface/workflow/component.d.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Attributes
+ */
+export type Attribute = {
```
catsby: It's not clear to me what this type does. It only seems to be used as part of `readAttributes`? Unless I'm mistaken all the fields here are required (there's no `?`). But the call sites don't always send that, unless I'm mistaken

- interface/step.js:https://github.com/growthxai/output/blob/c2347a61765fe3cbdfe272c2d8ca5e026d38cd16/sdk/core/src/interface/step.js#L19
- interface/workflow.js: https://github.com/growthxai/output/blob/c2347a61765fe3cbdfe272c2d8ca5e026d38cd16/sdk/core/src/interface/workflow.js#L115

What am I missing about this type?

sdk/core/src/sdk_interface/workflow/index.js
```diff
@@ -0,0 +1 @@
+export * as Component from '../workflow/component.js';
```
catsby: `from '../workflow/component.js';` is the `../workflow/` necessary here?

sdk/evals/src/verify.spec.ts
```diff
@@ -1,20 +1,10 @@
 import { describe, it, expect } from 'vitest';
 import { verify } from './verify.js';
-import { getMetadata } from '@outputai/core/sdk_utils';
 import { EvaluationBooleanResult, z } from '@outputai/core';
 import { Verdict } from './verdict.js';
 import type { CheckContext } from './verify.js';
 
 describe( 'verify', () => {
-
-  it( 'returns a function with metadata matching the given name', () => {
```
catsby, line 10: do we have this test covered elsewhere now?

sdk/core/src/sdk_interface/activity/path.d.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Return the first immediate directory of the file invoking the code that called this function.
+ *
+ * Excludes `@outputai/core`, node, and other internal paths.
+ */
+export function resolveInvocationDir(): string;
```
catsby: the implementation of this has `additionalIgnorePaths = []` in it, should we add that here?

sdk/core/src/internal_utils/component.js
```diff
@@ -0,0 +1,10 @@
+import { METADATA_ACCESS_SYMBOL } from '#consts';
+
+/**
+ * Add metadata "values" property to a given object
+ * @param {object} target
+ * @param {object} values
+ * @returns
```
catsby: I think we're missing what `@returns` here

sdk/core/src/sdk_interface/workflow/component.d.ts
szanata: Yes, boolean with lowercase b is the right type.

sdk/core/src/sdk_interface/workflow/component.d.ts
szanata: `inputSchema` and `outputSchema` are indeed optional. I will add `?`.  Yes this type is just for the readAttributes, just so we have proper types.

sdk/core/src/sdk_interface/workflow/index.js
szanata: No. That slipped when I was adding this folder and moving files around.

sdk/evals/src/verify.spec.ts
szanata: This test was testing the feature from `@outputai/core`, which should be tested in core itself (which it is).

sdk/core/src/internal_utils/component.js
szanata: It doesnt return anything

catsby: This doc on [Object.defineProperty()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#return_value) says it returns the object, in this instance `target`

sdk/core/src/activity_integration/event_id_integration.spec.js
```diff
@@ -1,52 +0,0 @@
...
-
-    emitEvent( 'cost:llm:request', { modelId: 'gpt-4o' } );
-
-    expect( handler ).toHaveBeenCalledWith( expect.objectContaining( {
-      eventId: expect.stringMatching( UUID_V4_REGEX ),
-      modelId: 'gpt-4o'
-    } ) );
-  } );
-
-  it( 'cost:http:request and http:request for the same fetch get distinct eventIds', () => {
```
catsby, line 24: is this test covered somewhere else now?

sdk/core/src/sdk_interface/workflow/component.d.ts
catsby: That only addresses 2 of the issues here. The 2 locations I linked both use `description`, which isn't here at all. `interface/step.js` sends `type` and `options`, but `interface/workflow.js` sends `aliases` and does not send `type` or `options`, which are both still required? `internal_activities/index.js` only sends `type`:

https://github.com/growthxai/output/blob/e1cd79bb5c439ed4fbf242efe5951684869b2357/sdk/core/src/internal_activities/index.js#L56

The only place that `Component.Attribute` is later used is in `src/eval_workflow.ts` (that I see at this time) and it uses `as Component.Attribute` to force whatever comes out of `readAttributes()` to fit `Attribute`, is that correct? Unless I'm mistaken we're accepting seemingly arbitrary shape as input but we're coercing to a set output, and the 2 don't seem to align. The fields are still there technically but the type returned won't show them. 

On top of all this, it seems odd to me that we set this information with `setMetadata()` but read it with `readAttributes()`. Previously it was `getMetadata()`, wasn't it? It seems like arbitrary metadata going in, but it's specifically typed attributes coming out. Is that correct?

If this is meant for us to have proper types, shouldn't the input method (`setMetadata`) enforce that?

sdk/core/src/sdk_interface/activity/events.spec.js
```diff
@@ -11,12 +11,12 @@ vi.mock( '#bus', () => ( {
   messageBus: { emit: emitMock }
 } ) );
 
-import { emitEvent } from './events.js';
+import { emit } from './events.js';
 
 // `eventId` stamping is the bus layer's responsibility (see bus.spec.js + the
 // integration tests in event_id_integration.spec.js). Assertions here use
 // `objectContaining` so they don't have to know about that enrichment.
```
catsby: integration test file `event_id_integration.spec.js` no longer exits in this PR

sdk/core/src/sdk_interface/workflow/index.d.ts
```diff
@@ -4,6 +4,5 @@
  *
  * @packageDocumentation
  */
-export { getExecutionContext } from './context';
-export { emitEvent } from './events';
-export * as Tracing from './tracing';
+export * as Component from '../workflow/component.js';
```
catsby: same issue with `../workflow/` right? and should we export from `component.d.ts`? Other `index.d.ts` files like `sdk_interface/index.d.ts` files export from the corresponding `.d.ts` files but here we export from `.js`, does that matter?

sdk/core/src/sdk_interface/activity/events.spec.js
szanata: Removed the comment

sdk/core/src/sdk_interface/workflow/index.d.ts
szanata: Fixed. And fixed all type files introduced in this PR

sdk/core/src/activity_integration/event_id_integration.spec.js
szanata: Yes, this is tested on bus.spec.js

szanata:
@catsby I changed the attributes to respect the same rules as the schemas, so only name and type are mandatory, others, like input/outputSchema, description are optional.

The `setMetadata` on the internal activities is not read via `readAttributes` as that is inaccessible externally. Hence why that doesnt need to respect the same shape.

sdk/core/src/internal_utils/component.js
szanata: Yes but this is just internal JSDocs, we dont need to document at that level, it is just a helper for developers. The function usage is void-like, so we can leave to that.

szanata:
I re-updated the code, now replacing the `Attributes` type and `readAttributes` in favor of a single `getComponentName()` which returns `string | undefined`. I think that solves the issue.

sdk/core/src/sdk_interface/workflow/component.d.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Extract the name from a Outputai component
+ * @param fn
+ * @returns
+ */
+export declare function getComponentName( fn: Function ): string | undefined;
```
bnchrch: 💅 `getComponentName` and `isComponent` both take bare `Function`. We steer clear of that elsewhere (the old call site used `Function & Record<symbol, unknown>`), and it can't see the symbol prop these actually read. Tighten the param to the component shape, or at least `(...args: unknown[]) => unknown`?

sdk/core/src/sdk_interface/workflow/component.d.ts
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Check if a function is an Outputai component
+ * @param fn
+ */
+export declare function isComponent( fn: Function ): boolean;
```
bnchrch: ❓ With `Component.isComponent` + `Component.getComponentName`, the `Component.` prefix stutters twice. If this namespace is the umbrella for the building-block family (workflow/step/evaluator), WDYT about `Primitive`? Reads as `Primitive.is(fn)` / `Primitive.name(fn)`. Not `WorkflowPrimitive` (repeats the `internal/workflow` path it's imported from, and `WORKFLOW` is itself a ComponentType so it sounds kind-specific).

sdk/core/package.json
```diff
@@ -12,9 +12,13 @@
...
-    "./sdk_activity_integration": {
-      "types": "./src/activity_integration/index.d.ts",
-      "import": "./src/activity_integration/index.js"
+    "./internal/workflow": {
+      "types": "./src/sdk_interface/workflow/index.d.ts",
+      "import": "./src/sdk_interface/workflow/index.js"
+    },
+    "./internal/activity": {
+      "types": "./src/sdk_interface/activity/index.d.ts",
+      "import": "./src/sdk_interface/activity/index.js"
```
bnchrch: ❓ The same idea, "core's surface for sibling packages and not end users", is spelled three ways here, and the folder doesn't match its path:

```
@outputai/core/internal/activity   ->  src/sdk_interface/activity/   (path: internal, dir: sdk_interface)
@outputai/core/internal/workflow   ->  src/sdk_interface/workflow/
@outputai/core/sdk_utils           ->  src/utils/                    (a third word for the same bucket)
```

Could we land on one word, `internal`, for both, with everything internal under it?

```
@outputai/core                     public (end users)
@outputai/core/internal/activity   src/internal/activity/
@outputai/core/internal/workflow   src/internal/workflow/
@outputai/core/internal/utils      src/internal/utils/   (was sdk_utils)
```

Two separate asks:
- Rename the `sdk_interface/` folder to `internal/` so dir == path. Cheap, could do it here.
- `sdk_utils` is internal plumbing too (clone/deepMerge/base64, and no "internal use only" banner unlike the others), so it'd sit as `internal/utils` under that banner. Bigger move since siblings import it, so happy for a follow-up. Or if it's meant to be genuinely public, let's mark it that way so the line is clear.

sdk/core/src/sdk_interface/activity/index.d.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * > [!WARNING]
+ * > **Internal use only.** Not part of the public API; may change without notice.
+ *
+ * @packageDocumentation
+ */
```
bnchrch: 📚 Nice split, but the logic isn't stated: `workflow/` is sandbox-safe (pure reads), `activity/` needs the runtime (async storage, the bus, the stack). One line here helps the next person pick the right bucket.

sdk/core/src/sdk/runtime/events.spec.js
```diff
@@ -11,12 +11,11 @@ vi.mock( '#bus', () => ( {
...
 } ) );
 
-import { emitEvent } from './events.js';
+import { emit } from './events.js';
 
-// `eventId` stamping is the bus layer's responsibility (see bus.spec.js + the
-// integration tests in event_id_integration.spec.js). Assertions here use
-// `objectContaining` so they don't have to know about that enrichment.
-describe( 'emitEvent', () => {
+// `eventId` stamping is the bus layer's responsibility (see bus.spec.js).
```
bnchrch, line 16: ❓ This drops the pointer to `event_id_integration.spec.js`, which the PR deletes, and that was the only end-to-end eventId test. bus.spec.js still covers stamping and the `external:` prefix is covered here, so just the composed path is gone. Deliberate?

sdk/core/src/sdk_interface/workflow/component.js
```diff
@@ -0,0 +1,5 @@
+import { METADATA_ACCESS_SYMBOL } from '#consts';
+
+export const getComponentName = fn => fn[METADATA_ACCESS_SYMBOL]?.name;
```
bnchrch: 🧪 `getComponentName`/`isComponent` don't have a direct spec, `eval_workflow.spec.ts` only hits them through a mock. A tiny `component.spec.js` (component vs plain fn, plus the no-name case that returns `undefined`) would lock the real behavior.

bnchrch (commented):
Solid reorg. Nice that `getComponentName` returns `string | undefined` now, that's the optional-name shape we wanted. The rest is nits and questions, notes inline.

sdk/core/package.json
szanata: About `/utils` I plan on refactor it later, hence why I didnt change.

Internally, I avoided using `/internal` because that sounds like it is internal to core, when they ar eactually exposed, hence why I changed to `./sdk_interface`.

sdk/core/src/sdk_interface/workflow/component.d.ts
szanata: The new call-shape of Function suffice for current usage.

sdk/core/package.json
szanata: I renamed everything to `/sdk/*`

sdk/core/src/sdk/runtime/events.spec.js
szanata: Yes, this is deliberated, the behavior is well tested already, the integration test was redundant.

sdk/core/src/sdk_interface/workflow/component.js
szanata: Added specs.

sdk/core/src/sdk_interface/activity/index.d.ts
szanata: I added much more docs and better org

sdk/core/src/sdk_interface/workflow/component.d.ts
szanata: Renamed to `ComponentMetadata.has()` and `ComponentMetdata.getName()`, as "Primitive" seems too generic for me.

bnchrch (approved):
Approving, this addressed everything.

💎 `sdk/README.md` is the standout. Documenting the helpers-vs-runtime split as a hard Temporal boundary (with the allowed / not-allowed lists) is exactly the context the next person needs, well past the one line I asked for.

`ComponentMetadata.has()`/`getName()` reads clean, and `/sdk/*` with dir == path is the consistency I was after. Fair calls on keeping `Function` and on deferring `sdk_utils`.

PR #277, opened 2026-06-22
feat(cli): add workflow history command with terminal waterfall

bnchrch:
## Overview

- New CLI command `output workflow history <workflowId>` renders a run's step timeline as a terminal **waterfall** — each step positioned by start offset and sized by duration — mirroring the Agents HQ "Timeline" view.
- Pages `GET /workflow/{id}/history` (50 events/page, follows `nextPageToken`, pins the resolved `runId`) and correlates the flat Temporal events into per-step spans: opens on `ACTIVITY_TASK_SCHEDULED`, attaches STARTED/terminal events, handles child workflows, and filters framework noise (`__internal#`, `$shared#callWebhook`).
- Parallel fan-outs of the same step are numbered `#1..#N`; the time axis auto-scales its ticks.
- **Output modes** — default text waterfall; `--format json` emits structured `{ workflow, runId, totalDurationMs, spans }`; `--raw` dumps the endpoint's verbatim response. Plus `--run-id`, `--include-payloads`, `--width`, and `--no-color`.

The event→span correlation and tick/bar geometry are a TypeScript port of Atlas's existing `OutputWorkflows::WorkflowHistory::Correlator` + `StepGantt`, which consume this same endpoint.

## Preview

Representative render (sample data, `--no-color`):

```
content-brief-generator · run 2fe0b36b · completed · 2m29s

                       0            30s            1m           1m30s                       2m
Compress Text          █████████████                                                             27s
Fetch Serp Results                  █                                                             2s
Scrape Serp Page #1                  ▏                                                            1s
Scrape Serp Page #2                  ████                                                         8s
Scrape Serp Page #4                  ██████                                                      13s
Scrape Serp Page #9                   ███████                                                    15s
Extract Page Topics #1                     █████                                                 11s
Extract Page Topics #3                     ██████                                                13s
Synthesize Coverage                              ██████████████████████████████████            1m12s
Generate Brief                                                                      ████████     16s
Emit Workflow Event                                                                        ▏     0ms
```

(rows trimmed for brevity; real runs show every step)

## Test plan

- [x] `npx eslint` — clean on all new files
- [x] `npx tsc --noEmit` — no type errors in the new files (pre-existing repo errors are from uninstalled workspace deps, unrelated)
- [x] `vitest run` — 19 new unit tests pass (correlator, span_labels, waterfall, command); sibling `workflow status` / `runs list` specs still green
- [x] `output workflow history --help` — command + flags register under the `workflow` topic
- [x] Manual smoke against a live API — `OUTPUT_API_URL` / `OUTPUT_API_AUTH_TOKEN` weren't available in my env; please run against a real run and cross-check durations/ordering vs the Agents HQ Timeline

## Notes for reviewers

- **Re-implemented Atlas's event→span correlation in TS rather than sharing one implementation.** The ticket was CLI-scoped and the endpoint returns raw Temporal events; Atlas correlates them in Ruby. `correlator.ts` is kept structurally faithful to the Ruby `Correlator`/`Span` so the two stay easy to diff. A future consolidation would be a server-side derived `spans` view both consume.
- **Step labels are humanized `activityType.name` (`capitalCase`), not catalog descriptions** — the CLI has no step catalog. Matches the UI for the common case.

Closes OUT-424


sdk/cli/src/services/workflow_history.ts
```diff
@@ -0,0 +1,130 @@
...
+
+export interface WorkflowMeta {
+  workflowId?: string;
+  runId?: string;
+  status?: string;
+  startTime?: string;
+  closeTime?: string | null;
+  historyLength?: number;
+  taskQueue?: string;
+  [key: string]: unknown;
```
catsby: What is this `[key: string]: unknown;` supposed to facilitate?

sdk/cli/src/services/workflow_history.ts
```diff
@@ -0,0 +1,130 @@
...
+  workflowId: string,
+  includePayloads: boolean,
+  runId: string | undefined,
+  pageToken: string | undefined,
+  acc: PageAccumulator
+): Promise<PageAccumulator> {
+  const response = await getWorkflowIdHistory( workflowId, { runId, pageSize: PAGE_SIZE, pageToken, includePayloads } );
+  if ( !response ) {
+    throw new Error( 'Failed to connect to API server. Is it running?' );
+  }
```
catsby: I believe this is dead code; `getWorkflowIdHistory()` calls `customFetchInstance()` which either returns a truthy object or throws. The `!response.data` guard is good though

catsby (approved):
There's a lot here but overall good

sdk/cli/src/services/workflow_history.ts
bnchrch: It was only there so the generated opaque `data.workflow` (`{ [key: string]: unknown }`) would cast to `WorkflowMeta`. Dropped it — now narrowing via `as unknown as` at the single cast site, which also tightens the type so field typos get caught. Fixed in 2b6719d10.

sdk/cli/src/services/workflow_history.ts
bnchrch: Good catch — removed. Kept the `!response.data` guard since `response.json()` can still resolve `undefined`. 2b6719d10

PR #280, opened 2026-06-22
fix(docs): escape MDX in generated changelog so the page renders (OUT-490)

bnchrch:
## Problem

The changelog at [docs.output.ai/changelog](https://docs.output.ai/changelog) rendered its intro but listed **no releases**, despite `releases.json` holding all 23 (v0.1.1 → v0.8.0).

Root cause: `regenerate.mjs` injects raw changeset summaries into MDX. A dependency-bump summary contains **bare `<` in prose** — `vite@>=7.1.0 <=7.3.1`, `hono@<4.12.12`, `Unescaped </style>`. MDX reads `<` as the start of a JSX tag and throws (`Unexpected character "=" before name`), which drops the whole `changelog/index` page. The old design hid this behind an imported snippet (`<Changelog/>`): the parse error broke the *snippet* compile, so the page kept its intro while the changelog rendered empty.

## Solution

- **`renderer.mjs`** — add `escapeMdxText()` and apply it to each changeset summary. Escapes `<`→`&lt;` and `{`→`&#123;` in prose only, leaving inline-code spans and fenced blocks verbatim. This is the fix that makes the page parse.
- **`changelog/index.mdx` + `regenerate.mjs`** — inline the generated `<Update>` blocks directly into the page between `AUTO-GENERATED` markers (Mintlify's canonical changelog shape), via an idempotent read→splice→write that fails loudly if the markers are missing. Removes the imported snippet (`docs/guides/snippets/changelog.mdx` deleted).
- **CI** — new `ops/assert_changelog_synced.sh`, run in both `docs_regenerate.yml` jobs: asserts the markers exist and the `<Update>` count matches the release count.

## Proof

`mint dev` (`./run.sh docs:mint`), `/changelog`:

- **Before:** `parsing error ./changelog/index.mdx:619:17 - Unexpected character "=" before name` → page 404s.
- **After:** intro + all 23 `<Update>` cards (v0.8.0 → v0.1.1) render; escaped `&lt;=` displays correctly as `<=` (`vite@>=7.1.0 <=7.3.1: >=7.3.2`).

## Test plan

- [x] `./run.sh docs:mint` → `/changelog` renders all 23 releases (browser-verified), no 404
- [x] `./ops/regenerate_docs.sh` is idempotent (second run → no `git diff`); no generator/formatter drift
- [x] `./ops/assert_changelog_synced.sh` passes; negative-tested (count mismatch + missing marker both exit 1)
- [x] `node --check` on both scripts; changed `.mjs` conforms to the eslint ruleset (no `let`, max-len ≤150)
- [ ] `npm run lint` / CI (not run locally — empty `node_modules` in worktree)

## Notes for reviewers

- **Inlined the `<Update>` blocks and deleted the snippet** (beyond the minimal escape fix) — it's Mintlify's canonical changelog pattern, aligns with `docs_regenerate.yml`'s existing `git add docs/guides/changelog`, and turns any future parse error into a loud 404 instead of a silent empty component.
- **The CI assertion guards empty/desync, not MDX parse errors** — `escapeMdxText` is the actual regression guard for this bug class. A full `mint` build check in CI was considered and deferred.

Closes OUT-490


catsby (approved):
I verified this works. Is it in scope to see if we can add links into the changelog entries? Looking at this, nothing is clickable to take me to the PR that made the change. Is that something we can set in `changesets`? Just curious if it's an easy toggle or something, we can do a follow-up if it's not

Image

bnchrch:
@catsby Ill create a follow up PR to do the PR linking.

Essentially we'd have to rely on commit history to retroactively do this and/or update how our changesets get generated.

PR #284, opened 2026-06-23
feat(core): Logger export [OUT-255]

szanata:
## Summary

Added a new `Logger` export for structured logging from both workflows and steps. Logs use the same `message` plus `metadata` shape as the internal worker logger and are routed through the worker's Winston logger.

All log messages are enriched with execution metadata:
- Workflow logs include `workflowType`, `workflowId`, and `runId`.
- Step logs include the same workflow fields, plus `activityType` and `activityId`.

```ts
import { Logger } from '@outputai/core';

Logger.info( 'I am a log', { extraInfo: 'none' } ); // works inside workflow and steps
```

Supported levels are:
- error
- warn
- info
- http
- verbose
- debug
- silly

The default displayed level is debug in development and info in production. Override it with `OUTPUT_LOG_LEVEL` env var. This setting also affects internal worker logs.

### Log Examples
This shows logs from withing activities and workflows:
image

## Test plan

- [x] Added unit tests
- [x] Added test workflows
- [x] Manually tested

### DIFF
--- a/sdk/core/src/interface/logger.d.ts
+++ b/sdk/core/src/interface/logger.d.ts
@@ -6,48 +6,48 @@ export type LogMetadata = Record<string, unknown>;
 /**
  * Log an error (level 0)
  * @param message Log message
- * @param metadata Additional information to be serialized
+ * @param metadata Additional information to be displayed
  */
 export declare function error( message: string, metadata?: LogMetadata ) : void;
 
 /**
  * Log a warn (level 1)
- * @param message
- * @param metadata
+ * @param message Log message
+ * @param metadata Additional information to be displayed
  */
 export declare function warn( message: string, metadata?: LogMetadata ) : void;
 
 /**
  * Log an info (level 2)
- * @param message
- * @param metadata
+ * @param message Log message
+ * @param metadata Additional information to be displayed
  */
 export declare function info( message: string, metadata?: LogMetadata ) : void;
 
 /**
  * Log http (level 3)
- * @param message
- * @param metadata
+ * @param message Log message
+ * @param metadata Additional information to be displayed
  */
 export declare function http( message: string, metadata?: LogMetadata ) : void;
 
 /**
  * Log verbose (level 4)
- * @param message
- * @param metadata
+ * @param message Log message
+ * @param metadata Additional information to be displayed
  */
 export declare function verbose( message: string, metadata?: LogMetadata ) : void;
 
 /**
  * Log debug (level 5)
- * @param message
- * @param metadata
+ * @param message Log message
+ * @param metadata Additional information to be displayed
  */
 export declare function debug( message: string, metadata?: LogMetadata ) : void;
 
 /**
  * Log silly (level 6)
- * @param message
- * @param metadata
+ * @param message Log message
+ * @param metadata Additional information to be displayed
  */
 export declare function silly( message: string, metadata?: LogMetadata ) : void;
--- a/sdk/core/src/interface/logger.d.ts
+++ b/sdk/core/src/interface/logger.d.ts
@@ -1,0 +1,53 @@
+/**
+ * Additional structured fields attached to a log record.
+ */
+export type LogMetadata = Record<string, unknown>;
+
+/**
+ * Log an error (level 0)
+ * @param message Log message
+ * @param metadata Additional information to be serialized
+ */
+export declare function error( message: string, metadata?: LogMetadata ) : void;
+
+/**
+ * Log a warn (level 1)
+ * @param message
+ * @param metadata
+ */
+export declare function warn( message: string, metadata?: LogMetadata ) : void;
+
+/**
+ * Log an info (level 2)
+ * @param message
+ * @param metadata
+ */
+export declare function info( message: string, metadata?: LogMetadata ) : void;
+
+/**
+ * Log http (level 3)
+ * @param message
+ * @param metadata
+ */
+export declare function http( message: string, metadata?: LogMetadata ) : void;
+
+/**
+ * Log verbose (level 4)
+ * @param message
+ * @param metadata
+ */
+export declare function verbose( message: string, metadata?: LogMetadata ) : void;
+
+/**
+ * Log debug (level 5)
+ * @param message
+ * @param metadata
+ */
+export declare function debug( message: string, metadata?: LogMetadata ) : void;
+
+/**
+ * Log silly (level 6)
+ * @param message
+ * @param metadata
+ */
+export declare function silly( message: string, metadata?: LogMetadata ) : void;
--- a/sdk/core/src/interface/logger.js
+++ b/sdk/core/src/interface/logger.js
@@ -1,22 +1,60 @@
 import { inWorkflowContext } from '@temporalio/workflow';
 import { proxySinks } from '@temporalio/workflow';
-import { validateLogArguments } from './validations/index.js';
 import { ACTIVITY_LOGGER_SYMBOL } from '#consts';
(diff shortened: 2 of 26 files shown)


sdk/core/src/interface/validations/schemas.spec.js
```diff
@@ -206,4 +207,63 @@ describe( 'validation schemas', () => {
...
+      const reservedFields = [
+        'label',
+        'level',
+        'message',
+        'metadata',
+        'namespace',
+        'splat',
+        'stack',
+        'timestamp'
+      ];
```
catsby: nit: this is missing some entries from the above `reservedLogMetadataFields` - can we reuse that one instead of redeclaring them here? Not a blocker just an idea

sdk/core/src/interface/validations/schemas.js
```diff
@@ -111,3 +111,30 @@ export const executeInParallelSchema = z.object( {
...
+  'namespace',
+  'splat',
+  'stack',
+  'timestamp',
+  'workflowId',
+  'workflowType',
+  'runId',
+  'activityId',
+  'activityType'
+] );
```
catsby: should we include `service` and/or `environment` here? We set those in the `development.js` and `production.js` files. Is it ok if users override them?

sdk/core/src/interface/logger.d.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Log a warn (level 1)
+ * @param message
+ * @param metadata
+ */
+export declare function warn( message: string, metadata?: LogMetadata ) : void;
+
+/**
+ * Log an info (level 2)
+ * @param message
+ * @param metadata
```
catsby: `@param message` and `@param metadata` are empty everywhere except for line 8-9, is that intentional?

catsby (approved):
Non-blocking questions

sdk/core/src/interface/validations/schemas.js
szanata: Good suggestion

catsby (approved):

sdk/core/src/interface/logger.js
```diff
@@ -0,0 +1,30 @@
+import { inWorkflowContext } from '@temporalio/workflow';
+import { proxySinks } from '@temporalio/workflow';
+import { validateLogArguments } from './validations/index.js';
+import { ACTIVITY_LOGGER_SYMBOL } from '#consts';
+
+// This is inoffensive and can be used outside workflow sandbox
+const sinks = proxySinks();
+
+const log = ( level, message, metadata ) => {
+  validateLogArguments( { message, metadata } );
```
bnchrch: ❗ A log line can crash the run. `log()` validates first and throws on a non-string message or a "reserved" metadata key.

- In a workflow the throw fails the task. It happens again every retry, so the workflow gets stuck.
- In a step the activity fails and retries, then the run fails.

The reserved names are common ones: `message`, `level`, `timestamp`, `stack`, `runId`, `label` (and `service`/`environment`, which aren't in the docs). So `Logger.info( 'event', { message: detail } )` throws today.

A log call shouldn't be able to stop the run. Could we drop or rename the clashing key instead of throwing? log_hooks already puts the framework fields last, so they win anyway. At least skip the throw inside a workflow. Happy to pair.

sdk/core/src/worker/sinks.js
```diff
@@ -10,6 +10,11 @@ export const sinks = {
    * Workflow lifecycle sinks
    */
   workflow: {
+    log: {
```
bnchrch, line 13: 💅 `workflow.log` is the only sink without an explicit `callDuringReplay: false`, and the only one `sinks.spec.js` doesn't assert. The default is already false, so replay is fine. Just nice to match the others and add the assertion so it stays obvious. Your call.

sdk/core/src/worker/log_hooks.js
```diff
@@ -47,6 +47,13 @@ messageBus.on( BusEventType.ACTIVITY_ERROR, ( { activityInfo, outputActivityKind
   } )
 );
 
+messageBus.on( BusEventType.ACTIVITY_LOG, ( { level, message, metadata, activityInfo } ) =>
```
bnchrch, line 50: ❓ User logs use the same `Activity`/`Workflow` namespaces as the built-in lifecycle logs, so `Logger.info( 'hi' )` looks the same as the framework's `Started X` lines. On purpose? Might be nice to tell user logs apart from runtime ones. Not blocking, just curious.

sdk/core/src/interface/logger.js
```diff
@@ -0,0 +1,30 @@
...
+  // When inside workflow, use sinks to send logs out
+  if ( inWorkflowContext() ) {
+    sinks.workflow.log( { level, message, metadata } );
+  // When inside activities, use the global function to ship logs
+  } else if ( typeof globalThis[ACTIVITY_LOGGER_SYMBOL] === 'function' ) {
+    globalThis[ACTIVITY_LOGGER_SYMBOL]( { level, message, metadata } );
+  // This fallback is used on unit tests
+  } else {
+    console.log( `logger.${level}`, message, metadata );
```
bnchrch: ❓ With no worker, every level goes to `console.log`, so `Logger.error` prints to stdout. Just for tests, or also when the logger runs outside a worker (evals, scripts)? If so, maybe send error/warn to `console.error`. Small.

sdk/core/src/worker/global_functions.js
```diff
@@ -0,0 +1,20 @@
+import { messageBus } from '#bus';
+import { ACTIVITY_LOGGER_SYMBOL, BusEventType } from '#consts';
+import { activityInfo as activityInfoFn } from '@temporalio/activity';
+
+const setFunction = ( key, fn ) => Object.defineProperty( globalThis, key, {
```
bnchrch, line 5: 💅 `configurable: false, writable: false` means `bindGlobalFunctions()` can only run once. A second call throws, and `delete globalThis[ ... ]` won't work (its spec's `beforeEach` does that, and passes only because there's one test). Fine now, but a second test or a re-init breaks. `configurable: true` fixes it. Nit.

bnchrch (commented):
One thing before merge, plus a question and a few nits inline:

❗ a log call can throw and get a workflow stuck (`interface/logger.js`)

Nice that this logs from inside workflows, that's the part mine couldn't do.

🔮 Not this PR: our own `console.*` (llm cost, prompt validation, http cost) and the internal HttpClient log still skip this logger, so they stay untraceable. Good follow-up.

sdk/core/src/worker/global_functions.js
szanata: This is on purpose. This has to be set once. Test should accommodate that.

sdk/core/src/interface/logger.js
szanata: Changed to arguments sanitation in favor of validation. This makes console pretty hard to break.

sdk/core/src/interface/logger.js
szanata: I have added translation

sdk/core/src/worker/log_hooks.js
szanata: This was deliberate, as I thought that user logs, even being different, belong to those categories.

I think I would release this version like this first and get some usage info first, and then see if people actually needs partition on this data.

sdk/core/src/worker/sinks.js
szanata: I agree

bnchrch (approved):

PR #287, opened 2026-06-24
refactor(http,llm): Using new Logger module from core [OUT-255]

szanata:
## Summary

- Refactored LLM and HTTP to use new `Logger` from Core module in place of `console.*`.

## Test plan

- [x] Unit tests refactored
- [x] Manually tested

### DIFF
--- a/docs/guides/packages/core.mdx
+++ b/docs/guides/packages/core.mdx
@@ -351,7 +351,6 @@ Some metadata field names are reserved because Winston, Output log hooks, or log
 - `level`
 - `message`
 - `metadata`
-- `namespace`
 - `runId`
 - `service`
 - `splat`
@@ -361,6 +360,18 @@ Some metadata field names are reserved because Winston, Output log hooks, or log
 - `workflowType`
 
 If you need to include a value with one of those meanings, use a domain-specific name instead, such as `providerMessage`, `errorStack`, or `sourceTimestamp`.
+
+### Namespace
+
+Logs from within workflows have the "Workflow" namespace, while those from activity context have "Activity".
+
+Namespace is a discrete field in the JSON output in production, and a prefix for the string message in development.
+
+The namespace can be customized by setting it in the metadata:
+
+```js
+Logger.warn('My awesome message', { namespace: 'Custom' });
+```
 
 ### Worker output
 
--- a/docs/guides/packages/core.mdx
+++ b/docs/guides/packages/core.mdx
@@ -373,6 +373,23 @@ The namespace can be customized by setting it in the metadata:
 Logger.warn('My awesome message', { namespace: 'Custom' });
 ```
 
+Use `Logger.createLogger()` when several logs should share the same namespace:
+
+```js
+const log = Logger.createLogger('LLM');
+
+log.info('Generating completion', { model: 'claude-sonnet-4' });
+log.warn('Provider request retrying', { attempt: 2 });
+```
+
+Metadata on an individual log call can still override the logger's default namespace:
+
+```js
+const log = Logger.createLogger('LLM');
+
+log.info('Using fallback provider', { namespace: 'LLM Fallback' });
+```
+
 ### Worker output
 
 The worker uses Winston for structured logging.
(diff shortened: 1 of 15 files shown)


sdk/llm/src/cost/index.spec.js
```diff
@@ -78,7 +76,6 @@ describe( 'calculateLLMCallCost', () => {
     } );
 
     expect( result ).toBeNull();
```
bnchrch, line 78: 🧪 The swap dropped the log checks instead of moving them to `Logger`, so we no longer test that these warnings fire:

- `Failed to fetch models pricing` and `Missing cost reference for model` (this file)
- the error log in the catch (this file)
- the no-cache / stale-cache warnings in `fetch_models_pricing.spec.js`
- the snake_case warning in `validations.spec.js`
- the warn / no-warn checks in `http/cost.spec.ts`

In tests `Logger.*` falls back to `console.log`, so a `console.log` spy works, or mock `@outputai/core`'s `Logger`. If we've decided log lines aren't worth testing that's fine, I'd just want it on purpose, not a side effect of the swap.

sdk/llm/src/prompt/validations.js
```diff
@@ -1,5 +1,6 @@
 import { ValidationError, z } from '@outputai/core';
 import { attributesSchema } from './block_options.js';
+import { Logger } from '@outputai/core';
```
bnchrch: 💅 `@outputai/core` gets imported twice now (line 1 and here). Merge them:
```js
import { ValidationError, z, Logger } from '@outputai/core';
```

sdk/llm/src/cost/index.js
```diff
@@ -13,13 +14,13 @@ export const calculateLLMCallCost = async ( { modelId, usage } ) => {
     const models = await fetchModelsPricing();
 
     if ( !models ) {
-      console.warn( 'Failed to fetch models pricing' );
+      Logger.warn( 'Failed to fetch models pricing' );
```
bnchrch: ❓ `Logger` has no namespace, so these land under the generic `Activity` logs with nothing saying they came from llm cost. `validations.js` keeps its `[output-llm]` prefix but these don't, so it's a bit inconsistent. Worth a short prefix here too so they stay traceable? Small.

bnchrch (commented):
Main thing: the refactor dropped the log assertions instead of pointing them at `Logger`, so a few "we warn on X" behaviors are now untested (inline). Couple of nits too.

sdk/llm/src/cost/index.spec.js
szanata: I changed on purpose, as I dont think those are worth checking...

sdk/llm/src/prompt/validations.js
szanata: My bad

sdk/llm/src/cost/index.js
szanata: I agree that this isnt ideal. I am adding namespace overwriting on the Logger.

docs/guides/packages/core.mdx
```diff
@@ -362,6 +361,18 @@ Some metadata field names are reserved because Winston, Output log hooks, or log
...
+### Namespace
+
+Logs from within workflows have the "Workflow" namespace, while those from activity context have "Activity".
+
+Namespace is a discrete field in the JSON output in production, and a prefix for the string message in development.
+
+The namespace can be customized by setting it in the metadata:
+
+```js
+Logger.warn('My awesome message', { namespace: 'Custom' });
```
bnchrch, line 373: ❓ wdyt about being able to apply name spaces across all instances? e.g.
```js
HttpLogger = new Logger("http")
HttpLogger.warn("foo")
```

or

```js
HttpLogger = Logger.createLogger("http")
HttpLogger.warn("foo")
```

Seems both more familiar, prevent us from having to remember in each package at each log line what name space to add, and would prevent having the collission issue with the key "namespace:"

szanata: I would steer away from both. I think the is very convenient to be able to just Logger.*() at any point, without dealing with instances. We could have a `.bindNamespace()` module that returns a logger "instance" with the namespace already set.

szanata: Done. Make de the default Logger an instance with a `.createLogger()` that creates another instance with bound namespace.

bnchrch (approved):

PR #289, opened 2026-06-25
feat(cli): run a workflow with ad-hoc JSON in the dev run pane

bnchrch:
## Problem

In the `output dev` run pane, the only way to run a workflow with a custom payload was `[Create new scenario]` → type a unique name → edit JSON → and on submit it **always** wrote a `scenarios/*.json` file before running. There was no way to fire a one-off payload without persisting it.

(The headless CLI already supports this — `output workflow run|start --input '{...}'` — so the gap was TUI-only.)

## Solution

- **Run pane is editor-first.** `[Enter input]` opens the JSON editor directly — no upfront name step.
- **`ctrl+r` runs the payload as-is** — nothing written to disk (reuses the existing no-write `submit()`).
- **`ctrl+s` saves & runs** — prompts for a name, writes the scenario, then runs.
- **Duplicate (`d`) is editor-first too** — seeds the editor with the source and a `_copy` default name; you only name it if you save.
- The editor opens on a friendly `{ "": "" }` template with the cursor already inside the first quotes.
- `JsonEditor` gained an optional `onSave` and a pure `initialCursor`; `RunModal` extracted pure `buildEntries` / `validateScenarioName` (unit-tested).
- dev TUI Help panel describes the `ctrl+r` / `ctrl+s` split.

## Test plan

- [x] `eslint` clean on the changed files
- [x] `tsc --noEmit -p sdk/cli/tsconfig.json` clean
- [x] `vitest` on `run_modal.spec.ts` + `json_editor.spec.ts` — 23 pass (incl. new `initialCursor` cases); the only suite failures are the known `copy_assets` dist-artifact red herring
- [ ] Manual `./run.sh dev`: `r` → `[Enter input]` → `ctrl+r` runs with **no** file under `scenarios/`; `ctrl+s` → name → saves **and** runs

## Notes for reviewers

- **`ctrl+r` run / `ctrl+s` save & run** — `ctrl+s` reads as "save", so it's the save action; plain keys insert into the JSON buffer, so both actions need a ctrl-chord. (Swapped from the initial `ctrl+s` / `ctrl+w` per review.)
- **Duplicate flow changed order** — the name is requested *after* editing (on save) rather than before. More consistent with editor-first.
- **No Ink render test** — `ink-testing-library` isn't a dependency and the repo unit-tests pure helpers, so I extracted/tested `buildEntries`, `validateScenarioName`, and `initialCursor` instead of driving keystrokes.

Closes OUT-497

### DIFF
--- a/.changeset/out-497-adhoc-json-run.md
+++ b/.changeset/out-497-adhoc-json-run.md
@@ -2,4 +2,4 @@
 "@outputai/cli": patch
 ---
 
-Add a "Run custom JSON" option to the `output dev` run pane. Edit a payload and press ctrl+s to run it as-is without saving, or ctrl+w to save it as a scenario first. Previously every ad-hoc run forced you to name and save a scenario file.
+Add an "Enter input" option to the `output dev` run pane. Edit a payload and press ctrl+r to run it as-is without saving, or ctrl+s to save it as a scenario first. Previously every ad-hoc run forced you to name and save a scenario file.
--- a/.changeset/out-497-adhoc-json-run.md
+++ b/.changeset/out-497-adhoc-json-run.md
@@ -1,0 +1,5 @@
+---
+"@outputai/cli": patch
+---
+
+Add a "Run custom JSON" option to the `output dev` run pane. Edit a payload and press ctrl+s to run it as-is without saving, or ctrl+w to save it as a scenario first. Previously every ad-hoc run forced you to name and save a scenario file.
--- a/sdk/cli/src/views/dev/modals/run_modal.spec.ts
+++ b/sdk/cli/src/views/dev/modals/run_modal.spec.ts
@@ -1,0 +1,41 @@
+import { describe, expect, it } from 'vitest';
+import { buildEntries, validateScenarioName } from './run_modal.js';
+
+describe( 'buildEntries', () => {
+  it( 'lists scenarios then the custom-JSON entry', () => {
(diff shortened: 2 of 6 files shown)


szanata:
Some UX issues:

1. The shortcut reads weird:
Ctrl+S to run and Ctrl+w to save and run is not intuitive.
How about
Ctrl+R to run and Ctrl+S to save and run?

2. When entering the screen to add a new input the new template is just a pair of brackets {}:Screenshot 2026-06-25 at 21 07 03
In the old version it was a bit friendlier with the quotes for the first `key:value` pair:image

3. When entering the screen the cursor is outside the brackets to edit. It would be better if the cursor was already positioned naturally in the position the user would start typing:Screenshot 2026-06-25 at 21 11 18

4. The label `Run Custom JSON` feels a bit brittle, maybe `Enter input`? Or something like that...

bnchrch:
Thanks @szanata — all four addressed in e7b666074:

1. **Shortcuts** — swapped to `ctrl+r` = run, `ctrl+s` = save & run (`ctrl+s` reads as "save"). Footer + Help panel updated.
2. **Template** — a new input now seeds `{ "": "" }` again instead of a bare `{}`.
3. **Cursor** — now starts inside the first `""` pair, ready to type.
4. **Label** — `[Run custom JSON]` → `[Enter input]`.

szanata (approved):
Thanks. Amazing work. 👍

PR #296, opened 2026-06-30
chore: change 'workflow dataset generate' to not use S3 directly

catsby:
## Summary

CLI commands `workflow debug` and `workflow cost` use the Output API to retrieve the tracefile, but `workflow dataset generate` used S3 directly, requiring valid AWS credentials. This PR changes that to use the API just like `debug` and `cost`.

Also updates some documentation on bucket policies.

## Test plan

- [x] unit
- [ ] manual


bnchrch (commented):
Nice cleanup. Dropping the S3 client and the four AWS env vars, and routing `--download` through the same services `debug` and `cost` use, is the right move and takes AWS creds off every user's machine. Two things to poke at before it goes in: where we handle the nullable `workflowId`, and the exit code when we produce zero datasets. Inline.

sdk/cli/src/commands/workflow/dataset/generate.ts
```diff
@@ -165,47 +166,76 @@ export default class DatasetGenerate extends Command {
...
+      return;
+    }
+
+    const ids = runs
+      .map( run => run.workflowId )
+      .filter( ( id ): id is string => Boolean( id ) );
+
+    const missing = runs.length - ids.length;
+    if ( missing > 0 ) {
+      this.warn( `Skipping ${missing} run(s) with no workflow ID.` );
```
bnchrch: ❓ We filter out runs with no `workflowId` here because `fetchWorkflowRuns` returns the auto-generated type, where `workflowId` is optional. But a run without one is useless to us, so every caller has to repeat this filter-and-warn. Could `fetchWorkflowRuns` do it once and hand back runs that always have an id?

```ts
// workflow_runs.ts
export interface WorkflowRun { workflowId: string; /* ...rest */ }
export interface WorkflowRunsResult { runs: WorkflowRun[]; skipped: number; count: number }
```

Then this method just loops `runs`, and `skipped` drives the warning. Is there ever a real run with no `workflowId`, or is that just the generated type being loose?

sdk/cli/src/commands/workflow/dataset/generate.ts
```diff
@@ -165,47 +166,76 @@ export default class DatasetGenerate extends Command {
...
 
-    const traces = await listRemoteTraces( workflowName, { limit } );
-    if ( traces.length === 0 ) {
-      this.log( 'No remote traces found.' );
+    // The trace-log endpoint always targets a workflow's latest run, so distinct
+    // runIds sharing one workflowId (continue-as-new, reset) collapse to one dataset.
+    const workflowIds = [ ...new Set( ids ) ];
+    if ( workflowIds.length === 0 ) {
+      this.log( '\nGenerated 0 dataset(s)' );
       return;
```
bnchrch: ❗ This path exits 0, but the bottom of the method exits 1 for the same result. All runs missing an id lands here and returns success, while having ids and every trace failing hits `this.error(..., { exit: 1 })`. Both are "0 datasets from N runs", so a script gets two different exit codes. Can we make them agree? I'd lean exit 1 once we found runs but made nothing.

catsby: Good point - I think this is more defensive code than anything, at this point we should have workflow ids unless Temporal returned workflow runs with no id, which sounds not possible... Our generated API types have `workflowId? string` so we're probably just guarding but in reality it will be there. 

I changed it in 9215e3a to be consistent and exit `1`

sdk/cli/src/commands/workflow/dataset/generate.ts
catsby: I think here in `generate.ts` is the only instance we want to filter those. In `runs/list.ts` and late in the `runs_panel.tsx` (from `use_poll.ts` I think?) use `fetchWorkflowRuns()` and specifically handle empty workflow ids, so I think we leave it

docs/guides/api/configuration.mdx
```diff
@@ -28,6 +28,38 @@ The API server reads these environment variables:
...
-| `OUTPUT_AWS_ACCESS_KEY_ID` | — | AWS access key for remote traces |
-| `OUTPUT_AWS_SECRET_ACCESS_KEY` | — | AWS secret key for remote traces |
+| `OUTPUT_TRACE_REMOTE_S3_BUCKET` | — | S3 bucket holding remote traces. Required for the API to serve traces from S3 |
+| `OUTPUT_AWS_REGION` | `us-west-1` | AWS region for the trace bucket |
+| `OUTPUT_AWS_ACCESS_KEY_ID` | — | AWS access key for reading remote traces |
+| `OUTPUT_AWS_SECRET_ACCESS_KEY` | — | AWS secret key for reading remote traces |
+
+### AWS Credential Permissions
+
+The API reads remote traces from S3 on behalf of the CLI — `output workflow debug`, `output workflow cost`, and `output workflow dataset generate --download` all call the API's trace-log endpoint rather than talking to S3 directly. The CLI itself needs no AWS credentials.
```
szanata: I think it is odd saying what CLI does in the `api/configurations.mdx` I think this statement:

> [...] `output workflow debug`, `output workflow cost`, and `output workflow dataset generate --download` all call the API's trace-log endpoint rather than talking to S3 directly. The CLI itself needs no AWS credentials.

Belongs on the CLI documentation.

sdk/cli/src/commands/workflow/dataset/generate.ts
szanata: Still, I think it is odd to have to filter runs without workflowIds here, I would agree with Ben, and think we should remove them upstream

sdk/cli/src/commands/workflow/dataset/generate.ts
```diff
@@ -165,47 +166,75 @@ export default class DatasetGenerate extends Command {
...
+    }
+
+    this.log( `Found ${workflowIds.length} run(s). Fetching traces...` );
 
     const dir = await resolveDefaultDatasetsDir( workflowName );
 
-    for ( const trace of traces ) {
-      const traceData = await downloadRemoteTrace( trace.key );
-      const extracted = extractDatasetFromTrace( traceData );
+    const outcomes = await Promise.all( workflowIds.map( id => this.generateFromRun( id, dir ) ) );
```
szanata: outcomes is a odd var name, could this be collapse into the same command?

```ts
await Promise.all( workflowIds.map( id => this.generateFromRun( id, dir ) ) )
  .filter( dataset => !!dataset ); // or filter( Boolean ), although I find that not very good.
``` ?

docs/guides/api/configuration.mdx
catsby: I re-worded this to not mention the CLI at all, but leave documentation that the API needs to be able to read from the bucket in order to serve tracefiles.

sdk/cli/src/commands/workflow/dataset/generate.ts
catsby: All of this seems to be defensive coding around the idea of Temporal returning workflow execution runs that have no workflow ID, which I think Temporal guarantees, so I did go and move the filtering in to `fetchWorkflowRuns()` and removed some of the defensive fall backs

szanata (approved):

PR #297, opened 2026-06-30
fix(core,http,llm): Disabling undici h2 and fixing proxy support for llm [OUT-506][OUT-507][OUT-505]

szanata:
### Core
- Disabled HTTP/2 (`allowH2: false`) in the global fetch dispatcher configured by `setGlobalDispatcher` when proxy env vars are detected;
- Disabled HTTP/2 (`allowH2: false`) in the webhook functions `sendHttpRequest` and `sendPostRequestAndAwaitWebhook` by using a dispatcher (`EnvHttpProxyAgent`).

### HTTP
- Added a custom dispatcher that disables HTTP/2 (`allowH2: false`) on fetch, it uses `EnvHttpProxyAgent`;
- Added support for dispatcher in the init argument of fetch; it has precedence over the custom dispatcher.

### LLM
- Disabled HTTP/2 (`allowH2: false`) in the dispatcher of the fetch client used when consuming the AI SDK and fetching model pricing;
- Replaced the `Agent` dispatcher in favor of `EnvHttpProxyAgent` to respect the proxy env vars. [OUT-506].

## Notes
- `EnnHttpProxyAgent` behaves like a normal `Agent` if there isnt proxy env vars.
- `boostrapFetchProxy` seems to be useless now, but it will not be removed yet.

## Test plan

- [x] Manual tested agains Temporal Cloud with production workflows
- [x] Added unit tests


sdk/llm/src/cost/fetch_models_pricing.js
```diff
@@ -1,8 +1,12 @@
 import { Logger } from '@outputai/core';
+import { EnvHttpProxyAgent, fetch } from 'undici';
 
 const costTableUrl = 'https://models.dev/api.json';
 const cacheTTL = 1000 * 60 * 60 * 24; // 1 day
 
+/** Custom dispatcher to disable HTTP/2. See [OUT-505] */
```
catsby: nit: generally we should not publicly link to private resources

sdk/llm/src/cost/fetch_models_pricing.js
```diff
@@ -25,7 +29,7 @@ export const fetchModelsPricing = async () => {
   if ( cache.content && cache.expiresAt > Date.now() ) {
     return cache.content;
   }
-  const res = await fetch( costTableUrl );
+  const res = await fetch( costTableUrl, { dispatcher } );
```
catsby: should we `try/catch` this and return the `cache.content` for network errors (timeout, DNS failure, etc)?

sdk/http/src/fetch/index.ts
```diff
@@ -50,8 +53,9 @@ export const fetch = async ( input: RequestInfo | Request, init?: RequestInit )
 
   await logRequest( { requestId, request } );
 
+  const dispatcher = init?.dispatcher ?? customDispatcher;
   try {
-    const response = await undici.fetch( request );
+    const response = await undici.fetch( request, dispatcher ? { dispatcher } : undefined );
```
catsby, line 59: at this point isn't `dispatcher` always truthy? It's always set on line 56 right?

sdk/core/src/worker/proxy.js
```diff
@@ -5,28 +5,16 @@ const log = createChildLogger( 'Proxy' );
 
 /**
  * Routes all `fetch()` calls (including those inside Temporal activities)
- * through an HTTP/HTTPS proxy when standard proxy env vars are set
- * (`HTTPS_PROXY`, `https_proxy`, `HTTP_PROXY`, `http_proxy`). No-op when
- * none are set. Invalid URLs are logged and skipped so the worker keeps
- * running.
- *
- * Call once at worker startup, before any network activity.
```
catsby, line 13: > Call once at worker startup, before any network activity.

nit: Is it no longer recommend to call this before any network activity?

catsby (approved):
Mostly nit picks you can take or leave. Great work here digging in to find this regression

sdk/core/src/worker/proxy.js
szanata: I removed this because this isnt entirely, true, depends where this is called, it doesnt know.

sdk/http/src/fetch/index.ts
szanata: I replaced so the user actually can set it to `undefined` (explicitly) and it will disable the dispatcher.

sdk/llm/src/cost/fetch_models_pricing.js
szanata: Replaced with a gh issue

sdk/llm/src/cost/fetch_models_pricing.js
szanata: Fixed.

PR #302, opened 2026-07-02
chore: Adding docs explain how users should drain fetch responses [OUT-509]

szanata:
## Summary

Fetch responses that doesn't have the body consumed (.text(), .json(), etc) might cause memory problems. This PR add docs explain that users should do that.

### DIFF
--- a/docs/guides/packages/http.mdx
+++ b/docs/guides/packages/http.mdx
@@ -176,6 +176,8 @@ const inputCost = Number( response.headers.get( 'x-cost-input-usd' ) ?? 0 );
 const outputCost = Number( response.headers.get( 'x-cost-output-usd' ) ?? 0 );
 
 addRequestCost( response, inputCost + outputCost );
+
+response.body?.cancel();
 ```
 
 `addRequestCost` accepts the total request cost as a number.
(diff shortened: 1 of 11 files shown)


docs/guides/packages/http.mdx
```diff
@@ -54,6 +54,26 @@ await client.put('companies/123', {
...
+
+```typescript
+const response = await client.get('https://example.com', {
+  redirect: 'follow'
+});
+
+try {
+  return response.url;
+} finally {
+  await response.body?.cancel();
```
catsby, line 73: Does something like this apply to the `Error Handling` section lower in this guide?

catsby: Same goes for the `Attaching Request Cost` section, where we do `response.headers.get()` but never read the body.

catsby (approved):
We might want to also update `test_workflow/http_simple`. Our `CLAUDE.md` file instructs Claude to "Follow patterns in `test_workflows/` for new workflows" but we don't have an example of draining there.

docs/guides/packages/http.mdx
szanata: 1. No

2. Yes, I have added it

szanata:
I have added test_workflows example.

PR #304, opened 2026-07-02
refactor(api): Improving start/run workflow performance [OUT-511]

szanata:
## Summary

When starting a workflow in the API, via `run`/`start` endpoints, it validates the `workflowName`argument by sending a Temporal query to a running workflow called catalog that contains all supported workflows for this task queue.

This have performance drawbacks as, if the workers are already strangled, there might not be an available room to take this processing.

This PR changes that by storing the valid workflows names in a `memo` field in that catalog workflow, so it can be described via Temporal API without the necessity of a worker. Behavior remains the same.

## Test plan

- [x] Add unit tests
- [x] Manually tested

### DIFF
--- a/.changeset/quick-hats-build.md
+++ b/.changeset/quick-hats-build.md
@@ -1,0 +1,5 @@
+---
+"output-api": patch
+---
+
+Improving start/run workflows performance by not querying an workflow to validate `workflowName` argument.
--- a/api/src/clients/temporal/catalog.js
+++ b/api/src/clients/temporal/catalog.js
@@ -1,46 +1,36 @@
-import { WorkflowNotFoundError, CatalogNotAvailableError } from '../errors.js';
+import { CatalogNotAvailableError, UnsupportedWorkflowError, WorkflowNotFoundError } from '../errors.js';
 import { logger } from '#logger';
 
-/**
- * Returns the catalog object from the catalog workflow
- *
- * @param {Client} client
- * @returns {object}
- * @throws {CatalogNotAvailableError}
- * @throws {Error}
- */
-export const getCatalog = async ( { client, taskQueue } ) => {
-  const catalogHandle = client.workflow.getHandle( taskQueue );
+const getAvailableWorkflows = async ( { client, taskQueue } ) => {
+  const handle = client.workflow.getHandle( taskQueue );
+
   try {
-    return await catalogHandle.query( 'get' );
+    const description = await handle.describe();
+    return description.memo?.workflowNames ?? {};
   } catch ( error ) {
     if ( error instanceof WorkflowNotFoundError ) {
-      throw new CatalogNotAvailableError( 3 );
+      throw new CatalogNotAvailableError( 3, taskQueue );
     }
-    // Annotate the catalog/query context (the only place that knows it) so the error_handler can
-    // surface it when it logs the failure centrally.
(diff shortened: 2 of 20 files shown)


api/src/clients/temporal/catalog.js
```diff
@@ -1,46 +1,36 @@
...
- */
-export const getCatalog = async ( { client, taskQueue } ) => {
-  const catalogHandle = client.workflow.getHandle( taskQueue );
+const getAvailableWorkflows = async ( { client, taskQueue } ) => {
+  const handle = client.workflow.getHandle( taskQueue );
+
   try {
-    return await catalogHandle.query( 'get' );
+    const description = await handle.describe();
+    return description.memo?.workflowNames ?? {};
```
catsby, line 9: do we need to check the `closeTime` around here like we do in `catalog_job`?

https://github.com/growthxai/output/blob/9d7a8708d9f9d13f91895263f90f583568239828/sdk/core/src/worker/catalog_workflow/catalog_job.js#L48-L51

api/src/middleware/error_handler.js
```diff
@@ -28,6 +34,7 @@ const NAMED_ERROR_STATUSES = {
   [TraceNotAvailableError.name]: 404,
   [WorkflowExecutionTimedOutError.name]: 408,
   [StepNotCompletedError.name]: 409,
+  [UnsupportedWorkflowError.name]: 424,
```
catsby, line 37: Previously we threw `WorkflowNotFoundError` which is a `404`, but now for the same scenario we're throwing a `424`. Just double checking that this change is intentional

sdk/core/src/worker/catalog_workflow/catalog.js
```diff
@@ -20,6 +25,11 @@ export class Catalog {
    */
   addWorkflow( workflow ) {
     this.workflows.push( workflow );
+    this.workflowNames[workflow.name] = workflow.name;
+    for ( const alias of workflow.aliases ) {
+      this.workflowNames[alias] = workflow.name;
```
catsby, line 30: I don't know how this worked before, but if we have Workflow A with no aliases, and Workflow B with an alias of "A", will this then overwrite the record of A? It seems like we could overwrite things here

api/src/clients/temporal/catalog.js
szanata: No. It is ok if the catalog is not running for some reason.

api/src/middleware/error_handler.js
szanata: Yes. Seems proper to me that this new error is a validation error

sdk/core/src/worker/catalog_workflow/catalog.js
szanata: Alias and workflow names are unique. You cant have an alias that is also a workflow name. This is validated during loading time.

catsby (approved):

.changeset/quick-hats-build.md
```diff
@@ -0,0 +1,5 @@
+---
+"output-api": patch
```
catsby, line 2: nit: shouldn't this mention `output-sdk` as well?

szanata: I have added another changeset

PR #307, opened 2026-07-04
refactor(core,api,cli): Optimizing trace destinations [OUT-514][OUT-516]

szanata:
### Trace changes
- Trace destinations keys are now sparse:
  _Before:_
  ```json
  {
    ...
    "trace": {
      "destinations": {
        "local": null,
        "remote": null
      }
    }
  }
  ```
  _After:_
  ```json
  {
    ...
    "trace": {
      "destinations": {}
    }
  }
- Optimizing calls to `getTraceDestinations` activity: do not call it if the workflow has `disableTrace: true` flag.
- Adding internal activities to trace files: they are part of the workflow and can cause breakage, so they must be part of the trace file.

### sendHttpRequest and sendPostRequestAndAwaitResponse header arguments change
- Both `sendHttpRequest` and `sendPostRequestAndAwaitResponse` now accept environment variable interpolation in its headers values:
  ```js
  sendHttpRequest( {
    url,
    headers: {
      Authorization: 'Bearer $AUTH_TOKEN'
    }
  } );
  ```
  The token `$AUTH_TOKEN` will be replaced by `process.env.AUTH_TOKEN` value when sending the request.

### sendHttpRequest body/headers omission/redacting
- `sendHttpRequest` now omits body and headers from response, avoiding having these values persisted to trace or Temporal history. They can still be included using the `responseOptions` argument:
  ```js
  sendHttpRequest( {
    url,
    responseOptions: {
      includeHeaders: true,
      includeBody: true
    }
  } );
  ```
- All response headers included via `responseOptions.includeHeaders` are redacted by header name. This covers common sensitive header names such as authorization, token, secret, password, cookie, and key, but it is a best-effort heuristic.

## Test plan

- [x] Manual tests
- [x] Unit tests


.changeset/spicy-drinks-win.md
```diff
@@ -0,0 +1,29 @@
...
+    }
+  }
+  ```
+  _After:_
+  ```json
+  {
+    "output": "foo",
+    "trace": {
+      "destinations": {}
+    }
```
catsby, line 26: nit: why `"destinations"` at all? why not just `"trace": {}`?

.changeset/spicy-drinks-win.md
```diff
@@ -0,0 +1,29 @@
...
+  }
+  ```
+- Internal activities like `getTraceDestinations` and `sendHttpRequest` are no longer omitted in the trace files.
```
catsby, line 29: ❓ Do we omit or sanitize sensitive items that might be in HTTP headers, etc?

docs/guides/migrations/v0.9.0-to-v0.10.0.mdx
catsby, line 1: There's a tracing specific page we might need to update:

- https://github.com/growthxai/output/blob/1ad3dbce15bbeb404e5577954b88115b51d7c081/docs/guides/operations/tracing.mdx

api/src/clients/temporal/workflow/run.js
```diff
@@ -41,11 +41,7 @@ export const run = async ( { client }, workflowName, input, options = {} ) => {
   } catch ( error ) {
     // Workflow failures are returned as data, not thrown
     if ( error instanceof WorkflowFailedError ) {
-      logger.warn( 'Workflow execution failed', {
-        workflowId,
-        errorMessage: error.message,
-        hasTrace: Boolean( extractErrorDetail( error, 'trace' ) )
```
catsby, line 47: should we still indicate in the log if a trace exists or not? previously operators would at least know if a trace was available

.changeset/spicy-drinks-win.md
szanata: Because it is easier to read the var values:
If the value is `trace: { destination: {} }`, then:
`trace.destinations.foo === null`

But if the value is `trace: {}`, then:
`trace.destinations.foo // throw TypError`

.changeset/spicy-drinks-win.md
szanata: You are right.

api/src/clients/temporal/workflow/run.js
szanata: I have updated the code to handle both input/output. I also documented this and put some warnings about this feature. It has to be used properly so to no expose sensitive information as it is a convenience feature.

docs/guides/migrations/v0.9.0-to-v0.10.0.mdx
szanata: I have updated the file.

docs/guides/operations/tracing.mdx
```diff
@@ -55,6 +60,31 @@ OUTPUT_AWS_SECRET_ACCESS_KEY=...
...
+    {
+      "Effect": "Allow",
+      "Action": [ "s3:PutObject", "s3:AbortMultipartUpload" ],
+      "Resource": "arn:aws:s3:::my-traces/*"
+    }
+  ]
+}
+```
+
+The API server reads those same traces back to serve `output workflow debug`, `cost`, and `dataset generate --download` — see [API Configuration](/api/configuration#aws-credential-permissions) for the read-side policy. The worker and API can share one key pair or use two identities each scoped to their own actions.
```
catsby: Is this duplication of the `### AWS Credential Permissions` just below it?

sdk/core/src/interface/webhook.js
```diff
@@ -38,6 +41,7 @@ export async function sendHttpRequest( { url, method = 'GET', payload = undefine
  * @param {string} url
  * @param {unknown} [payload]
  * @param {object} [headers]
+ * @param {object} [options.responseOptions] - Response options
```
catsby: are we still including these in the return? the function signature suggests we are not

sdk/core/src/worker/tuner_options.example.js
catsby, line 1: how is this file used?

catsby (approved):
Pre-approving with just a few minor things that should be cleaned up before merging

sdk/core/src/worker/tuner_options.example.js
szanata: I commited by mistake

PR #308, opened 2026-07-05
refactor(core): Drop support for legacy activity/workflow executions [OUT-468]

szanata:
## Summary

- Removing support to execute non response-wrapped activities and legacy `startChildWorkflow()` child workflow executions (<= v0.7.x). Those were kept to allow replay of old workflow executions, but the 30 days agreement on that has expired.

## Test plan

- [x] Refactored unit tests
- [x] Manually tested

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ agent-os/
 
 # Pnpm publish report
 pnpm-publish-summary.json
+
+docker-compose.remote.yml
--- a/sdk/core/src/interface/workflow.spec.js
+++ b/sdk/core/src/interface/workflow.spec.js
@@ -373,30 +373,6 @@ describe( 'workflow()', () => {
       expect( getTraceDestinations ).toHaveBeenCalledWith( info.memo.traceInfo );
     } );
 
-    it( 'falls back to workflow type matching when replaying an old child call without memo.stack', async () => {
-      const { workflow } = await import( './workflow.js' );
-      const { ParentClosePolicy } = await import( '@temporalio/workflow' );
-      setWorkflowInfo( { workflowType: 'old_parent_wf', memo: { traceInfo: { workflowId: 'root-workflow' } } } );
-      executeChildMock.mockResolvedValueOnce( { output: { child: 'replayed' } } );
-      const fn = vi.fn();
-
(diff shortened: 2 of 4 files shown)


sdk/core/src/interface/workflow.spec.js
```diff
@@ -373,30 +373,6 @@ describe( 'workflow()', () => {
...
-      expect( fn ).not.toHaveBeenCalled();
-      expect( executeChildMock ).toHaveBeenCalledWith( 'old_child_wf', expect.objectContaining( {
-        args: [ { id: 7 } ],
-        parentClosePolicy: ParentClosePolicy.TERMINATE,
-        memo: { traceInfo: { workflowId: 'root-workflow' } }
-      } ) );
-      expect( proxyActivitiesMock ).not.toHaveBeenCalled();
-    } );
-
     it( 'does not fallback to child execution when the replayed workflow type matches an alias', async () => {
```
catsby, line 376: nit: should we rename and adjust this test, since `child execution` is no longer a thing?

.gitignore
```diff
@@ -27,3 +27,5 @@ agent-os/
 
 # Pnpm publish report
 pnpm-publish-summary.json
+
+docker-compose.remote.yml
```
catsby, line 31: nit: is this intentional?

sdk/core/src/interface/workflow.spec.js
catsby, line 5: can we get rid of this here? It's also mentioned around line 79 of this file

catsby (approved):
minor nits for cleanup

sdk/core/src/interface/workflow.spec.js
szanata: We still have child workflows, we dont use the legacyn `startChildWorkflow()` function, they are invoked differently now.

.gitignore
szanata: Yes, this is a test file I created. I plan to add support to it later, for now I dont want to commit as it is not ideal.

sdk/core/src/interface/workflow.spec.js
szanata: this is part of the activity response payload. The test reflects the current code.

PR #310, opened 2026-07-07
feat(core): Add worker tuner options support [OUT-510]

szanata:
## Summary

Adding support to Temporal worker `tuner` options (https://docs.temporal.io/develop/worker-tuning-reference). The config is provided using an environment variable `TEMPORAL_WORKER_TUNER` option, its value is parsed as JSON. Example:
```bash
TEMPORAL_WORKER_TUNER='{
  "workflowTaskSlotSupplier": {
    "type": "fixed-size",
    "numSlots": 10
  },
  "activityTaskSlotSupplier": {
    "type": "resource-based",
    "tunerOptions": {
      "targetMemoryUsage": 0.8,
      "targetCpuUsage": 0.9
    },
    "minimumSlots": 1,
    "maximumSlots": 100,
    "rampThrottle": "50ms"
  },
  "localActivityTaskSlotSupplier": {
    "type": "fixed-size",
    "numSlots": 10
  },
  "nexusTaskSlotSupplier": {
    "type": "fixed-size",
    "numSlots": 10
  }
}'
```

## Test plan

- [x] Added unit tests
- [x] Manually tested

### DIFF
--- a/sdk/core/src/worker/configs.js
+++ b/sdk/core/src/worker/configs.js
@@ -1,9 +1,34 @@
 import * as z from 'zod';
 import { isStringboolTrue } from '#helpers/string';
+import { isPlainObject } from '#helpers/object';
 
 class InvalidEnvVarsErrors extends Error { }
 
 const coalesceEmptyString = v => v === '' ? undefined : v;
+
+const jsonObjectEnvSchema = z.preprocess(
+  coalesceEmptyString,
+  /* eslint-disable consistent-return */
+  z.string().optional().transform( ( value, ctx ) => {
+    if ( value === undefined ) {
+      return undefined;
+    }
+
+    try {
+      const parsed = JSON.parse( value );
+      if ( !isPlainObject( parsed ) ) {
+        ctx.addIssue( { code: 'custom', message: 'Expected a JSON representing plain JS object' } );
+        return z.NEVER;
+      }
+
+      return parsed;
+    } catch ( error ) {
+      ctx.addIssue( { code: 'custom', message: `Expected valid JSON: ${error.message}` } );
+      return z.NEVER;
+    }
+  } )
+  /* eslint-enable consistent-return */
+);
 
 const durationSchema = z.preprocess(
   coalesceEmptyString,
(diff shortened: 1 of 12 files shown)


derencius (approved):

sdk/core/src/worker/configs.js
```diff
@@ -1,10 +1,35 @@
...
+      }
+
+      return parsed;
+    } catch ( error ) {
+      ctx.addIssue( { code: 'custom', message: `Expected valid JSON: ${error.message}` } );
+      return z.NEVER;
+    }
+  } )
+  /* eslint-enable consistent-return */
+);
```
catsby: I think we should add some validation with zod schemas here. If I read things correctly, the exact shapes (`WorkerTuner`, `ResourceBasedTuner`) are exported, but the function `asNativeTuner` that would normally validate them is not... I think it's worth adding our own to validate the shape here, otherwise I think if there's a typo or bad shape that's otherwise valid JSON, I think Temporal will throw an error and we might not have the right information to share at that point, possibly leaving users confused as to what is wrong with the shape they've given.

catsby: Follow-up - Temporal might not throw an error, but they may also just drop things. Since it's 2 different shapes but we present 1 env var, users might think to combine them, like so:

```json
{
  "workflowTaskSlotSupplier": {
    "type": "fixed-size",
    "numSlots": 10
  },
  "activityTaskSlotSupplier": {
    "type": "resource-based",
    "tunerOptions": {
      "targetMemoryUsage": 0.8,
      "targetCpuUsage": 0.9
    },
    "minimumSlots": 1,
    "maximumSlots": 100,
    "rampThrottle": "50ms"
  },
  "localActivityTaskSlotSupplier": {
    "type": "fixed-size",
    "numSlots": 10
  },
  "nexusTaskSlotSupplier": {
    "type": "fixed-size",
    "numSlots": 10
  },
  "workflowTaskSlotOptions": {
    "minimumSlots": 2,
    "maximumSlots": 1000,
    "rampThrottle": "10ms"
  }
}
```

All the required things as per our docs say about a `TunerHolder` shape, but also included `workflowTaskSlotOptions` from `ResourceBasedTuner`, which I believe Temporal will just silently drop. You could argue that it doesn't make sense to combine them, but our docs don't make that clear and the 1 env supporting 2 shapes make it confusing

szanata: I have improved the docs and added schema validation.

catsby (approved):

PR #311, opened 2026-07-08
feat: Workflow Monitor CLI command

catsby:
## Summary

Adds a new CLI command, `workflow monitor [id]` that monitors the execution of a workflow by ID.

- tails and emits step status updates as they happen (can attach to executing workflows)
- uses Workflow History endpoint, and adds a resume cursor to avoid re-reading the entire history each poll
- follows `continue-as-new` chains to the new run automatically
- `--format json` emits `NDJSON`, one event per line, for consumption by scripts/agents
- on workflow completion/error, exit code matches result (`0` for success, `1` for failures, `130` on Ctrl+C detach)
- retries transient poll failures (network blips, 5xx) up to 5 times before giving up; fails fast on non-transient errors

## Test plan

- unit tests: monitor command loop (313 lines), history service paging/resume/drain, long-poll handler, color/log utils
- manual: `output workflow monitor <id>` against a running workflow — attach mid-run, idle long-poll, continue-as-new follow, failure exit code

Local testing: 

- clone repo, checkout branch `out_482_workflow_monitor`
- in a terminal session (1), run: `$ ./run.sh dev`
- in another session, run a workflow and record the workflow ID:

```console
WORKFLOW_ID=$(curl -sS -f -X POST http://localhost:3001/workflow/start \
  -H "Content-Type: application/json" \
  -d '{"workflowName":"sleep","input":{"urls":["https://a.com","https://b.com","https://c.com","https://d.com"],"delayMs":5000}}' \
  | jq -er '.workflowId')
```

- use the `WORKFLOW_ID` in the `workflow monitor` command:

```console
output workflow monitor ${WORKFLOW_ID}
```

of use `json`: 

```console
output workflow monitor ${WORKFLOW_ID} --format json
```

Example plain:

https://github.com/user-attachments/assets/5ec87cc7-636c-424d-98df-b7a7edbcef57

Example JSON:

https://github.com/user-attachments/assets/84b7474c-5a02-4ad7-a364-a2a7a0cf1dfb

### DIFF
--- a/api/src/clients/temporal/workflow/get_history.js
+++ b/api/src/clients/temporal/workflow/get_history.js
@@ -46,8 +46,12 @@ export const getHistory = async ( { client, connection }, workflowId, options = 
     throw new InvalidPageTokenError();
   }
 
-  const isFirstPage = !pageToken;
-  const metadata = isFirstPage ? await ( async () => {
+  // Re-describe on every `wait` call, not just the first page: a resumable poller
+  // (`workflow monitor`) never requests the first page again, so if metadata were only
+  // fetched there, `workflow.status` would freeze at whatever it was on the very first
+  // poll and the caller could never observe the run finishing.
+  const shouldDescribe = !pageToken || wait;
+  const metadata = shouldDescribe ? await ( async () => {
     const { workflow } = await describeWorkflow( { client }, workflowId, { runId } );
     return { workflow, resolvedRunId: workflow.runId };
   } )() : { workflow: null, resolvedRunId: runId };
--- a/api/src/clients/temporal/workflow/get_history.js
+++ b/api/src/clients/temporal/workflow/get_history.js
@@ -1,3 +1,4 @@
+import { temporal as temporalConfig } from '#configs';
 import { InvalidPageTokenError } from '../../errors.js';
 import { decodeEventPayloads, serializeEvent } from '../../event_serialization.js';
 import { describeWorkflow } from './describe_workflow.js';
@@ -32,10 +33,14 @@ import { fetchHistoryPage } from './fetch_history_page.js';
  * @param {object} options.pageSize - Amount of results default=20
  * @param {object} options.pageToken - Used to retrieve next page, must be used together with runId
  * @param {object} options.includePayloads - Omit/display payloads, default=false
+ * @param {boolean} [options.wait] - Long-poll for a new event when already caught up to the
+ *   end of history, bounded by `temporal.historyWaitTimeoutMs`, instead of returning
+ *   immediately. Lets a resumable poller (`workflow monitor`) avoid restarting from page 1
+ *   on every tick.
  * @returns {WorkflowHistoryResults}
  */
 export const getHistory = async ( { client, connection }, workflowId, options = {} ) => {
-  const { runId, pageSize = 20, pageToken, includePayloads = false } = options ?? {};
+  const { runId, pageSize = 20, pageToken, includePayloads = false, wait = false } = options ?? {};
 
   if ( pageToken && !runId ) {
     throw new InvalidPageTokenError();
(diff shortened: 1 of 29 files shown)


sdk/cli/src/commands/workflow/monitor.ts
```diff
@@ -0,0 +1,265 @@
...
+    );
+
+    const sigintHandler = (): void => {
+      emit( { detached: true }, '\nDetached (the workflow keeps running).' );
+      process.exit( SIGINT_EXIT_CODE );
+    };
+    process.on( 'SIGINT', sigintHandler );
+
+    try {
+      for ( ; ; ) {
```
szanata: No love for `while`?

api/src/handlers/workflow_history.js
```diff
@@ -19,15 +25,17 @@ export function createWorkflowHistoryHandler( client ) {
   return async ( req, res ) => {
     const workflowId = req.params.id;
     const pathRunId = readPinnedRunId( req );
-    const { runId, pageSize, pageToken, includePayloads } = querySchema.parse(
+    const { runId, pageSize, pageToken, includePayloads, wait, waitMs } = querySchema.parse(
```
szanata: I honestly think `wait` and `waitMs` could be combined into single variable on all call places (internal/api signature). I think `wait` and `waitMs` is a bit too generic and redundant.

Some suggestions: `collectForMs` or `longPollTimeoutMs`.

If set and value > 0, do long pool. Otherwise dont.

sdk/cli/src/utils/monitor_log.ts
```diff
@@ -0,0 +1,61 @@
...
+ * transitions as they're observed on each poll.
+ */
+import type { Span, SpanStatus } from '#services/workflow_history/correlator.js';
+import { ANSI, formatDurationLabel, makeTint } from '#utils/waterfall.js';
+
+const GLYPH: Record<SpanStatus, string> = {
+  pending: '·',
+  running: '●',
+  completed: '✓',
+  failed: '✗'
```
szanata, line 14: Maybe it is better to have continued as new workflows have its own glyph here instead of adding `↻ ` to the message itself, having all the business formatting logic combined.

api/src/configs.js
```diff
@@ -12,6 +12,10 @@ const envSchema = z.object( {
   TEMPORAL_NAMESPACE: z.string().optional().default( 'default' ),
   TEMPORAL_WORKFLOW_EXECUTION_TIMEOUT: z.string().optional().default( '24h' ),
   TEMPORAL_WORKFLOW_EXECUTION_MAX_WAITING: z.coerce.number().optional().default( 300_000 ), // 5minutes
+  // Bounds a single long-poll (waitNewEvent) call on GET /workflow/:id/history so a resumable
+  // poller (e.g. `workflow monitor`) gets a fast, incremental response instead of restarting
+  // its history fetch from page 1 on every tick.
+  TEMPORAL_HISTORY_WAIT_TIMEOUT_MS: z.coerce.number().optional().default( 15_000 ),
```
szanata: This var name and description is wrong.

This should be TEMPORAL_HISTORY_MAX_WAIT_TIMEOUT_MS as it only works if the user passed `wait=true` in the request. It will cap the user's `waitFor` so it cant exceed this value.

api/src/clients/temporal/workflow/get_history.js
```diff
@@ -32,27 +33,54 @@ import { fetchHistoryPage } from './fetch_history_page.js';
  * @param {object} options.pageSize - Amount of results default=20
  * @param {object} options.pageToken - Used to retrieve next page, must be used together with runId
  * @param {object} options.includePayloads - Omit/display payloads, default=false
+ * @param {boolean} [options.wait] - Long-poll for a new event when already caught up to the
+ *   end of history, bounded by `temporal.historyWaitTimeoutMs`, instead of returning
+ *   immediately. Lets a resumable poller (`workflow monitor`) avoid restarting from page 1
```
szanata: We should not mention workflow monitor here. API and CLI are different projects, the API should only mention what it does, not who uses it.

szanata (commented):
I left some notes I think could be addressed.

Another question I have is, why not make this a stream or SSE?

To me at least, having a request that takes some time to resolve because it is accumulating events feels a bit "hacky", If the objective is always have new events without going over the whole history again, streaming or SSE make much more sense.

sdk/cli/src/commands/workflow/monitor.ts
catsby: Changed to use `while` in d7cf55a

api/src/handlers/workflow_history.js
catsby: Changed in d7cf55a

api/src/configs.js
catsby: changed in d7cf55a

api/src/clients/temporal/workflow/get_history.js
catsby: Scrubbed this and other `workflow history` in d7cf55a

sdk/cli/src/utils/monitor_log.ts
catsby: I split continued as new out to it's own helper function to construct the string message. I split it from the other glyphs because it's not a true span status as we've been using in our functions that correlate Temporal activity events to status we show

szanata (approved):

PR #312, opened 2026-07-08
refactor(core): Refactoring step call to remove workflow rewrite [OUT-488]

szanata:
## Summary
- Refactored worker workflow rewrite and `workflow()` interface to replace `this.invokeStep()` / `this.invokeEvaluator()` style activity dispatchers on the workflow handler, in favor of a `globalThis` function created during runtime:
  - The old approach required rewriting workflow functions to convert arrow functions to regular functions (to have `this`), which was a source of complexity and possible edge cases;
  - The new function defined on `globalThis` takes advantage of Temporal sandbox isolation;
  - Invoking activities via a simple `globalThis` function is much easier to rewrite in the webpack loader:
  _Given this user code:_
    ```js
    import { foo } from 'step.js';

    workflow( {
      name: 'test',
      fn: async () => {
        await foo();
      }
    } );
    ```
    _The loader will now rewrite internally to:_
    ```js
    workflow( {
      name: 'test',
      fn: async () => {
        await globalThis[globalThis.Symbol.for('@outputai/core:__invoke_activity')]( 'stepName' );
      }
    } );
    ```
    _Versus how it was before:_
    ```js
    workflow( {
      name: 'test',
      fn: async function () {
        await this.invokeStep( 'stepName' );
      }
    } );
    ```
- Refactored child workflow detection to use the same `globalThis` mechanism: if the dispatcher function was already defined, the call is a nested workflow call and should start a child workflow.
- Removed most of the workflow body rewrite during worker start;
- Removed distinction between step, evaluator, shared step and shared evaluator during rewrite, treating them all as simple activities;
- Added top-level activity call validation, preventing steps and evaluators from being invoked outside functions:
  ```js
  import { foo } from 'step.js';

  foo();
  ```
  _This throws an error._;
- Added validation to prevent `globalThis` shadowing in rewritten activity call sites;
- Added validation requiring steps/evaluators files to use named exports only, and requiring workflow imports from steps/evaluators files to use named imports/destructured requires only;
  ```js
  // valid imports
  import { foo } from './steps.js';
  const { bar } = require( './evaluators.js' );

  // invalid imports
  import foo from './steps.js';
  import * as steps from './steps.js';
  const steps = require( './steps.js' );
  ```
  ```js
  // valid exports
  export const foo = step( { name: 'foo' } );

  // invalid exports
  export default step( { name: 'foo' } );
  export * from './other_steps.js';
  ```
- Removed `"$shared"` activity namespace in favor of loading each shared activity into each workflow namespace, further simplifying AST rewrite and activity invocation, as all activities are handled equally. This adds a new constraint that workflow activities cannot have the same name as shared activities, which is a fair trade-off.
- Refactored `activityOptions` resolution chain and resolution order, making it easier to read and understand the precedence of the options:
  - Old order: `invocation > definition > parent > default`
  - New order: `invocation > parent > definition > default`
- Refactored child `activityOptions` propagation
  - Old behavior: Parent workflow resolved the options chain with the child invocation options and started child workflow creating a new memo with these `activityOptions`;
  - New behavior: Parent resolves its own `activityOptions` and sets to `memo` as `parentActivityOptions`, when it invokes a child, it forwards its `memo` and the original `invocationOptions.activityOptions` arguments, so the child will resolve their own options chain.
- Updated `Objects.deepMerge()` to support merging multiple objects from left to right, with later objects taking precedence.

### DIFF
--- a/sdk/core/src/helpers/object.js
+++ b/sdk/core/src/helpers/object.js
@@ -59,18 +59,24 @@ export const deepMergeWithResolver = ( a, b, resolver ) => {
 };
 
 /**
- * Creates a new object merging object "b" onto object "a" biased to "b":
- * - Object "b" will overwrite fields on object "a";
- * - Object "b" fields that don't exist on object "a" will be added;
- * - Object "a" fields that don't exist on object "b" will be preserved;
+ * Creates a new object recursively merging the rightmost object over the previous one.
+ * Give two objects, (L)eft and (R)right
+ * - Object "R" will overwrite fields on object "L";
+ * - Object "R" fields that don't exist on object "L" will be added;
+ * - Object "L" fields that don't exist on object "R" will be preserved;
  *
- * If "b" isn't an object, a new object equal to "a" is returned
+ * If "R" isn't an object, a new object equal to "L" is returned
  *
  * @param {object} a - The base object
  * @param {object} b - The target object
  * @returns {object} A new object
  */
-export const deepMerge = ( a, b ) => deepMergeWithResolver( a, b, ( _, b ) => b );
+export const deepMerge = ( base, ...rest ) => {
+  if ( !isPlainObject( base ) ) {
+    throw new Error( 'First argument is not an object.' );
+  }
+  return rest.reduce( ( merged, o ) => deepMergeWithResolver( merged, o, ( _, b ) => b ), base );
+};
 
 /**
  * Adds an non-writable, non-configurable and non-enumerable property to an object
--- a/sdk/core/src/helpers/object.js
+++ b/sdk/core/src/helpers/object.js
@@ -44,7 +44,11 @@ export const deepMergeWithResolver = ( base, ...args ) => {
   const resolver = args.at( -1 );
 
   if ( !isPlainObject( base ) ) {
-    throw new Error( 'First argument is not an object.' );
+    throw new Error( 'First argument (base object) is not an object.' );
+  }
+
+  if ( typeof resolver !== 'function' ) {
+    throw new Error( 'Last argument (resolver) is not a function.' );
   }
 
   return objects.reduce( ( merged, object ) => {
--- a/sdk/core/src/helpers/object.js
+++ b/sdk/core/src/helpers/object.js
@@ -24,38 +24,45 @@ export const clone = v => {
 };
 
 /**
- * Creates a new object merging object "b" onto object "a", using a resolver function to define the value to keep.
- * - Object "b" fields that also exists on "a" will have their value defined by the "resolver" function
- * - Object "b" fields that don't exist on object "a" will be added;
- * - Object "a" fields that don't exist on object "b" will be preserved;
+ * Creates a new object recursively merging the rightmost object over the previous one using a resolver function.
+ * Give two objects, (L)eft and (R)right
+ * - Object "R" will overwrite fields on object "L" based on the result of the resolver function;
+ * - Object "R" fields that don't exist on object "L" will be added;
+ * - Object "L" fields that don't exist on object "R" will be preserved;
  *
- * If "b" isn't an object, a new object equal to "a" is returned
+ * If "R" isn't an object, a new object equal to "L" is returned.
  *
- * @param {object} a - The base object
- * @param {object} b - The target object
- * @param {function} resolver - A function that return the value to be kept. First argument is value a, second is value b
+ * The resolver function will define the final value of the merge, it receives two args value "L" and "R".
+ *
+ *
+ * @param {object} base - The base object
+ * @param {...(object|function)} args - Target objects followed by the resolver function
  * @returns {object} A new object
  */
-export const deepMergeWithResolver = ( a, b, resolver ) => {
-  if ( !isPlainObject( a ) ) {
-    throw new Error( 'Parameter "a" is not an object.' );
+export const deepMergeWithResolver = ( base, ...args ) => {
+  const objects = args.slice( 0, -1 );
+  const resolver = args.at( -1 );
+
+  if ( !isPlainObject( base ) ) {
+    throw new Error( 'First argument is not an object.' );
   }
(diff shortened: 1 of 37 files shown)


sdk/core/src/worker/webpack_loaders/workflow_rewriter/index.mjs
```diff
@@ -33,15 +33,14 @@ export default function stepImportRewriterAstLoader( source, inputMap ) {
     const filename = this.resourcePath;
     const ast = parse( String( source ), filename );
     const fileDir = dirname( filename );
-    const { stepImports, sharedStepImports, evaluatorImports, sharedEvaluatorImports } =
-      collectTargetImports( ast, fileDir, cache, filename );
+    const { activityImports } = collectTargetImports( ast, fileDir, cache, filename );
```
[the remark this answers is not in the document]
szanata: Manually fixed.

sdk/core/src/interface/workflow.js
```diff
@@ -30,78 +24,73 @@ export function workflow( { name, description, inputSchema, outputSchema, fn, op
...
+        memo.activityOptions, // parent options
+        invocationOptions.activityOptions // invocation options
+      );
 
-      if ( isChild ) {
-        const result = await executeChild( name, {
+      // If the parent workflow already installed the activity dispatcher,
+      // this means that other calls to workflow() are suppose to start child workflows
+      const isChildWorkflowCall = !!globalThis[C.INVOKE_ACTIVITY_SYMBOL];
+      if ( isChildWorkflowCall ) {
```
catsby: nit: considering the recent [bug disclosure about leaking contexts](https://github.com/temporalio/sdk-typescript/issues/2170) should we do an additional check here?

```js
const dispatcherTag = globalThis[C.INVOKE_ACTIVITY_SYMBOL];
const isChildWorkflowCall = !!dispatcherTag;
if ( isChildWorkflowCall && dispatcherTag.workflowId !== workflowId ) {
  throw new Error(
    `Activity dispatcher contamination detected: found a dispatcher for workflow "${dispatcherTag.workflowId}" ` +
    `while executing "${workflowId}". This indicates globalThis was not reset between workflow executions.`
  );
}
```

It probably shouldn't be necessary but I think that bug report demonstrated that Temporal's isolation guarantee broke silently, so maybe we should be definsive here? 

Just a thought, not blocking

sdk/core/src/helpers/object.js
```diff
@@ -59,23 +59,29 @@ export const deepMergeWithResolver = ( a, b, resolver ) => {
...
  *
- * If "b" isn't an object, a new object equal to "a" is returned
+ * If "R" isn't an object, a new object equal to "L" is returned
  *
  * @param {object} a - The base object
  * @param {object} b - The target object
  * @returns {object} A new object
  */
-export const deepMerge = ( a, b ) => deepMergeWithResolver( a, b, ( _, b ) => b );
+export const deepMerge = ( base, ...rest ) => {
```
catsby: with the signature change here, we should probably add some tests to `sdk/core/src/helpers/object.spec.js` to make sure the new signature is widely supported and we don't break it later

catsby (approved):
Minor nit on checking the dispatcher, and I think we should add a test or two after the function signature change

sdk/core/src/worker/webpack_loaders/workflow_rewriter/rewrite_fn_bodies.js
```diff
@@ -1,193 +1,88 @@
...
+      path.replaceWith( createInvokeActivityCall( activityName, path.node.arguments ) );
+      state.rewrote = true;
     }
   } );
+
   return state.rewrote;
 };
+
+// Entry point: resolve collected component imports into plain dispatcher calls.
+export default function rewriteFnBodies( { ast, activityImports } ) {
```
bnchrch: 💅 Name's a bit off now: it rewrites activity calls in any function, not just `fn` bodies. `rewriteActivityCalls` inside is fine, it's the file and export still saying `fn_bodies`. Rename whenever.

bnchrch (commented):
❗ Like the direction, the new dispatcher is much easier to follow. One blocker below, rest is minor. Two of these land on files this PR doesn't touch, so they're here in the body rather than inline.

---

❗ **Blocker: breaks the default eval workflow** — `sdk/evals/src/eval_workflow.ts`

`defaultFn` still grabs the dispatcher off `this`:

```ts
const { invokeEvaluator } = this as { ... };
```

but the new `workflow.js` calls `fn(input, ctx)` with no `this`, so it throws `Cannot destructure ... of undefined` on first run. `sdk/evals` isn't touched here, and the rewriter can't help since the evaluator is picked by name at runtime.

`simple_eval`, `simple_alias` and `blog_generator` all hit it. CI is green only because e2e runs `simple` and never an eval.

Fix (symbol is global, no core export needed):

```ts
const invoke = globalThis[Symbol.for('@outputai/core:__invoke_activity')];
const result = await invoke(evalDef.name, {...}) as EvaluationResult;
```

❓ Follow-up, or just missed? Either way I think it lands here, plus a `test_eval` step in e2e.

---

💅 **Stale `$shared#` string** — `sdk/cli/src/services/workflow_history/correlator.ts`

Now that `$shared#` is gone (keys are `workflowName#name`), this never matches and `callWebhook` stops getting filtered from the history view.

```ts
const NOISE_STEP_EXACT = [ '$shared#callWebhook' ];
```

Match the bare `callWebhook`, or drop it. Might be worth grepping for other `$shared#` strings too.

sdk/core/src/worker/webpack_loaders/workflow_rewriter/rewrite_fn_bodies.js
szanata: You are right

sdk/core/src/interface/workflow.js
szanata: I have added that! 👍

PR #313, opened 2026-07-09
refactor(llm): Handling Anthropic grammar error as retryable (OUT-518)

szanata:
## Summary

Refactored Anthropic's error `"Grammar compilation timed out."` handling not to throw `FatalError` as this is transient and `FatalError`s terminate the workflow execution without retries.

It seems that the Anthropic API throws this error (HTTP status code 400) when grammar compilation times out for a structured output schema, but after some investigation it was assessed that this error is indeed transient.

### DIFF
--- a/sdk/llm/src/utils/error_handler.js
+++ b/sdk/llm/src/utils/error_handler.js
@@ -77,7 +77,16 @@ export const mapAiError = error => {
     return error;
   }
 
-  if ( APICallError.isInstance( error ) && !error.isRetryable ) {
+  const isApiError = APICallError.isInstance( error );
+  const isGrammarCompilationError = error.message === 'Grammar compilation timed out.';
+  const isAnthropic = error.url?.includes( 'anthropic.com' );
+
+  // This error is actually transient, so instead of FatalError, return it
+  if ( isApiError && error.statusCode === 400 && isGrammarCompilationError && isAnthropic ) {
+    return error;
+  }
+
+  if ( isApiError && !error.isRetryable ) {
     // Non-retryable API failures are already classified by AI SDK as permanent provider failures.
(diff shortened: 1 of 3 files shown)


derencius (approved):

sdk/llm/src/utils/error_handler.js
```diff
@@ -77,7 +77,16 @@ export const mapAiError = error => {
     return error;
   }
 
-  if ( APICallError.isInstance( error ) && !error.isRetryable ) {
+  const isApiError = APICallError.isInstance( error );
+  const isGrammarCompilationError = error.message === 'Grammar compilation timed out.';
```
catsby, line 81: nit: in `mapAiError` above we use `error.message.includes()` but here we use strict equality, should we use `includes()` instead?

sdk/llm/src/utils/error_handler.js
```diff
@@ -77,7 +77,16 @@ export const mapAiError = error => {
...
+  const isAnthropic = error.url?.includes( 'anthropic.com' );
```
catsby: Will this `anthropic.com` check catch Claude on Vertex or Bedrock?

szanata: I have loosened the check and remove the url verification.

sdk/llm/src/utils/error_handler.js
szanata: I think for this error, since it is very short, it is ok to use `===`

sdk/llm/src/utils/error_handler.js
```diff
@@ -77,7 +77,16 @@
...
```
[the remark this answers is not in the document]
szanata: The code changed...

catsby (approved):

PR #315, opened 2026-07-10
chore: Refactor HTTP logging, remove Morgan library

catsby:
## Summary

The `morgan` library we were using wasn't giving us much, and we were serializing to JSON from it just to parse it in `winston`, so it was a bit wasteful. The only thing we were really getting was response times, so I replaced that part with our own.

Also, we were logging `HTTP Request` for every HTTP log, with the actual data we'd want in the metadata. This changes that to put more meaningful things in the actual message, as well as moving `status` (a reserved attribute in Datadog) to `http.status_code` as recommended

### DIFF
--- a/api/src/middleware/http_logger.js
+++ b/api/src/middleware/http_logger.js
@@ -1,4 +1,3 @@
-import { performance } from 'node:perf_hooks';
 import { logger } from '#logger';
 import { isProduction } from '#configs';
 
@@ -54,13 +53,13 @@ const getPayloadSizeFields = ( status, error, contentLength ) => {
  */
 const gatherLogInfo = ( req, res, { method, url, responseTime } ) => {
   const status = res.statusCode;
-  const error = res.locals?.error ?? null;
+  const error = res.locals?.error;
   const requestContentLength = req.headers?.['content-length'];
 
   return {
     method,
     url,
-    'http.status_code': status,
+    statusCode: status,
     contentLength: String( res.getHeader( 'content-length' ) ?? '0' ),
     responseTime,
     requestId: req.id,
@@ -69,7 +68,7 @@ const gatherLogInfo = ( req, res, { method, url, responseTime } ) => {
       errorMessage: error.message
     } : {} ),
     ...getPayloadSizeFields( status, error, requestContentLength ),
-    workflowName: req.body?.workflowName || undefined
+    workflowName: req.body?.workflowName
   };
 };
 
@@ -79,14 +78,15 @@ const gatherLogInfo = ( req, res, { method, url, responseTime } ) => {
  * @param {Record<string, unknown>} info
  */
 const emit = info => {
-  const { method, url, 'http.status_code': status, contentLength, responseTime, errorType, errorMessage } = info;
-  const level = levelForStatus( status );
-  const summary = `${method} ${url} ${status} ${responseTime}ms`;
+  const { statusCode, ...rest } = info;
+  const { method, url, contentLength, responseTime, errorType, errorMessage } = rest;
+  const level = levelForStatus( statusCode );
 
   if ( isProduction ) {
-    logger[level]( summary, info );
+    logger[level]( `${method} ${url} ${statusCode} ${responseTime}ms`, { ...rest, 'http.status_code': statusCode } );
   } else {
-    logger[level]( `${method} ${url} ${status} ${contentLength}b ${responseTime}ms ${errorType ? `[${errorType}: ${errorMessage}]` : ''}`.trimEnd() );
+    const errorSuffix = errorType ? ` [${errorType}: ${errorMessage}]` : '';
+    logger[level]( `${method} ${url} ${statusCode} ${contentLength}b ${responseTime}ms${errorSuffix}` );
   }
 };
 
@@ -103,9 +103,9 @@ export const createHttpLoggingMiddleware = () => ( req, res, next ) => {
     return;
   }
 
-  const start = performance.now();
+  const start = Date.now();
   const method = req.method;
-  const url = req.originalUrl || req.url;
+  const url = req.originalUrl ?? req.url;
   const state = { logged: false };
 
   const log = () => {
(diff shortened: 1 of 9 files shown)


api/src/middleware/http_logger.js
```diff
@@ -31,85 +46,79 @@ const getPayloadSizeFields = ( status, error, contentLength ) => {
...
 
   return {
-    method: tokens.method( req, res ),
-    url: tokens.url( req, res ),
-    status,
-    contentLength: tokens.res( req, res, 'content-length' ) || '0',
-    responseTime: Number.parseFloat( tokens['response-time']( req, res ) ) || 0,
+    method,
+    url,
+    'http.status_code': status,
```
szanata: I would avoid putting a dot in this property, it makes much harder to "serde".
You mentioned that `status` is a reserved world in Datadog. So I am down to use `statusCode` which is the standard in js network.

api/src/middleware/http_logger.js
```diff
@@ -31,85 +46,79 @@ const getPayloadSizeFields = ( status, error, contentLength ) => {
...
-    if ( parseError instanceof SyntaxError ) {
-      logger.warn( 'Failed to parse HTTP log JSON', {
-        raw: message.trim(),
-        parseError: parseError.message
-      } );
-    } else {
-      throw parseError;
-    }
+const emit = info => {
+  const { method, url, 'http.status_code': status, contentLength, responseTime, errorType, errorMessage } = info;
```
szanata: You even deconstruct as `status` here, so changing to `statusCode` is the clear direction.

api/src/middleware/http_logger.js
```diff
@@ -31,85 +46,79 @@ const getPayloadSizeFields = ( status, error, contentLength ) => {
...
+  const level = levelForStatus( status );
+  const summary = `${method} ${url} ${status} ${responseTime}ms`;
+
+  if ( isProduction ) {
+    logger[level]( summary, info );
```
szanata: var `summary` is used only on this line, does it need to be a var. Also, summary and info overlap a lot of info, is that intentional?

api/src/middleware/http_logger.js
```diff
@@ -31,85 +46,79 @@ const getPayloadSizeFields = ( status, error, contentLength ) => {
...
-/**
- * Passes the trimmed message to the HTTP logger.
- * @param {string} message - Human-readable log line
- */
-const devStreamWrite = message => {
-  logger.http( message.trim() );
-};
+  const start = performance.now();
+  const method = req.method;
+  const url = req.originalUrl || req.url;
```
szanata: what is originalUrl vs url?

api/src/middleware/http_logger.js
```diff
@@ -31,85 +46,79 @@ const getPayloadSizeFields = ( status, error, contentLength ) => {
...
+  }
 
-/**
- * Passes the trimmed message to the HTTP logger.
- * @param {string} message - Human-readable log line
- */
-const devStreamWrite = message => {
-  logger.http( message.trim() );
-};
+  const start = performance.now();
```
szanata: You dont need performance module for this, this is overkill.

A simple

```js
const startAt = Date.now();
...
Date.now() - startAt
```

is enought. Also, since this would be a Number in milliseconds already, you dont need `.toFixed()`.

api/src/middleware/http_logger.js
```diff
@@ -31,85 +46,79 @@ const getPayloadSizeFields = ( status, error, contentLength ) => {
...
-      errorType: res.locals.error.constructor.name,
-      errorMessage: res.locals.error.message
+    ...( error ? {
+      errorType: error.constructor.name,
+      errorMessage: error.message
     } : {} ),
-    ...getPayloadSizeFields( status, res.locals.error, requestContentLength ),
-    workflowName: req.body?.workflowName ? req.body.workflowName : undefined
+    ...getPayloadSizeFields( status, error, requestContentLength ),
+    workflowName: req.body?.workflowName || undefined
```
szanata: This should be `??`, because if the workflow name would be `''` or `0`, this would assign undefined.

api/src/middleware/http_logger.js
szanata: Also, `??` is preferable for this.

api/src/middleware/http_logger.js
```diff
@@ -31,85 +46,79 @@ const getPayloadSizeFields = ( status, error, contentLength ) => {
...
  * @param {import('express').Response} res
+ * @param {{ method: string, url: string, responseTime: number }} timing
  * @returns {Record<string, unknown>}
  */
-const gatherLogInfo = ( tokens, req, res ) => {
-  const status = Number.parseInt( tokens.status( req, res ), 10 ) || 0;
-  const requestContentLength = req?.headers?.['content-length'];
+const gatherLogInfo = ( req, res, { method, url, responseTime } ) => {
+  const status = res.statusCode;
+  const error = res.locals?.error ?? null;
```
szanata: `?? null` isnt necessary. The only scenario where it would assign `null` is if error is `null` or `undefined`.

szanata (approved):
I am always down for refactors + lib removal. I left some comments and tweaks proposals.

I will pre-approve, but I think you could take a look in those suggestions.

api/src/middleware/http_logger.js
catsby: The `http.status_code` is specifically called out by Datadog as the span attribute name, so I'm going to leave it. I'll change the var name to `statusCode` though

api/src/middleware/http_logger.js
catsby: The overlap was intentional, `summary` is the human readable bit to show on the line and easily read in Datadog (or other aggregates), `info` is the more structured metadata. Agreed though that `summary` as a var isn't useful so I'll inline it

api/src/middleware/http_logger.js
catsby: `originalUrl` is the full path as the client sent it, `req.url` is a modified version after Express middleware rewrite things (if they do). I'll update to use `??` though

api/src/middleware/http_logger.js
catsby: Good call, however now that I look closer if `req.body` or `req.body.workflowName` is `null` or `undefined`, then the end result is `undefined` anyway so `??` isn't needed at all, so I'll remove that

PR #318, opened 2026-07-13
feat(core): expose step-side Temporal access via @outputai/core/temporal (OUT-519)

pedrofs:
## Why

Steps that stream incremental results back to their invoking workflow — e.g. an LLM step feeding token batches to a workflow whose update handler serves a long-polling client — have no supported path to the Temporal layer. Porting the Ask AI chat workflow from Flow (growthxai/clients-workflows#765) required hand-rolling a `Client` from the worker's own `TEMPORAL_*` env vars, duplicating connection config the worker already parses plus the `temporal-namespace` metadata-header gotcha for Temporal Cloud API-key auth. The SDK already builds exactly this client internally (`CatalogJob`); this PR exposes it to step authors safely.

## What

New **public** subpath `@outputai/core/temporal` (node-runtime only — documented and typed as never importable from the workflow sandbox):

- **`getWorkflowClient()`** — memoized `Client` sharing the worker's live `NativeConnection` + namespace. Registered at worker startup, cleared on shutdown; inherits TLS/API key/proxy with zero extra config.
- **`signalInvokingWorkflow(name, ...args)`** — scoped helper that resolves the invoking run from the existing activity AsyncLocalStorage context and pins `workflowId` + `runId`, so a stale activity from a retried/superseded run can never signal a newer run reusing the same workflow id (strictly safer than the workflowId-only signaling projects hand-roll today).

Both throw a non-retryable `FatalError` outside a running worker. Docs added to `docs/guides/packages/core.mdx` (including guidance to prefer plain return-value dataflow and batch aggressively; out-of-band channels for high-frequency streams), typedoc entry point registered.

## Testing

- 8 new unit tests (`sdk/core/src/temporal/workflow_client.spec.js`); full `sdk/core` suite green (716 tests). The `sdk/cli` suite fails identically on a clean `main` checkout in my environment — unrelated.
- Design proven downstream: the clients-workflows chat port runs this exact pattern (hand-rolled client) live against a local worker end-to-end.

## Notes

- No issue was open for this; happy to file one if required — this came out of the Flow→Output chat migration, where the capability gap was hit in practice.
- Deliberately additive: no changes to existing exports or the workflow sandbox surface.

---
## Amendment

by @szanata 

Reworked the public surface of `@outputai/core/temporal` to match the worker architecture:

- `createTemporalClient()`: returns a new Temporal Client on each call, sharing the worker's live connection + namespace (TLS/API key/proxy inherited). Do not close client.connection; the worker owns it and `close()` throws a non-retryable `FatalError`.
- `getCurrentWorkflowHandle()`: activity-only helper that returns a handle pinned to the invoking run's `workflowId` + `runId` (reuses one memoized Client). Replaces the earlier `getWorkflowClient()`/`signalInvokingWorkflow()` pair.

Also: docs updated for the new API, unit coverage for both helpers, and a `temporal_client` test workflow that exercises signaling via the handle and via a fresh client. Patch changeset added for `@outputai/core`.


szanata:
I have reworked the PR to better match the architecture and added a test workflow.

It will expose two methods over `@outputai/core/temporal`:
- `createTemporalClient()`: Creates a new client when called;
- `getCurrentWorkflowHandle()`: Return a Temporal handle for the current workflow (must be called within activities). It will create a client once and reuse it.

Another change is that when exposing the Temporal client, if a user tries to close its connection (`client.connection.close()`) it will fail, as the connection is reused from the worker and managed by it.

PR #321, opened 2026-07-15
feat(core): Adding custom events to hooks + refactor (OUT-520)

szanata:
## Summary

- Added `emit()` to `@outputai/core/hooks` endpoint, so users can emit custom events to be listened to with `on()`. It is designed to be used within steps and will not work inside a workflow since the `/hooks` endpoint is not sandbox safe to be imported, if used elsewhere, context fields will be omitted;
- Refactored event emission to have two buses, one for worker, workflow and activity lifecycle and one for in-step events, like those from other SDK modules and user events. Simplified event emission and handling;
- Refactored events shape, all events listened using `on()` now have an envelope wrapper that adds the step/workflow info without risks of collision and are independent of the payload type:
  ```js
  emit( 'foo', 'bar' );
  
  on( 'foo', event => {
    console.log( event );
    assert( event, {
      eventId: "string", // uuid
      eventDate: number, // epoch
      activityInfo: {}, // result of Temporal's activityInfo() fn
      outputActivityKind: "string", // step, evaluator, internal_step, etc
      workflowDetails: {}, // serialization of some fields from Temporal's workflowInfo() fn
      payload: [original emitted event payload] // in this case "bar" 
    } );
  } );
  ```
   This includes events from SDK, like: `http:request`, `cost:llm:request`, `cost:http:request`;
- Added internal activity events to activity lifecycle hooks: `onActivityStart`, `onActivityEnd`, `onActivityError`;
- Removed `$catalog` workflow errors from `onError()`;
- 

## Test plan

- [x] Manually tested
- [x] Refactored/added unit tests

### DIFF
--- a/docs/guides/migrations/v0.10.0-to-v0.11.0.mdx
+++ b/docs/guides/migrations/v0.10.0-to-v0.11.0.mdx
@@ -1,9 +1,9 @@
 ---
 title: "v0.10.0 → v0.11.0"
-description: "Upgrading Output.ai projects from v0.10.0 to v0.11.0: API request logging, Temporal rollout safety, shared activity names, and child workflow activity-option precedence."
+description: "Upgrading Output.ai projects from v0.10.0 to v0.11.0: hook event envelopes, API request logging, Temporal rollout safety, shared activity names, and child workflow activity-option precedence."
 ---
 
-This guide covers customer-facing changes to API request logging in `output-api` and workflow execution in `@outputai/core`.
+This guide covers customer-facing changes to hooks and workflow execution in `@outputai/core`, plus API request logging in `output-api`.
 
 ## API HTTP request logging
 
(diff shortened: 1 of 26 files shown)


test_workflows/src/hooks.js
```diff
@@ -16,17 +17,23 @@
...
+// on( 'test', payload => console.log( colorize( 'on(test)' ), payload ) );
 
 // // Generic on error
 // onError( payload => console.log( colorize( 'onError()' ), payload ) );
 
 // // Worker start
-// onBeforeWorkerStart( () => console.log( colorize( 'onBeforeWorkerStart()' ) ) );
+// onBeforeWorkerStart( () => {
+//   console.log( colorize( 'onBeforeWorkerStart()' ) )
+//   emit( 'test', { message: 'foo' } );
```
catsby, line 28: Is this a valid `emit()` call? I didn't think there was step context before workers start and the PR description says "designed to be used within steps".

sdk/core/src/hooks/index.d.ts
```diff
@@ -116,7 +116,7 @@ export interface HookPayloadBase {
 export interface ActivityPayloadBase extends HookPayloadBase {
   /** Temporal's activityInfo(). */
   activityInfo: Info;
-  /** Output component kind for the activity, e.g. step or evaluator. */
+  /** Output component kind for the activity, e.g. step, evaluator, or internal_step. */
   outputActivityKind: string;
```
catsby: is this field optional or required? It's shown as `outputActivityKind?:` in some places

docs/guides/migrations/v0.10.0-to-v0.11.0.mdx
```diff
@@ -142,6 +142,127 @@ If a dashboard or aggregation relied on fractional milliseconds in `responseTime
...
+
+The envelope has this shape:
+
+```ts
+{
+  eventId: string;
+  eventDate: number;
+  activityInfo: Info;
+  workflowDetails: WorkflowDetails;
+  outputActivityKind?: string;
```
catsby: `outputActivityKind` shown as optional here

sdk/core/src/sdk/runtime/events.d.ts
```diff
@@ -2,14 +2,12 @@
...
-   *
-   * When called inside an Output activity context, the framework automatically
-   * attaches `activityInfo`, `workflowDetails`, and `outputActivityKind` onto the emitted payload.
+   * Emits an event on step message bus.
    *
    * @param eventName - The name of the event to emit
-   * @param payload - An optional payload to send to the event
+   * @param payload - An optional payload to send to the event.
    */
   emit( eventName: string, payload?: object ): void;
```
catsby: The wrapped `EventEmitter`'s `emit()` function returns `boolean`, should we return that here instead of `void`?

catsby (approved):
pre-approving with minor feedback

test_workflows/src/hooks.js
szanata: You can use outside steps, but you dont receive the context. Nothing blocking thou.

sdk/core/src/hooks/index.d.ts
szanata: Not optional. I will fix it.

sdk/core/src/sdk/runtime/events.d.ts
szanata: Yeah, that would be preferable.

sdk/core/src/hooks/index.d.ts
szanata: Actually, changed my mind and made all context field optional.

test_workflows/src/hooks.js
szanata: I have updated the docs/types to make that behavior explicit

PR #322, opened 2026-07-17
refactor(http): outputFetch/createKyClient over fetch/httpClient (OUT-402)

szanata:
## HTTP client refactor

Replacing `fetch()` and `httpClient()` with `outputFetch()` and `createKyClient()` at `@outputai/http`.

### `outputFetch()`

The `fetch()` function was renamed to have a non conflicting name. The main difference is how it handles realms (Node vs Undici), before the refactor it was using `undici.install()`, globally overwriting Node's fetch types with Undici's. Now it supports both `globalThis.Request/Headers` and `unidici.Request/Headers`, it converts the shapes internally without adding the global side-effect.

### `createKyClient()`

The `httpClient()` was renamed to a more fitting name for a function and gets its Ky version upgraded from 1 to 2. The behavior is the same as before.

The ky upgrade affects all modules in the platform, they are now using v2.

### Re-exports

All specific re-exports from both `ky` and `undici` where dropped (`HTTPError`, `TimeoutError`, `HttpClientOptions`, `RequestInfo`, `RequestInit`). `/http` already exposed `undici` as a named export, now it does the same for `ky`, so there isn't cherry pick anymore. Users can access native features easily.

### Global proxy

The core module was setting a global dispatcher for all `fetch` requests made in the platform, so they would use a proxy agent. This isn't necessary as every place performing HTTP requests (`/http`, `/llm`, `/core`, `/cli`) is proxy aware already, that was only adding a unexpected behavior and a global side-effect on user core, so it was dropped.

### Peer dependencies

Both `undici` and `ky` are now peer dependencies accepting a range of versions, `>8 <9` and `>2 <3` respectively.


sdk/http/src/instrumented_fetch/index.ts
```diff
@@ -0,0 +1,97 @@
...
+  const url = request.url;
+  const startedAt = performance.now();
+
+  await logRequest( { requestId, request } );
+
+  // this allows for users not only to override the dispatcher but also to define it as undefined and remove it altogether.
+  const dispatcher = Object.hasOwn( init ?? {}, 'dispatcher' ) ? inputDispatcher : customDispatcher;
+  try {
+    const response = await undici.fetch( request, dispatcher ? { dispatcher } : undefined );
+    const durationMs = performance.now() - startedAt;
```
catsby: do we need `performance`? it was flagged in a prior review as not needed. I don't mind it just verifying

sdk/http/src/fetch/events.js
```diff
@@ -0,0 +1,10 @@
+import { Event } from '@outputai/core/sdk/runtime';
+
+export const emitError = ( { requestId, method, url, status, durationMs } ) =>
+  Event.emit( 'http:request', { requestId, method, url, status, durationMs, outcome: 'error' } );
+
+export const emitSuccess = ( { requestId, method, url, status, durationMs } ) =>
+  Event.emit( 'http:request', { requestId, method, url, status, durationMs, outcome: 'success' } );
+
+export const emitFailure = ( { requestId, method, url, durationMs } ) =>
```
catsby, line 9: previously these kinds of things went through `emitHttpRequestEvent()`, and they used a consistent shape didn't they? 

https://github.com/growthxai/output/blob/d815a8e30d3e934932b55f38a63e78954f1d633b/sdk/http/src/fetch/logger.ts#L9

and the outcome was a type, but now the outcome is a string, is that correct?

szanata: Yes, I split the log vs event behavior in two files. That type was internal only, the external type wasn't used here at all. The new behavior is covered by unit tests, so I don't think it necessary to type internally.

sdk/http/src/instrumented_fetch/logger.js
```diff
@@ -0,0 +1,43 @@
...
+  Tracing.addEventAttribute( { eventId: requestId, attribute: new Tracing.Attribute.HTTPRequestCount( request.url, requestId ) } );
+};
+
+export const logError = async ( { requestId, response } ) =>
+  Tracing.addEventError( {
+    id: requestId, details: {
+      status: response.status,
+      statusText: response.statusText,
+      headers: redactHeaders( response.headers ),
+      body: await parseBody( response )
```
catsby: we unconditionally call `parseBody()` here, but in `logRequest()` we only call it if `config.logVerbose` is set, so we're always logging the full body when `logError()` is called, regardless if the user enabled verbose logging. is this intentional?

sdk/http/tsconfig.json
```diff
@@ -1,23 +1,24 @@
...
+    "moduleResolution": "NodeNext",
     "lib": ["ES2022"],
     "strict": true,
     "outDir": "./dist",
     "rootDir": "./src",
     "declaration": true,
     "esModuleInterop": true,
     "skipLibCheck": true,
+    "allowJs": true,
+    "checkJs": false,
```
catsby, line 14: is this something we should look to turn on in the future? seems type checking is a good thing long-term

sdk/http/src/instrumented_ky/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ * });
+ *
+ * const response = await client.get('users/1');
+ * const data = await response.json();
+ * ```
+ *
+ * @param options - The ky options to extend the base client.
+ * @returns A ky instance extended with Output.ai tracing hooks.
+ */
+export const createKyClient = ( options: Options = {} ) =>
```
catsby: do we need to export `Options` here so users know what those options are?

catsby: also should `createKyClient()` have an explicit return type?

sdk/http/package.json
```diff
@@ -14,16 +14,19 @@
   "scripts": {
     "build": "rm -rf ./dist && tsc"
   },
+  "peerDependencies": {
+    "ky": ">=2 <3",
+    "undici": ">=8 <9"
+  },
```
catsby, line 20: do we need to add something like this to the generated `package.json` file that projects created with `output init` get? 

https://github.com/growthxai/output/blob/69a0f11914cfdf8dac5fd67c791a1952e4e04c69/sdk/cli/src/templates/project/package.json.template

sdk/http/src/instrumented_fetch/utils.js
```diff
@@ -10,16 +9,16 @@ const HEADER_REDACTION_EXEMPT = new Set( [
 ] );
 
 /** Matches red int "hot-red-pie", but not int "redact" */
```
catsby:
```suggestion
/** Matches red in "hot-red-pie", but not in "redact" */
```

coding_assistants/claude/plugins/outputai/skills/output-dev-step-function/SKILL.md
```diff
@@ -194,21 +194,21 @@ fn: async input => {
...
 ```typescript
-import { httpClient } from '@outputai/http';
+import { createKyClient } from '@outputai/http';
 import { FatalError, ValidationError } from '@outputai/core';
 
 const RETRY_STATUS_CODES = [ 408, 429, 500, 502, 503, 504 ];
 const FATAL_STATUS_CODES = [ 401, 403, 404 ];
 
-const httpClientInstance = httpClient( {
+const httpClientInstance = createKyClient( {
```
catsby: nit: `httpClientInstance` lined up with `httpClient()` well, should we rename this to just `client`?

.changeset/pink-wings-judge.md
```diff
@@ -0,0 +1,32 @@
+---
+"@outputai/http": minor
+---
+
+## HTTP clients
+
+- Replaced `fetch` with `instrumentedFetch`, the explicitly named Fetch-compatible API for traced HTTP requests.
```
catsby: nit: `instrumentedFetch` reads very strange to me, what do you think about `outputFetch` or similar? It's clear that it's the fetch provided by the Output framework, and our guides recommend to use that. "Instrumented" is almost an implementation detail users may not normally care about.

sdk/http/src/instrumented_fetch/logger.js
szanata: Yes, it was intentional, but don't make much sense, I don't recall why we did that wait. I will update it.

sdk/http/src/instrumented_ky/index.ts
szanata: > do we need to export Options here so users know what those options are?

No, semantically we dont need to re-export it, considering we already re-export the whole `ky`.

> also should createKyClient() have an explicit return type?

I just added it, thanks.

sdk/http/package.json
szanata: No, it is `http` that has the restriction, the `package.json` of the project just needs to reference `http`. If users bring their own `ky`/`undici`, they will have to satisfy `https` rules.

coding_assistants/claude/plugins/outputai/skills/output-dev-step-function/SKILL.md
szanata: Yeah, I renamed `kyClient`.

.changeset/pink-wings-judge.md
szanata: I renamed to `outputFetch`.

sdk/http/src/instrumented_fetch/logger.js
szanata: Done.

sdk/http/tsconfig.json
szanata: Maybe in the future, for this specific lib, it isnt necessary

coding_assistants/claude/plugins/outputai/skills/output-dev-step-function/SKILL.md
szanata: I actually renamed to just `client` and I reviewed all other docs to keep it `client` everywhere.

catsby (approved):

PR #323, opened 2026-07-20
feat(llm): Added stricter LiquidJs configs for prompt parsing (OUT-493)

szanata:
## Summary

Added `strictFilters: true`, `strictVariables: true`, `lenientIf: true` settings to LiquidJS in order to make prompt content interpolation for deterministic.

[Reference](https://liquidjs.com/tutorials/options.html#Strict)

### DIFF
--- a/.changeset/cozy-peas-stand.md
+++ b/.changeset/cozy-peas-stand.md
@@ -1,5 +1,12 @@
 ---
-"@outputai/llm": patch
+"@outputai/llm": minor
 ---
 
-Added stricter LiquidJs settings for prompt files parsing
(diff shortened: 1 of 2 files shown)


.changeset/cozy-peas-stand.md
```diff
@@ -0,0 +1,5 @@
+---
+"@outputai/llm": patch
```
catsby: nit: should this be a minor bump instead?

catsby (approved):

.changeset/cozy-peas-stand.md
szanata: Done

PR #325, opened 2026-07-21
feat(core): Added hooks flush (OUT-527)

szanata:
## Summary

- Added a mechanism to await for all hooks to complete before worker shutdown, with a max awaiting time of 30s. This replace the current behavior which was a fixed await period of 30s.

## Test plan

- [x] Added tests
- [x] Manually tested

### DIFF
--- a/.changeset/swift-balloons-march.md
+++ b/.changeset/swift-balloons-march.md
@@ -1,0 +1,5 @@
+---
+"@outputai/core": patch
+---
+
+Added a mechanism to await for all hook callbacks to complete before shutting down the worker. Max awqaiting period is 30s.
--- a/.changeset/swift-balloons-march.md
+++ b/.changeset/swift-balloons-march.md
@@ -2,4 +2,4 @@
 "@outputai/core": patch
 ---
 
-Added a mechanism to await for all hook callbacks to complete before shutting down the worker. Max awqaiting period is 30s.
+Added a mechanism to await for all hook callbacks to complete before shutting down the worker. Max awaiting period is 30s.
--- a/sdk/core/src/hooks/index.js
+++ b/sdk/core/src/hooks/index.js
@@ -1,22 +1,24 @@
 import { mainEventBus, stepEventBus } from '#bus';
 import { BusEventType } from '#consts';
 import { createChildLogger } from '#logger';
+import { pendingHooks } from './pending_hooks.js';
 
 const log = createChildLogger( 'Hooks' );
 
 /**
- * Invokes a function within a try catch and log the error
+ * Invokes a function within a try catch and log the error.
  *
  * @param {Function} fn
  * @param {any} args - Args to invoke the function with
  * @param {string} hookName - hookName to identify this hook function in the logs
  */
-const callHookCb = async ( fn, args, hookName ) => {
-  try {
-    await fn( args );
-  } catch ( error ) {
-    log.error( `${hookName} hook error`, { message: error.message, stack: error.stack } );
-  }
+const callHookCb = ( fn, args, hookName ) => {
+  const promise = Promise.resolve()
+    .then( () => fn( args ) )
+    .catch( e => log.error( `${hookName} hook error`, { message: e.message, stack: e.stack } ) )
+    .finally( () => pendingHooks.delete( promise ) );
+  pendingHooks.add( promise );
+  return promise;
 };
 
 // General Life-cycle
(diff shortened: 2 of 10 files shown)


sdk/core/src/worker/index.spec.js
```diff
@@ -262,6 +264,7 @@ describe( 'worker/index', () => {
     expect( catalogJobInstance.run ).toHaveBeenCalled();
 
     await settleWorker();
+    await vi.waitFor( () => expect( flushPendingHooksMock ).toHaveBeenCalledOnce() );
```
catsby, line 266: are we testing correctly here? it seems we mock the pending hooks and it returns immediately. Should we test one that never does return, and verify that we haven't logged 'Bye' yet, or that shutdown has **not** proceeded?

sdk/core/src/hooks/pending_hooks.js
```diff
@@ -0,0 +1,19 @@
...
+ * Await all pending hooks to flush for a certain time
+ * @returns
+ */
+export const flushPendingHooks = async () => {
+  const state = { timeout: null };
+  try {
+    await Promise.race( [
+      Promise.allSettled( [ ...pendingHooks ] ),
+      new Promise( r => state.timeout = setTimeout( r, flushTimeoutMs ) )
+    ] );
```
catsby, line 17: assuming the timeout fires, can/should we log something if some hooks remained? 

```javascript
await Promise.race( [ 
  ...
] );

if ( pendingHooks.size > 0 ) {
  log.warn( 'Flush timed out; abandoning pending hooks', { pending: pendingHooks.size, timeoutMs: flushTimeoutMs } );
}
```

sdk/core/src/worker/index.js
```diff
@@ -214,13 +215,18 @@ execute()
...
+    log.info( 'Flushing events...' );
+    await flushPendingHooks();
+    log.info( 'Bye' );
+  } )
+  .catch( async error => {
     log.error( 'Fatal error', { error: error.message, stack: error.stack } );
 
     mainEventBus.emit( BusEventType.RUNTIME_ERROR, { error } );
 
-    const timeToFlushEvent = configs.processFailureShutdownDelay;
```
catsby, line 221: we have an ENV var that's tied to this `OUTPUT_PROCESS_FAILURE_SHUTDOWN_DELAY` that I don't think is used anymore

sdk/core/src/hooks/index.js
```diff
@@ -1,22 +1,24 @@
 import { mainEventBus, stepEventBus } from '#bus';
 import { BusEventType } from '#consts';
 import { createChildLogger } from '#logger';
+import { pendingHooks } from './pending_hooks.js';
 
 const log = createChildLogger( 'Hooks' );
 
 /**
- * Invokes a function within a try catch and log the error
+ * Invokes a function within a try catch and log the error.
```
catsby, line 9: Is this still accurate? I don't think we're using a `try/catch` anymore

sdk/core/src/hooks/pending_hooks.js
```diff
@@ -0,0 +1,19 @@
+export const pendingHooks = new Set();
+
+const flushTimeoutMs = 30_000;
+
+/**
+ * Await all pending hooks to flush for a certain time
+ * @returns
```
catsby: do we return anything?

.changeset/swift-balloons-march.md
```diff
@@ -0,0 +1,5 @@
+---
+"@outputai/core": patch
+---
+
+Added a mechanism to await for all hook callbacks to complete before shutting down the worker. Max awqaiting period is 30s.
```
catsby:
```suggestion
Added a mechanism to await for all hook callbacks to complete before shutting down the worker. Max awaiting period is 30s.
```

catsby (approved):
Pre-approving pending some cleanups

sdk/core/src/hooks/index.js
szanata: we are using the equivalent of try catch in promises, .try/.catch, so it still is.

sdk/core/src/worker/index.spec.js
szanata: It is correct in my opinion. For the worker, its only concern is to call `flushPendingHooks()` and assume it will return after couple seconds.

Its behavior is to always return, if it doesn't return that would be a bug, so I don't see a point in testing that.

The implementation details that make it always return are covered in `flushPendingHooks()` own test file.

sdk/core/src/hooks/pending_hooks.js
szanata: Yeah, good ideia. I just added it.

sdk/core/src/worker/index.js
szanata: Removed.

PR #327, opened 2026-07-22
chore: rename 'http.status_code' log attribute to just 'statusCode'

catsby:
## Summary

As discussed, renames the recently titled 'http.status_code' log attribute to just 'statusCode' (from https://github.com/growthxai/output/pull/315). Originally named such to line up with Datadog, this PR makes it more generic


szanata (approved):
Good stuff

PR #329, opened 2026-07-23
chore: fix `output dev` TUI crash with unbounded cache

catsby:
## Summary

The TUI from `output dev` cached workflow run information but never checked or flushed it, so it grew unbounded. Eventually that causes a crash because it consumes too much memory and hits limits from the underlying Node process.

In this change we introduce a Least Recently Used (LRU) cache instead, so we eventually purge old things.

### DIFF
--- a/sdk/cli/src/views/dev/hooks/use_run_detail.ts
+++ b/sdk/cli/src/views/dev/hooks/use_run_detail.ts
@@ -10,6 +10,7 @@ import {
 import type { TraceData, DebugNode } from '#types/trace.js';
 import { normalizeWorkflowStatus } from '#utils/normalize_workflow_status.js';
 import { TERMINAL_STATUSES } from '#utils/format_workflow_result.js';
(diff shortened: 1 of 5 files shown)


szanata (changes requested):
Clever implementation 👍 

_Sorry, I clicked request changes by accident._

szanata (approved):

sdk/cli/src/views/dev/hooks/use_run_detail.ts
```diff
@@ -35,7 +36,8 @@ const EMPTY_DETAIL: RunDetail = {
   steps: [],
   loading: false
 };
-const runDetailCache = new Map<string, RunDetail>();
+const RUN_DETAIL_CACHE_MAX = 50;
```
szanata, line 39: We could even go lower than that if necessary, but if you tested with 50 and are comfortable, I am ok.

PR #330, opened 2026-07-24
fix: preserve terminal scroll if TUI crashes

catsby:
## Summary

Fix issue in development when if the TUI should crash terminal scroll is lost and users likely cannot see the actual root issue, just the most recent lines of the crash.

### DIFF
--- a/sdk/cli/src/commands/dev/index.ts
+++ b/sdk/cli/src/commands/dev/index.ts
@@ -187,9 +187,7 @@ export default class Dev extends Command {
     // mid-teardown and let a Ctrl+C orphan the stack. Print the raw error so
     // Node's stack trace survives, then re-exit non-zero. Fire-once: a second
     // catchable crash during teardown is a no-op. A hard V8 abort() (SIGABRT)
-    // is uncatchable and not covered here. The object wrapper holds the
-    // fire-once flag because the linter forbids `let` (same idiom as
-    // exitAltScreenOnce above).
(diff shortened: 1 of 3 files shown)


sdk/cli/src/commands/dev/index.ts
```diff
@@ -151,11 +173,40 @@ export default class Dev extends Command {
...
+    // clean-exit path and the signal handlers. Tear docker down via cleanup()
+    // FIRST, then restore the terminal and print the crash: unmounting Ink
+    // resolves the awaited waitUntilExit(), which resumes run() and strips the
+    // signal listeners — so unmounting before cleanup would drop them
+    // mid-teardown and let a Ctrl+C orphan the stack. Print the raw error so
+    // Node's stack trace survives, then re-exit non-zero. Fire-once: a second
+    // catchable crash during teardown is a no-op. A hard V8 abort() (SIGABRT)
+    // is uncatchable and not covered here. The object wrapper holds the
+    // fire-once flag because the linter forbids `let` (same idiom as
+    // exitAltScreenOnce above).
```
szanata: I think this part of the description about the state object because of lint isnt pertinent.

szanata (approved):

sdk/cli/src/commands/dev/index.ts
catsby: You're right - removed in 891b555

PR #331, opened 2026-07-24
feat: enable TUI to reattach to existing dev sessions

catsby:
## Summary

Enable the CLI TUI to re-attach to an existing `output dev` session. When attaching to a `dev` session that was started in a detached state (`output dev -d`), the CLI will attach to the existing session and start the TUI. Exiting the TUI will then **detach**, and not shut down the `dev` setup.

To teardown a detached session, use the new `output dev down` command.

### Would you like to know more? 

`output dev` now classifies the project's stack with `docker compose ps` rather than probing host ports unconditionally:

| Stack state | Behavior | Owns teardown? |
|---|---|---|
| Nothing live (incl. all containers exited) | Probe ports, start in foreground | Yes — quitting tears down |
| Some live, not all healthy | `up -d` to converge, then monitor | No — quitting leaves it up |
| All live and healthy | Attach and monitor | No — quitting leaves it up |

When the session does not own the stack, startup prints `Quitting leaves these services running — stop them with 'output dev down'.` and the footer reads `ctrl+c detach (keeps running)`.

Teardown for a stack this session doesn't own is explicit: **`output dev down`**.

### Other details

- `output dev -d` now does pre-flight port probe - we can detect if a non-container process is holding a port we expect
- `-d` is now an "inspected async spawn" rather than _fire-and-forget_. A failed _bind_ now gives an actionable hint instead of a raw compose error
- failed `docker compose ps` now warns instead of silently assuming "no containers exist"

## Test plan

- [ ] `npx output dev -d`, then `npx output dev` from the same project — TUI attaches
- [ ] Footer reads `detach (keeps running)`; startup prints the "Quitting leaves…" line
- [ ] Exit with `ctrl+c` — services still up (`http://localhost:8080`, run a workflow)
- [ ] `npx output dev down` — services stop
- [ ] `npx output dev` with nothing running — foreground start, quitting tears down
- [ ] `docker compose stop` the stack, then `npx output dev` — treated as a fresh start (foreground, tears down on quit), **not** an attach
- [ ] Hold port 3001 with a non-Docker process, then `npx output dev -d` — aborts with a hint naming `OUTPUT_API_HOST_PORT` instead of starting an unreachable container

### DIFF
--- a/sdk/cli/src/commands/dev/index.ts
+++ b/sdk/cli/src/commands/dev/index.ts
@@ -109,7 +109,7 @@ export default class Dev extends Command {
       // Reconcile: converge a stack that's up but incomplete to the compose
       // spec, then monitor. Inspected (not fire-and-forget) so a failed bind
       // surfaces the same actionable port-collision hint the fresh-start path
-      // gives. Recovers the OUT-477 orphaned-stack case.
+      // gives. Recovers the orphaned stack a crashed session leaves behind.
       this.log( '🔄 Existing services detected — reconciling before attaching...\n' );
       await this.upDetachedOrError( dockerComposePath, pullPolicy );
     } else {
@@ -243,8 +243,8 @@ export default class Dev extends Command {
       exitAltScreenOnce();
       // An abnormal exit (health timeout, docker crash) resolves outside the
       // signal path. Tear down here only if cleanup hasn't already run, so an
-      // owned stack never leaks containers holding ports — the OUT-477 root
-      // cause. teardownIfOwned no-ops for attach/reconcile sessions.
+      // owned stack never leaks containers that keep holding published ports.
+      // teardownIfOwned no-ops for attach/reconcile sessions.
       if ( !state.cleaningUp ) {
         await teardownIfOwned().catch( teardownError => {
           // Don't mask the original failure, but the stack may still be up
@@ -271,8 +271,8 @@ export default class Dev extends Command {
   }
 
   // Probe only when no container of ours is live — otherwise our own stack is
-  // legitimately holding the ports and a probe would abort on it, which is the
-  // OUT-477 failure this detection replaced.
+  // legitimately holding the ports and the probe would abort on it, which is
+  // the failure this detection replaced.
   private async probePortsIfFreshStart( dockerComposePath: string ): Promise<void> {
     const services = await this.getServiceStatusOrWarn( dockerComposePath );
     if ( classifyStackState( services ) === STACK_STATE.NONE ) {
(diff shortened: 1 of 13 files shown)


sdk/cli/src/commands/dev/index.ts
```diff
@@ -161,42 +203,121 @@ export default class Dev extends Command {
...
+  private async probePorts(): Promise<void> {
+    const takenPorts = await findUnavailablePorts( Object.values( config.ports ) );
+    if ( takenPorts.length > 0 ) {
+      this.error( formatPortCollisionsHint( takenPorts, config.ports ), { exit: 1 } );
+    }
+  }
+
+  // Probe only when no container of ours is live — otherwise our own stack is
+  // legitimately holding the ports and a probe would abort on it, which is the
+  // OUT-477 failure this detection replaced.
```
szanata: We should not add tickets numbers to code, based on our discussion about this a couple weeks ago.

sdk/cli/src/commands/dev/index.ts
```diff
@@ -161,42 +203,121 @@ export default class Dev extends Command {
...
+    if ( classifyStackState( services ) === STACK_STATE.NONE ) {
+      await this.probePorts();
+    }
+  }
+
+  // Query the stack, falling back to "nothing running" when the query itself
+  // fails. A fresh start is the right recovery, but [] is not a neutral default
+  // — it asserts "no containers exist", which every caller acts on. Announce it,
+  // or a broken `ps` silently port-probes onto the user's own containers and
+  // reproduces the exact OUT-477 message this detection replaced, with a remedy
```
szanata: Same here

sdk/cli/src/commands/dev/index.ts
```diff
@@ -161,42 +203,121 @@ export default class Dev extends Command {
...
+  // (change the port) that is wrong for the actual cause.
+  private async getServiceStatusOrWarn( dockerComposePath: string ): Promise<ServiceStatus[]> {
```
szanata: This isnt a strong suggestion, is just a tip, if you have a function with just a `return` statement, you can convert to an arrow function. But this is personal taste and it is not enforced by our lint rules.

szanata (approved):

sdk/cli/src/commands/dev/index.ts
catsby: 🤦‍♂️ reworded to remove issue references in 5b909cf

sdk/cli/src/commands/dev/index.ts
catsby: 👍 I changed to an arrow function

PR #332, opened 2026-07-24
DRAFT: inject maxTokens defaults from models.dev when prompts omit 

catsby:
# DRAFT

## Summary

Use models.dev information to add defaults if no `maxTokens` is provided.

This is still a draft though

### DIFF
--- a/sdk/llm/src/ai_sdk_options.js
+++ b/sdk/llm/src/ai_sdk_options.js
@@ -1,7 +1,54 @@
 import { loadImageModel, loadTextModel, loadTools } from './ai_model.js';
 import { resolveMessageProviderOptions } from './prompt/block_options.js';
 import { ROLE, isRole } from './utils/message.js';
-import { FatalError } from '@outputai/core';
+import { resolveModelMaxOutputTokens } from './cost/fetch_models_pricing.js';
+import { FatalError, Logger } from '@outputai/core';
+
+const logger = Logger.createLogger( 'LLM' );
+
+// Warn at most once per model to avoid log spam in hot generation loops.
+const warnedUnknownModels = new Set();
+
+// Upper bound on the max output ceiling we inject. models.dev ceilings can be very large
+// (e.g. 64000); a provider already recognizes such models and picks a sane default, so we
+// only need to lift the low unknown-model fallback (often 4096). Capping keeps us from
+// pushing a provider-recognized model above a size that is unsafe for non-streaming calls.
+const MAX_OUTPUT_TOKENS_CAP = 32000;
+
+/**
+ * When a prompt does not set `maxTokens`, default the AI SDK's `maxOutputTokens` to the
+ * model's ceiling from models.dev, capped at MAX_OUTPUT_TOKENS_CAP. Without this, models
+ * the provider does not recognize fall back to its low default (often 4096) and silently
+ * truncate output. Only a limit models.dev actually knows is injected; genuinely unknown
+ * models are left to the provider (a warning is emitted), and a cold/stale table injects
+ * nothing (a background refresh is in flight).
+ *
+ * @param {object} options - AI SDK options being assembled (mutated in place)
+ * @param {object} config - Loaded prompt config ({ provider, model })
+ */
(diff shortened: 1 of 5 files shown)


sdk/llm/src/cost/fetch_models_pricing.js
```diff
@@ -46,7 +62,14 @@ export const fetchModelsPricing = async () => {
     state.errorMessage = buildErrorMessage( error.code ?? error.name ?? error.constructor.name );
   }
 
+  if ( !state.errorMessage && ( state.table === null || typeof state.table !== 'object' ) ) {
+    state.errorMessage = buildErrorMessage( 'malformed response body' );
+  }
+
```
szanata, line 68: This check is wrong. `.json()` will ensure the response body isnt malformed. It will break if the response isn 't a valid json. You could check if the table doesnt have any keys inside

sdk/llm/src/cost/fetch_models_pricing.js
```diff
@@ -55,7 +78,63 @@ export const fetchModelsPricing = async () => {
     return null;
   }
 
-  cache.content = buildModelMap( state.table );
+  const { costs, limits } = buildModelMaps( state.table );
+  cache.content = costs;
+  cache.limits = limits;
```
szanata, line 83: I think these could be combined in .content:

```js
cache.content = { costs, limits  }
```

That reads muck much cleaner.

sdk/llm/src/cost/fetch_models_pricing.js
```diff
@@ -55,7 +78,63 @@ export const fetchModelsPricing = async () => {
...
+  }
+  const maxOutputTokens =
+    cache.limits.get( `${provider}/${model}` ) ??
+    cache.limits.get( model );
+  if ( maxOutputTokens !== undefined ) {
+    return { status: 'known', maxOutputTokens };
+  }
+  // Absent from the table. If the table is stale a refresh is in flight and the model may
+  // appear once it lands, so report 'cold' (no warning) rather than a false 'unknown'.
+  return { status: isStale ? 'cold' : 'unknown', maxOutputTokens: null };
```
szanata, line 139: The status should be a const:

```
const MaxOutputTokenStatus = {
  KNOWN: 1,
  UNKNOWN: 2
  STALE: 3
}
```

sdk/llm/src/ai_sdk_options.js
```diff
@@ -1,7 +1,54 @@
...
+ * model's ceiling from models.dev, capped at MAX_OUTPUT_TOKENS_CAP. Without this, models
+ * the provider does not recognize fall back to its low default (often 4096) and silently
+ * truncate output. Only a limit models.dev actually knows is injected; genuinely unknown
+ * models are left to the provider (a warning is emitted), and a cold/stale table injects
+ * nothing (a background refresh is in flight).
+ *
+ * @param {object} options - AI SDK options being assembled (mutated in place)
+ * @param {object} config - Loaded prompt config ({ provider, model })
+ */
+const applyModelMaxOutputDefault = ( options, config ) => {
```
szanata, line 29: This method mutates its input, I think this should be avoided. It could return the  the tokens or undefined if stale/unkown.

sdk/llm/src/ai_sdk_options.js
```diff
@@ -34,6 +81,8 @@ export const loadAiSdkTextOptions = prompt => {
 
   if ( prompt.config.maxTokens ) {
     options.maxOutputTokens = prompt.config.maxTokens;
+  } else {
+    applyModelMaxOutputDefault( options, prompt.config );
```
szanata, line 85:
```js
options.maxOutputTokens = resolveModelMaxOutputDefault( prompt.config );
```

This reads better and avoid mutations.

sdk/llm/src/cost/fetch_models_pricing.js
szanata: Also "stale" is clear than "cold", which without good knowledge is confusing.

catsby:
Closing this for now since I'm switching tasks and I don't know that this is a good approach as-is

PR #337, opened 2026-07-29
feat(core): Using parsed data from zod schemas as input and return of workflows, steps and evals (OUT-536)

szanata:
## Summary

- `workflow()`, `step()`, and `evaluator()` now pass the value returned by the Zod `inputSchema` parser to their handler. Workflows and steps also return the value produced by their `outputSchema` parser. Zod transforms, coercions, defaults, and object-key stripping therefore affect the values handled and returned by these components instead of only validating them.

### DIFF
--- a/.changeset/swift-things-sort.md
+++ b/.changeset/swift-things-sort.md
@@ -1,0 +1,5 @@
+---
+"@outputai/core": minor
+---
+
+- `workflow()`, `step()`, and `evaluator()` now pass the value returned by the Zod `inputSchema` parser to their handler. Workflows and steps also return the value produced by their `outputSchema` parser. Zod transforms, coercions, defaults, and object-key stripping therefore affect the values handled and returned by these components instead of only validating them.
--- a/docs/guides/migrations/v0.10.0-to-v0.11.0.mdx
+++ b/docs/guides/migrations/v0.10.0-to-v0.11.0.mdx
@@ -553,6 +553,12 @@ Because schemas run at the component boundary, `.transform()`, `z.coerce`, `.pre
 - **Idempotency (re-parse):** Every workflow run parses its Temporal args again — including runs started with `continueAsNew`. The next run does not know it is a continuation. If you pass values that were already transformed, a non-idempotent transform runs twice (for example appending a suffix on each continue-as-new). Prefer idempotent transforms, or pass wire-format input into `continueAsNew` (the same shape you would pass when starting the workflow).
 - **Catalog JSON Schema:** The worker catalog still converts Zod schemas with `z.toJSONSchema`. Bare `.transform()` often cannot be represented and may omit the schema from the catalog; `z.coerce` and `.pipe()` usually convert. Authors who control both the schema and catalog consumers should treat catalog JSON Schema as best-effort for advanced Zod features.
 
+### Published workflow package types
+
+`WorkflowFunctionWrapper`, `StepFunctionWrapper`, and `EvaluatorFunctionWrapper` are now typed as `Wrapper<InputSchema, OutputSchema>` (schema generics), not `Wrapper<Fn<In, Out>>`. Runtime is unchanged; TypeScript breaks for packages that still ship `.d.ts` built against the old arity (for example a published workflows catalog).
+
+Rebuild and republish those packages against `@outputai/core` v0.11.0 so their declarations match, or release them together with the core bump.
+
 ### Migration steps
 
 Review the `inputSchema` and `outputSchema` of every workflow, step, and evaluator:
@@ -562,6 +568,7 @@ Review the `inputSchema` and `outputSchema` of every workflow, step, and evaluat
 - Check object schemas for unknown properties that Zod strips by default. Use `z.looseObject( { ... } )` when those properties must be preserved.
 - Ensure transforms on **workflow** schemas are deterministic for Temporal replay.
 - Ensure workflow `inputSchema` transforms are idempotent if you `continueAsNew` with values that will be parsed again, or pass wire-format input instead.
+- Republish any package that exports workflow/step/evaluator wrappers so its `.d.ts` matches the new `Wrapper` arity.
 
 ## Workflow runtime changes
 
--- a/docs/guides/migrations/v0.10.0-to-v0.11.0.mdx
+++ b/docs/guides/migrations/v0.10.0-to-v0.11.0.mdx
@@ -516,14 +516,15 @@ In v0.10.x, workflow, step, and evaluator schemas validated values, but componen
 
 For example:
 
-```ts
+```js
 const calculateTotal = step( {
   name: 'calculate_total',
   inputSchema: z.object( {
     count: z.coerce.number().default( 1 )
   } ),
   outputSchema: z.object( {
-    total: z.number()
+    total: z.number(),
+    currency: z.string().default( 'USD' )
   } ),
   fn: async input => ( {
     total: input.count * 10,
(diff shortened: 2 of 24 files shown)


docs/guides/migrations/v0.10.0-to-v0.11.0.mdx
```diff
@@ -506,6 +506,48 @@ Audit libraries that receive a custom `fetch` callback. Ensure the callback uses
...
+  fn: async input => ( {
+    total: input.count * 10,
+    internalNote: 'not part of the public result'
+  } )
+} );
+
+const result = await calculateTotal( {
+  count: '2',
+  requestId: 'request-123'
+} );
```
catsby, line 540: Doesn't this example fail a type check? I'm not sure what it's providing in terms of examples

sdk/core/src/interface/evaluator.spec.js
```diff
@@ -9,8 +9,8 @@ import {
 import { ValidationError } from '#errors';
 
 const validateDefinitionMock = vi.hoisted( () => vi.fn() );
-const validateInputMock = vi.hoisted( () => vi.fn() );
-const validateOutputMock = vi.hoisted( () => vi.fn() );
+const validateInputMock = vi.hoisted( () => vi.fn( input => input ) );
+const validateOutputMock = vi.hoisted( () => vi.fn( output => output ) );
```
catsby: what are these mocks actually testing here?

sdk/core/src/interface/validations/index.js
catsby, line 1: are there no updates to [validations/index.spec.js](https://github.com/growthxai/output/blob/6ce5320057eb22d0275c535cd0ddd44e8ccce442/sdk/core/src/interface/validations/index.spec.js) to verify the returned types are correct? Or are those covered somewhere else?

sdk/core/src/interface/workflow.js
```diff
@@ -99,9 +90,7 @@ export function workflow( { name, description, inputSchema, outputSchema, fn, op
     };
 
     try {
-      validator.validateInput( input );
-      const output = await fn( input, WorkflowContext.build() );
-      validator.validateOutput( output );
+      const output = validator.validateOutput( await fn( validator.validateInput( input ), WorkflowContext.build() ) );
```
catsby: does this impact users who are using `continueAsNew()`? are they now passing in parsed data that will get reparsed? Is that safe for things like `.transform()`?

docs/guides/migrations/v0.10.0-to-v0.11.0.mdx
```diff
@@ -506,6 +506,48 @@ Audit libraries that receive a custom `fetch` callback. Ensure the callback uses
...
+```
+
+In v0.10.x, `input.count` was the string `'2'`, `input.requestId` remained available, and `result.internalNote` was returned. In v0.11.0, `input.count` is the number `2`, Zod's default object behavior removes `requestId`, and the returned result is `{ total: 20 }`.
```
catsby: Does this affect the tracefile? Do we record `{ count: 2 }` or `{ count: '2', requestId: 'r-123' }`?

.changeset/swift-things-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+"@outputai/core": minor
```
catsby, line 2: we change `evaluator.js` is that part of `@outputai/evals`, and if so should we include that here?

docs/guides/migrations/v0.10.0-to-v0.11.0.mdx
```diff
@@ -506,6 +506,48 @@ Audit libraries that receive a custom `fetch` callback. Ensure the callback uses
...
+
+### Migration steps
+
+Review the `inputSchema` and `outputSchema` of every workflow, step, and evaluator:
+
+- Update handlers and callers to use the parsed Zod input and output types.
+- Check schemas using `.transform()`, `z.coerce`, `.default()`, `.catch()`, or `.preprocess()` for values that now change at runtime.
+- Check object schemas for unknown properties that Zod strips by default. Use `.passthrough()` when those properties must be preserved.
```
catsby: > Use `.passthrough()` when those properties must be preserved

I believe `.passthrough` is deprecated: https://zod.dev/v4/changelog#deprecates-strict-and-passthrough

sdk/core/src/interface/validations/index.js
```diff
@@ -10,84 +11,63 @@ import {
...
-/**
- * Validates data using a Zod schema
- * @param {unknown} data
- * @param {ZodType} schema
- * @param {string} prefix - Validation error prefix
- * @throws {ValidationError} If validation fails
- * @returns {void}
- */
-const validate = ( data, schema, prefix = '' ) => {
+const capitalize = word => word.at( 0 ).toUpperCase() + word.slice( 1 );
```
catsby, line 14: can we re-add JSDoc for `validate()` and add one for `capitalize`, or was this intentional?

catsby (commented):
Does this not require updating [step.d.ts](https://github.com/growthxai/output/blob/6ce5320057eb22d0275c535cd0ddd44e8ccce442/sdk/core/src/interface/step.d.ts#L22) or [workflow.d.ts](https://github.com/growthxai/output/blob/6ce5320057eb22d0275c535cd0ddd44e8ccce442/sdk/core/src/interface/workflow.d.ts#L130)?

.changeset/swift-things-sort.md
szanata: It is not, at least not explicitly.

sdk/core/src/interface/validations/index.js
szanata: Yeah, was intentional, these are internal methods, and are very straight forward, JS docs were not really necessary.

docs/guides/migrations/v0.10.0-to-v0.11.0.mdx
szanata: Affects the output. In the trace files:
 - input is always the pre-parsed value, what exactly was used to invoke workflow/step
 - but the output will be the post-parsed value, what the function really returned after parsing it with zod.
 
I think this is right behavior as keep in the file the real input/output, not intermediate values.

sdk/core/src/interface/evaluator.spec.js
szanata: They are just stub for wiring, they arent testing anything.

sdk/core/src/interface/validations/index.js
szanata: I will add an assertion for that, but without testing zod features.

sdk/core/src/interface/workflow.js
szanata: Data will get reparsed. This is fine on most cases.

`.transform()`, `.preprocess()`, `.coerce()`, etc are a different case, as they can be indeponent. And this would be not only a problem for `continueAsNew()`, but also normal input and output themselves.

I think that is acceptable If people want they can use them, and that might break stuff. I will take the instance of documenting this, instead of trying to create zod-like object that doesnt allow transformations, or having different handling for continued as new handler, as that could break determinism as well.

docs/guides/migrations/v0.10.0-to-v0.11.0.mdx
szanata: I have updated the types to actually support the type parsing. New type rules:

### fn (handler)

Input arg: post-parse → z.infer<inputSchema> (defaults / coerce / transforms already applied)
Return: pre-parse → z.input<outputSchema> (what Zod accepts before outputSchema runs)

### Wrapper (caller)
Arg: pre-parse → z.input<inputSchema>
Return: post-parse → z.infer<outputSchema>

### continueAsNew
Arg: pre-parse → z.input<inputSchema> (same as starting a workflow)

docs/guides/migrations/v0.10.0-to-v0.11.0.mdx
szanata: Good catch. I have updated to `z.looseObject( { ... } )`

catsby (approved):

PR #374, opened 2026-09-01
chore: Add Gemini grounded search tracking

catsby:
## Summary

Track and report Google Vertex grounding costs for Gemini 2.x and 3.

Using grounded search in prompts (by adding `tools.googleSearch: {}` to the frontmatter) instructs the model to use searching which bills per request separately from tokens. It bills per grounding prompt on Gemini 2.x, and per grounding query on Gemini 3. Our `models.dev/api.json` doesn't have any rate information for those costs (at this time). That cost wasn't tracked in the tracefile or included in the new `llm:generation:cost` or legacy `cost:llm:request`reporting. 

This PR adds parsing `groundingMetadata.webSearchQueries` across all response steps and prices it from a local rate table; unknown model families record the quantity as an unpriced line.

### DIFF
--- a/sdk/llm/src/utils/grounding.js
+++ b/sdk/llm/src/utils/grounding.js
@@ -6,8 +6,8 @@
  * Verified 2026-09-01: https://cloud.google.com/vertex-ai/generative-ai/pricing
  */
 const UNITS = [
-  { match: /^gemini-3/, label: 'grounding_query', ppm: 14_000 },
-  { match: /^gemini-2/, label: 'grounding_prompt', ppm: 35_000 }
+  { match: /^gemini-3/, label: 'grounding_query', ppm: 14_000, perQuery: true },
+  { match: /^gemini-2/, label: 'grounding_prompt', ppm: 35_000, perQuery: false }
 ];
 
 /**
@@ -29,7 +29,7 @@ export const isGroundingLabel = label => label === GROUNDING_UNKNOWN_LABEL || la
  * here if image or retrieval grounding is enabled.
  *
  * @param {string} modelId - Id of the model that produced the response
- * @param {object} [providerMetadata] - AI SDK provider metadata of the final step
+ * @param {object} [providerMetadata] - AI SDK provider metadata of a single step
  * @returns {{ label: string, amount: number } | null} Grounding label and amount, or null when ungrounded
  */
 export const parseGroundingUsage = ( modelId, providerMetadata ) => {
@@ -46,6 +46,6 @@ export const parseGroundingUsage = ( modelId, providerMetadata ) => {
 
   return {
     label: unit.label,
-    amount: unit.label === 'grounding_query' ? queries.length : 1
+    amount: unit.perQuery ? queries.length : 1
   };
 };
--- a/sdk/llm/src/utils/grounding.js
+++ b/sdk/llm/src/utils/grounding.js
@@ -1,0 +1,51 @@
+/**
+ * Grounding with Google Search is billed per request, not per token, and models.dev carries no
+ * per-request rate, so the table lives here. Gemini 3 bills per Grounding Query; Gemini 2.x bills
+ * per Grounding Prompt, once, regardless of query count. Different units, so the family decides
+ * both the label and the amount. Rates are price-per-million to match `ppm` semantics.
+ * Verified 2026-09-01: https://cloud.google.com/vertex-ai/generative-ai/pricing
+ */
+const UNITS = [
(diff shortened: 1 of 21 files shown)


sdk/llm/src/utils/wrap.js
```diff
@@ -16,6 +16,20 @@ const createResponseProxy = ( { response, properties } ) => new Proxy( response,
...
+ * AI SDK v7 attaches provider metadata to `finalStep`; a text/stream response missing `finalStep`
+ * signals a shape regression, so surface it before silently substituting the response-level value.
+ * Image responses have no `finalStep` by design, so callers pass `expectFinalStep: false` there.
+ */
+const resolveProviderMetadata = ( response, { expectFinalStep = true } = {} ) => {
+  if ( expectFinalStep && !response.finalStep ) {
+    Logger.warn( 'Response missing finalStep; using response-level provider metadata', { namespace: 'LLM' } );
+  }
+  return response.finalStep?.providerMetadata ?? response.providerMetadata;
+};
```
szanata: This isnt necessary. The previous code was already reading the providerMetadata from the right field. I don't think we need to test for shape regression on their side.

sdk/llm/src/utils/legacy_cost_attribute.js
```diff
@@ -71,5 +75,13 @@ export const convertCostToLegacy = cost => {
...
   }
 
+  // Included even when unrated (status 'missing'), same as cacheReadItem above,
+  // so legacy consumers still see that a grounded call happened.
+  const groundingItem = cost.items.find( ( { group } ) => group === 'request' );
+  if ( groundingItem ) {
+    const ppm = groundingItem.status === 'missing' ? 0 : groundingItem.ppm;
+    usage.push( { type: groundingItem.label, ppm, amount: groundingItem.amount } );
+  }
+
```
szanata: I think this should not be added to legacy cost attribute. This now can break any .reduce/.map as we are adding new possible items to the usage array. 

The legacy attribute should be frozen and new things go only to the new attribute.

sdk/llm/src/utils/usage.js
```diff
@@ -113,6 +115,17 @@ export const parseLLMUsage = ( { prompt, usage } ) => {
     }
   }
 
+  // Grounding is billed per step, so aggregate across every step rather than only the final one.
+  // Per-query families (Gemini 3) return queries.length per step; per-prompt families (Gemini 2.x)
+  // return 1 per grounded step. Summing yields total queries and grounded-step count respectively.
+  const grounding = ( steps ?? [] )
+    .map( step => parseGroundingUsage( modelId, step?.providerMetadata ) )
```
szanata: We should not call `parseGroudingUsage` if there isnt `providerMetadata` object, as it doesnt work without it. Looking at this call now, seems that it is optional, whereas it isnt.

Either:
```
( steps ?? [] )
   .filter( step => step?.providerMetadata )
    .map( step => parseGroundingUsage( modelId, step.providerMetadata ) )
...
```

Or send the whole `step` as argument: `parseGroudingUsage( { modelId, step } )`.

sdk/llm/src/utils/grounding.js
```diff
@@ -0,0 +1,51 @@
...
+];
+
+/**
+ * Unknown family: still record the quantity so the call is visibly unpriced rather than silently
+ * token-only. No rate resolves for this label, so the item lands MISSING and the cost attribute
+ * goes INCOMPLETE.
+ */
+export const GROUNDING_UNKNOWN_LABEL = 'grounding';
+
+export const GROUNDING_PPM = Object.fromEntries( UNITS.map( u => [ u.label, u.ppm ] ) );
```
szanata: This is a enum style map, so the best practice is for PascalCase naming: `GroudingPPM.*`

But given the intended usage, I would say this makes more sense as a map: 

```js
export const GroudingPpmMap = new Map( UNITS.map( u => [ u.label, u.ppm ] );
...
GroudingPpmMap.get( 'x' )
```

sdk/llm/src/utils/grounding.js
```diff
@@ -0,0 +1,51 @@
...
+
+export const isGroundingLabel = label => label === GROUNDING_UNKNOWN_LABEL || label in GROUNDING_PPM;
```
szanata: This method isnt necessary, it can be replace by testing group === 'request' in the only place it is used.

szanata (changes requested):

szanata (approved):

PR #379, opened 2026-09-07
refactor(core): New shutdown handling (OUT-565)

szanata:
## Summary

Hardened worker startup and shutdown and the catalog workflow handling.

- Changed catalog reconciliation to terminate a stale catalog instead of asking it to complete, so a catalog that stopped processing workflow tasks no longer fails every deploy that follows it.
- Increased the catalog workflow task timeout to 30s, enough for a cold worker bundle to activate.
- Added two retries when publishing the catalog before failing the worker.
- Added a startup gate so a worker never starts unless its catalog matches its own source code.
- Changed exit codes so kill signals drain in-flight work and exit `0`, while only real failures exit `1`. Alerts that watch for non-zero worker exits will stop firing on every deploy.
- Added graceful draining for uncaught exceptions and unhandled rejections, with a force quit if the drain stalls.
- Fixed kill signals during startup killing the process abruptly, since the handler was only installed once the worker was built.

#### Force quite (two consecutive interruptions)
quit

#### Interruption
interruption

#### Interruption with Workflow load
interruption_with_load

#### Unhandled Rejection
uncaught_error

#### Uncaught Error
uncaught_error

#### Connectin Lost
connection_lost

#### Catalog Start
catalog_started

#### Catalog Start Skipped 
catalog_already_running

#### Old Catalog Termination
catalog_terminated

#### Catalog Start failure
catalog_failure


sdk/core/src/worker/index.spec.js
```diff
@@ -199,242 +223,320 @@ describe( 'worker/index', () => {
...
+    it( 'logs a shutdown failure without skipping the remaining steps', async () => {
+      await bootWorker();
+      workerRunnerInstance.stop.mockRejectedValueOnce( new Error( 'drain failed' ) );
+
+      interruption.controller.abort( new interruption.KillSignError( 'SIGTERM' ) );
+      await waitForExit();
+
+      expect( mockLog.warn ).toHaveBeenCalledWith( 'Stopping Worker error', { error: 'drain failed' } );
+      expect( mockConnection.close ).toHaveBeenCalledOnce();
+      expect( process.exit ).toHaveBeenCalledWith( 0 );
```
catsby: The PR description reads _"Changed exit codes so kill signals drain in-flight work and exit `0`, while only real failures exit `1`"_

If I read this test correctly, it seems to simulate the drain itself failing, while still exiting `0`. Wouldn't the drain failing be a "real failure"?

sdk/core/src/worker/interruption.spec.js
```diff
@@ -1,98 +1,177 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
-import { setupInterruptionHandler } from './interruption.js';
 
-const { mockLog } = vi.hoisted( () => ( {
-  mockLog: { info: vi.fn(), warn: vi.fn() }
+const FORCE_QUIT_GRACE_MS = 1000;
+const FORCE_QUIT_AFTER_FAILURE_MS = 60_000;
```
catsby: I think this have been renamed in interruption.js, should we match them here?

sdk/core/src/worker/interruption.js
```diff
@@ -1,33 +1,72 @@
+import { serializeError } from '#helpers/error_serializer';
 import { createChildLogger } from '#logger';
 
 const log = createChildLogger( 'Interruption' );
 
-const FORCE_QUIT_GRACE_MS = 1000;
+const DOUBLE_SIGNAL_IGNORE_TIME = 1000;
+const KILL_AFTER_UNCAUGHT_TIME = 60_000;
```
catsby, line 7: Does this impact the `TEMPORAL_SHUTDOWN_FORCE_TIME` or `TEMPORAL_SHUTDOWN_GRACE_TIME` settings?

sdk/core/src/worker/index.js
```diff
@@ -46,194 +45,150 @@ const {
...
-  } )
-  .catch( async error => {
-    log.error( 'Fatal error', { error: serializeError( error ) } );
+    await shutdown();
 
-    mainEventBus.emit( BusEventType.RUNTIME_ERROR, { error } );
+    const hasError = signal.aborted && !( signal.reason instanceof KillSignError );
+    if ( hasError ) {
+      log.error( 'Worker error', { error: serializeError( signal.reason ) } );
+      mainEventBus.emit( BusEventType.RUNTIME_ERROR, { error: signal.reason } );
```
catsby, line 170: can this `emit` call user hook code? do we need to make sure there's a `catch` if that code throws?

sdk/core/src/worker/index.js
```diff
@@ -46,194 +45,150 @@ const {
...
+  const shutdownClients = [
+    { label: 'Stopping Worker', running: () => state.workerRunner?.running, stop: () => state.workerRunner.stop() },
+    { label: 'Stopping Connection Monitor', running: () => state.connectionMonitor?.running, stop: () => state.connectionMonitor.stop() },
+    { label: 'Interrupting Catalog Publisher', running: () => state.catalogPublisher?.running, stop: () => state.catalogPublisher.interrupt() },
+    { label: 'Closing Connection', running: () => state.connection, stop: () => state.connection.close() }
+  ];
+  for ( const { label, running, stop } of shutdownClients ) {
+    if ( running() ) {
+      log.info( `${label}...` );
+      await stop().catch( e => log.warn( `${label} error`, { error: e.message } ) );
```
catsby: we use `serializeError()` in a few places in this file, do we need that here?

sdk/core/src/worker/index.js
szanata: Emitting events never throw. And the code that handles user hooks is safe:
```js
  const promise = Promise.resolve()
    .then( () => fn( args ) )
    .catch( e => log.error( `${hookName} hook error`, { error: serializeError( e ) } ) )
    .finally( () => pendingHooks.delete( promise ) );
```
at https://github.com/growthxai/output/blob/a4a41b660c4b952c167c5cdae68e23166ecea3be/sdk/core/src/hooks/index.js#L16

sdk/core/src/worker/index.js
szanata: Good idea!

sdk/core/src/worker/interruption.js
szanata: Yeah, if the worker didnt shutdown in 1min we will kill it anyway, ignoring Temporal shutdown. It is an improvement vs what we have today, where any uncaught error/unhandled rejection kills immediately.
And to be honest, 1min is fine, because we set the default for shutdown at 25s. Users might differ thou, but if this becomes an issue in the future we could introduce an ENV var.

sdk/core/src/worker/interruption.spec.js
szanata: Yes, I missed that

sdk/core/src/worker/index.spec.js
szanata: I interpreted that only the thing that causes the shutdown - the first error or the kill signal - translates to the exit code. Any errors that happen during shutdown are inconsequential and they might happen because of the shutdown itself, so in my interpretation they don't change the exit code.

sdk/core/src/worker/index.js
szanata: Done. I also move this whole function to a dedicated file, so it has more in depth tests.

sdk/core/src/worker/interruption.spec.js
szanata: Done.

sdk/core/src/worker/interruption.js
catsby: If users change the defaults so that either `TEMPORAL_SHUTDOWN_FORCE_TIME` or `TEMPORAL_SHUTDOWN_GRACE_TIME` exceed `KILL_AFTER_UNCAUGHT_TIME`, or if combined they exceed it, they won't be honored though, right?

PR #381, opened 2026-09-08
feat(llm): Added built-in fallback models pricing table (OUT-572)

szanata:
## Summary

- Added a static subset of the models pricing table with support for the 6 shipped providers. This will be used if the live provider is down and there is no cache.

### DIFF
--- a/sdk/llm/src/utils/models_pricing.js
+++ b/sdk/llm/src/utils/models_pricing.js
@@ -44,7 +44,7 @@ const parseData = data => {
     }
     return map;
   } catch ( error ) {
-    logger.error( `Models pricing: Data parsing failure "${error.name}".` );
+    logger.error( `Models pricing: Data parsing failure "${error.message}".` );
     return null;
   }
 };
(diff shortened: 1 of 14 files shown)


sdk/llm/src/utils/models_pricing_fallback.json
catsby, line 1: How was this file constructed, and how do we reconstruct it in the future as `models.dev/api.json` advances? I think we should a script that rebuilds this so it's reproducible.

sdk/llm/src/utils/models_pricing.js
```diff
@@ -64,5 +67,6 @@ export const fetchModelsPricing = async () => {
     return cache.content;
   }
 
-  return null;
+  logger.warn( 'Models pricing: using built-in fallback.' );
+  return fallbackTable;
```
catsby: should we cache this? it looks like if there is no pricing cache we attempt to pull it from models.dev, and if that fails then we end up falling back to this, but we don't cache it. If I read that correctly then every attempt to price will attempt to call models.dev, is that correct?

sdk/llm/src/utils/models_pricing_fallback.json
szanata: This will be done in another task. This one is just about adding support for this.

sdk/llm/src/utils/models_pricing.js
szanata: Yes, I am doing just that.

PR #389, opened 2026-09-14
refactor(core): Exit Code 1 on any worker drain error

szanata:
## Summary

- Removed the condition to that caused drain errors of the type `GracefulShutdownPeriodExpiredError` to exit with code `0`. That was not possible in reality as Temporal doesn't reliably raises that exception.
- Added exit code to the last log line.


docs/guides/migrations/v0.12.0-to-v0.13.0.mdx
```diff
@@ -51,18 +51,18 @@ Every path that ends the worker changed shape. The exit code is also explicit no
 | Scenario | v0.12.0 | v0.13.0 |
 | --- | --- | --- |
 | `SIGTERM`, `SIGINT` or `SIGUSR2` while polling | Stops polling and asks in-flight activities to cancel at once, then waits for them for as long as they take. Exits `0` | Stops polling and lets in-flight activities run for `TEMPORAL_SHUTDOWN_GRACE_TIME` before asking them to cancel. Exits `0` |
-| In-flight work outlives the drain | Nothing bounds it, so the worker waits until the platform sends `SIGKILL` and the hook flush is never reached. Exit code is whatever the platform reports | Stops draining at `TEMPORAL_SHUTDOWN_FORCE_TIME`, warns with `GracefulShutdownPeriodExpiredError`, flushes hooks. Exits `0`, since the bound worked as intended |
+| In-flight work outlives the drain | Nothing bounds it, so the worker waits until the platform sends `SIGKILL` and the hook flush is never reached. Exit code is whatever the platform reports | Stops draining at `TEMPORAL_SHUTDOWN_FORCE_TIME`, warns with `Stopping Worker error`, flushes hooks. Exits `1` |
```
catsby, line 54: v0.13.0 has already been released, shouldn't this `v0.12.0 → v0.13.0` section remain untouched? It's still accurate isn't it? I would think we'd have a v0.13.0 → v0.13.1 or something. I believe this is what the Claude review is objecting on as well.

szanata: I wanted to clarify that the release notes were wrong and this behavior doesnt hold.

catsby (approved):
Despite the github-review bot insisting, local testing is not agreeing with what it's reporting on the exit `0`, as tested by two different engineers. `GracefulShutdownPeriodExpiredError` does show up, but it's the `IllegalStateError` that's being warned about consistently, so this doesn't represent a behavior change from v0.13.0

sdk/core/src/worker/index.js
```diff
@@ -174,11 +174,10 @@ execute()
     log.info( 'Flushing hook callbacks...' );
     await flushPendingHooks( hookFlushTimeoutMs );
 
-    // An expired drain is the bound working as intended, not a worker failure, so it keeps exit code 0.
-    const workerFailedToStop = shutdownFailures.some( v => v.service === 'worker' && !( v.error instanceof GracefulShutdownPeriodExpiredError ) );
+    const workerFailedToStop = shutdownFailures.some( v => v.service === 'worker' );
```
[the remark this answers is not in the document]
szanata: That doesn hold true in the actual execution.

PR #391, opened 2026-09-15
feat(llm): Record usage for generation failures

szanata:
# feat(llm): Record usage and cost for generation failures

Usage and cost were read from the value a generation returned, so a call that threw reported nothing -
even when the provider had already answered and the failure came afterwards, in output validation, in
the tool loop, or in our own post-response code. Those tokens were billed by the provider and invisible
to us.

Usage is now collected from the AI SDK lifecycle events as the call progresses, so billing no longer
depends on reaching a return statement. Any generation that completed at least one step reports what it
spent, whether it ends with a response, an error, or an abort.

Attribute and event shapes are unchanged. Failed calls simply stop being free, which raises reported
cost for unchanged traffic - covered by the migration guide, together with the one error break: a
failure in our own response handling now throws a non-retryable `FatalError` wrapping the original
error, and the generic streaming messages were normalized.

## What changed

- `Metering` (new, `utils/metering.js`) collects step usage as the SDK reports it and bills once, from
  the response on the paths that have one and from the recorded steps on the paths that do not. Billing
  is idempotent and an empty collector does not count as billed, so a later lifecycle event can still
  report. Retries keep billing per attempt, because each attempt calls the provider.
- `wrapGeneration` is split into `wrapTextGeneration` and `wrapImageGeneration`, since image calls
  expose no lifecycle events and have no step loop.
- `wrapTextGeneration` and `wrapStream` hand the call the hooks it wires into the AI SDK -
  `onStepEndHook` for `onStepEnd`, plus `onEndHook`, `onErrorHook` and `onAbortHook` for streams - so
  usage survives the exits that report no response.
- `wrapStream` records the abort reason and a rejected structured output on the LLM trace, so a
  cancelled stream and an output that never validated are no longer traced as clean calls.
- `drainStream` consumes the whole stream after a failure before rethrowing, because step usage is only
  reported while something keeps reading. A stream that goes quiet for 250 ms after failing is
  abandoned, and the original error is thrown anyway.
- `resolveUsage` falls back to summing the steps whenever the aggregate usage is missing or carries no
  token counts, which is what an aborted or truncated stream reports.
- A failure in our own response handling - reading the response, merging sources, building the returned
  proxy - is recorded on the trace and rethrown as a non-retryable `FatalError`, since it is
  deterministic and a retry pays for the model call again to fail the same way. The generic streaming
  messages were normalized to `Streaming aborted.`, `Streaming failed.` and
  `Streaming completed without a response.`.
- `extractSources` drops entries carrying neither a url nor an id, and logs a malformed `sources`
  payload instead of failing the call. Surviving sources are returned exactly as the SDK produced them;
  only the deduplication key is trimmed.
- Image serialization moved to `utils/image.js`, where a throwing `base64` getter logs and yields no
  images rather than failing a call that already paid for the image.

### generateText / Agent.generate

- Billed: output or schema validation failure (`NoObjectGeneratedError`, `NoOutputGeneratedError`), a
  tool loop that breaks after a completed step, an abort after a completed step, our own post-response
  code throwing (`Agent.generate` message store, response mapping, proxy build - the last two as a
  `FatalError`)
- Not billed: any failure before the first step completes, an abort before the first step

### generateTextWithStreaming / Agent.generateWithStreaming

- Billed: provider error part mid stream, an abort after a completed step, a structured output that
  never validates, the "completed without a response" guard, a message store failure after the stream
  finished
- Not billed: failure before the first chunk, an abort before the first step, a provider that stalls
  without closing the stream and whose step events had not arrived before it went quiet for 250 ms

### streamText / Agent.stream

The consumer drives consumption, so capture depends on the stream being read.

- Billed: normal completion, an error part with steps already collected, an abort after a completed
  step
- Not billed: an abort before the first step, a consumer that abandons the stream, a failure before the
  first step, process exit before a terminal event

### generateImage

No lifecycle events and no step loop, so a failure always precedes any usage. Nothing is billed on any
error path, by design.


sdk/llm/src/utils/wrap.js
```diff
@@ -99,98 +118,110 @@ export const wrapGeneration = async ( { name, prompt, fn } ) => {
...
-      throw handleError( { traceId, error } );
-    }
 
+    metering.recordResponse( response );
+    await metering.bill();
+
+    const { text: result, finalStep, usage } = response;
+    const sources = extractSources( response );
+    Tracing.addEventEnd( { id: traceId, details: { result, usage, providerMetadata: finalStep?.providerMetadata, sources } } );
+    const proxyResponse = createResponseProxy( { response, properties: { cost: metering.attributes.cost, sources, result } } );
```
catsby: this used to be wrapped in a `try/catch` but it's not anymore, that I can see. Is that intentional, and is there any concern of something throwing in here?

sdk/llm/src/utils/wrap.js
```diff
@@ -99,98 +118,110 @@ export const wrapGeneration = async ( { name, prompt, fn } ) => {
...
-        error: callbackError instanceof Error ? callbackError.message : String( callbackError )
-      } );
+    } catch ( e ) {
+      Logger.error( 'Stream onError() callback failed', { namespace: 'LLM', error: e?.message ?? String( e ) } );
     }
   };
 
+  /** The consumer owns `stream.output`, so its rejection only reaches the trace if we observe it too */
+  const recordOutputError = stream => {
+    stream?.output?.catch( error => handleError( { traceId, error } ) );
```
catsby: is `handleError` being attached multiple times? It looks like here, but also in `onErrorHook` and maybe somewhere in `handleAbort()` as well?

sdk/llm/src/utils/usage.js
```diff
@@ -58,19 +59,37 @@ export class LLMGenerationUsage extends Tracing.Attribute.BaseAttribute {
...
+ * Returns usage from response or reconstruct it from steps.
+ *
+ * The token check is load-bearing: aborted and truncated calls still report an aggregate, but with
+ * every token count undefined, and the steps are then the only record of what was already spent.
+ */
+const resolveUsage = ( { usage, steps } ) => {
+  if ( usage && ( exists( usage.inputTokens ) || exists( usage.outputTokens ) ) ) {
+    return usage;
+  }
+  return Array.isArray( steps ) && steps.length > 0 ? extractUsageFromSteps( steps ) : null;
```
catsby, line 71: If I read `extractUsageFromSteps()` correctly it can return `{}` which is truthy, right? Does that matter here?

sdk/llm/src/utils/usage_tools.js
catsby, line 1: is there value in having these ~23 lines in their own file instead of just in `usage.js`?

sdk/llm/src/utils/wrap.js
```diff
@@ -30,66 +28,87 @@ const handleError = ( { traceId, error: originalError } ) => {
...
+    await metering.bill();
 
-      Tracing.addEventEnd( { id: traceId, details: { result: mappedImages, usage, providerMetadata } } );
-      return createResponseProxy( { response, properties: { cost, result: image } } );
-    } else {
-      const { text: result, finalStep } = response;
-      const { providerMetadata } = finalStep;
-      const sources = extractSources( response );
+    const { usage, image, images, providerMetadata } = response;
+    const mappedImages = images.map( ( { mediaType, base64 } ) => ( { size: calculateBase64FileSize( base64 ), mediaType } ) );
```
catsby: Do we not need the previous `Array.isArray()` check anymore?

sdk/llm/src/utils/wrap.js
szanata: yes, it isnt wrapped because this part has no IOs so it is suppose to not throw anyway. `extractSources` can handle null/undefined and incomplete values. This was wrapped before because of the `response = await fn()`.

sdk/llm/src/utils/wrap.js
szanata: Yes, it is attached on all possible failure points. However they mostly don't happen together. But even if they would do, the trace system will use only the last error, so deduplicating doesn't need to be a concern.

sdk/llm/src/utils/usage.js
szanata: No. The code will not break, it will simple not calculate usage if both `usage` and `usageFromSteps` are empty, since there is no data.

sdk/llm/src/utils/wrap.js
szanata: To be on the safe side:
- Updated everything to be on try/catches
- Added try/catch around `extractSources` and improved it. Its failures are log only, to not discard valid responses.

sdk/llm/src/utils/usage_tools.js
szanata: I think it makes easier to test and also this can be mocked as dependency, simplifying tests for the `usage.js`.

sdk/llm/src/utils/wrap.js
szanata: We need. And a re-added as part of making this code works with null/empty responses.

szanata: I moved this a separate file, with a try/catch around it, and since this is used only for tracings, failures are log only, so we dont discard valid responses.

catsby (approved):
👍 pending any open Claude-bot issues

